PR #7, opened 2016-11-02
Updating configurations for third-party integrations

dexter1691:


.coveragec
```diff
@@ -1,6 +1,20 @@
...
+    @authentication_classes
+    @permission_classes
+    @csrf_exempt
+
+omit =
+    */templates/*
+    */static/*
+    */python?.?/*
+    */migrations/*
+    */tests/*
```
deshraj: @dexter1691 can you please add a newline here?

settings/prod.py
```diff
@@ -5,20 +5,22 @@
...
-        'PORT' : '5432',
+        'USER': os.environ.get('RDS_USER', ''),
+        'PASSWORD': os.environ.get('RDS_PASSWORD', ''),
+        'HOST': os.environ.get('RDS_HOST', ''),
+        'PORT': '5432',
     }
 }
 
-INSTALLED_APPS += ('storages',)
+INSTALLED_APPS = ('storages',)
```
deshraj: @dexter1691 I think we don't need to change this, right ?

dexter1691: INSTALLED_APPS was not defined. That's what the error warning was.

PR #10, opened 2016-11-03
Frontend setup

aka-jain:
Initiallizing frontend directory for EvalAI UI code base. Integrating with Gulp, SASS and Angular.


dexter1691:
Are we planning to serve the files using Django or Node.js. I thought @deshraj said that serving the files using Django will be easier. Serving with Node.js will mean that we have to run things on two different ports and build a reverse proxy. Not sure we want to do that.

deshraj:
@dexter1691 I think we agreed on having a django server for REST APIs and a nodejs server running on different port which will consume APIs, no? 

This will keep frontend decoupled from backend. Let me know if you have concerns.

dexter1691:
I see. How will it run on AWS. Will it run on same server? Or on different server?
If same server: how do you deal with reverse proxy
If different server: we will have to scale / manage two servers.

deshraj:
Ammm...Good point. Earlier I was planning to go with the first method i.e setting both django and nodejs on the same EBS env. 

Setting the reverse proxy would be a bit complex on EBS and running two projects in the same EBS enviroment doesn't seem like a good idea. We can serve everything using django only(Involves less pain).

frontend/.gitignore
```diff
@@ -0,0 +1,6 @@
+# files/directories to be ignored
+
+.sass-cache/
+
+node_modules/
+bower_components/
```
deshraj: @aka-jain can you add a newline here ?

frontend/index.html
```diff
@@ -0,0 +1,54 @@
...
+
+<!-- evalAI main scripts -->
+<script type="text/javascript" src="dist/assets/js/app.js"></script>
+<script type="text/javascript" src="dist/assets/js/controllers.js"></script>
+<script type="text/javascript" src="dist/assets/js/directives.js"></script>
+<!-- <script type="text/javascript" src="dist/assets/js/services.js"></script> -->
+
+
+</body>
+</html>
```
deshraj: Please add a newline here also.

aka-jain: Sure @deshraj. Will do it.

frontend/.gitignore
aka-jain: Sure.

dexter1691:
@aka-jain: Plz make the necessary changes and merge the master branch to approve the PR. Thanks!

aka-jain:
@dexter1691 : sure, will do it as soon as I complete the setup

aka-jain:
Closing this PR and creating a new one as there is a huge change in UI code base.

PR #26, opened 2016-11-12
Static and Media files settings on S3 modified

deshraj:


settings/prod.py
```diff
@@ -24,14 +24,19 @@
 
 INSTALLED_APPS += ('storages',)  # noqa
 
+AWS_STORAGE_BUCKET_NAME = "evalai"
```
dexter1691: Would be good to have evalai/static bucket. Putting the static files in the top most hierarchy doesn't sound right.

deshraj: So, the bucket name is `evalai` and static files will be stored in `evalai/static` and media files will be stored in `evalai/media`. Does that sound good?

dexter1691: My bad!! Didn't notice that. LGTM

PR #31, opened 2016-11-12
Challenges cyclic import problem fix

TroJan:
This pr updates the current code base, so that run server can be properly run. It comments out some import to resolve circular imports.

### DIFF
--- a/apps/hosts/models.py
+++ b/apps/hosts/models.py
@@ -1,7 +1,6 @@
 from __future__ import unicode_literals
 
 from accounts.models import (TimeStampedModel, UserStatus)
-from challenges.models import (Challenge)
 
 from django.db import models
 from django.contrib.auth.models import User
@@ -21,7 +20,7 @@ class HostPermission(TimeStampedModel):
     WRITE = 'Write'
     READ = 'Read'
     RESTRICTED = 'Restricted'
-    name = models.Chaield(max_length=30)
+    name = models.CharField(max_length=30)
     status = models.CharField(max_length=30, unique=True)
 
     def __unicode__(self):
@@ -29,8 +28,10 @@ class HostPermission(TimeStampedModel):
 
 
 class ChallengeHostTeams(TimeStampedModel):
-    challenge = models.ForeignKey(Challenge)
     team_name = models.CharField(max_length=64, null=False, blank=False)
+
+    class Meta:
+        app_label = 'hosts'
 
 
 class ChallengeHost(TimeStampedModel):
--- a/settings/common.py
+++ b/settings/common.py
@@ -11,10 +11,12 @@ https://docs.djangoproject.com/en/1.10/ref/settings/
 """
 
 import os
-
+import sys
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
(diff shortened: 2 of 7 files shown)


dexter1691:
@TroJan: Thanks for pointing this out! 

Thanks for the cyclic import issue fix. 

The `challenge` field in the `ChallengeHostTeams` model is redundant since we already have that information in the form of `creator` field inside `ChallengeHostTeams`. This also makes sense since the same team can create multiple competitions. 

For instance: VQA team might create VQA Real Image Competition and VQA Abstract Image Competition. 

@deshraj: This also sounds like a good workflow. To create a competition, first create a team. We allow single user in the teams so that's alright. A team has to exist for a competition to be created.

apps/hosts/models.py
```diff
@@ -21,17 +21,20 @@ class HostPermission(TimeStampedModel):
...
+    name = models.CharField(max_length=30)
     status = models.CharField(max_length=30, unique=True)
 
     def __unicode__(self):
         return self.name
 
 
 class ChallengeHostTeams(TimeStampedModel):
-    challenge = models.ForeignKey(Challenge)
+    # challenge = models.ForeignKey(Challenge)
```
dexter1691: Can you completely get rid of this line.

apps/hosts/models.py
```diff
@@ -1,7 +1,7 @@
 from __future__ import unicode_literals
 
 from accounts.models import (TimeStampedModel, UserStatus)
-from challenges.models import (Challenge)
+# from challenges.models import (Challenge)
```
dexter1691: Same here. Please get rid of this.

settings/common.py
```diff
@@ -11,10 +11,12 @@
 """
 
 import os
-
+import sys
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+APPS_DIR = os.path.join(BASE_DIR, 'apps')
 
+sys.path.append(APPS_DIR)
```
deshraj: I don't think we would need this since `apps` directory is already there on PYTHONPATH(root of the project).

deshraj:
@dexter1691 This workflow sounds good to me. Also, it makes us a different from others. :-)

apps/hosts/models.py
TroJan: Sure

settings/common.py
TroJan: If i remove apps from system path, and change all import to be of the form `from apps.X.models import Y` I still get an error 'No module named X.models`

Any suggestions to get rid of this?

settings/common.py
```diff
@@ -41,13 +43,13 @@
...
 
 OUR_APPS = [
-    'apps.analytics',
-    'apps.challenges',
-    'apps.hosts',
-    'apps.jobs',
-    'apps.participants',
-    'apps.teams',
-    'apps.web',
+    'analytics',
```
deshraj: @TroJan can you add `accounts` here so that we can merge the PR?

Issue #35, opened 2016-11-13
Create endpoint to get all the teams


deshraj:
Details about the endpoint:
- Url: /api/teams/challenge/<challenge_id>
- Header: Authorization: Token <token>
- Method: GET
- Response:
  - status: HTTP_200_OK

Also, please add the pagination for the same. 

For more details, feel free to discuss on thread or gitter channel https://gitter.im/Cloud-CV/EvalAI

deshraj:
Fixed in #60

PR #60, opened 2016-11-16
Team: Add Get api to list all teams for a particular challenge.

TroJan:
Issue #35

### DIFF
--- a/apps/teams/views.py
+++ b/apps/teams/views.py
@@ -1,3 +1,35 @@
+from django.conf import settings
 from django.shortcuts import render
 
-# Create your views here.
+from rest_framework import permissions, status
+from rest_framework.authentication import (TokenAuthentication,)
+from rest_framework.decorators import (api_view,
+                                       authentication_classes,
+                                       permission_classes,)
+from rest_framework.pagination import PageNumberPagination
+from rest_framework.response import Response
+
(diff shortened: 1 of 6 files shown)


apps/teams/views.py
```diff
@@ -1,3 +1,37 @@
...
+
+from challenges.models import Challenge
+
+from .models import Team
+from .serializers import TeamChallengeSerializer
+
+
+@api_view(['GET'])
+@permission_classes((permissions.IsAuthenticated,))
+@authentication_classes((SessionAuthentication, TokenAuthentication))
```
deshraj: I am skeptical about having `SessionAuthentication`. Do we need it?

apps/teams/views.py
```diff
@@ -1,3 +1,37 @@
...
+def team_list(request, challenge_pk):
+
+    try:
+        challenge = Challenge.objects.get(pk=challenge_pk)
+    except Challenge.DoesNotExist:
+        response_error = {'error': 'Challenge does not exist'}
```
deshraj: Lets make the dictionary `response_data` all over the APIs.

apps/teams/views.py
TroJan: yeah for rest we dont need.

PR #61, opened 2016-11-16
Add django-cors package to handle CORS.

TroJan:
@deshraj : settings for cors.

### DIFF
--- a/settings/common.py
+++ b/settings/common.py
@@ -56,6 +56,7 @@ OUR_APPS = [
 THIRD_PARTY_APPS = [
     'allauth',
     'allauth.account',
+    'corsheaders',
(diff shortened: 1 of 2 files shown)


settings/common.py
```diff
@@ -144,3 +146,5 @@
 }
 
 # ALLAUTH SETTINGS
+
+CORS_ORIGIN_ALLOW_ALL = True
```
deshraj: Lets have a limited set of allowed hostnames.

1. localhost
2. 127.0.0.1
3. ec2-54-242-225-49.compute-1.amazonaws.com

deshraj: Sure, go ahead.

PR #93, opened 2016-11-20
Sign Up page with API integration

aka-jain:
Done with registration form.


frontend/src/css/modules/base.scss
```diff
@@ -114,6 +140,239 @@ i.icon{
...
+    display: none;
+  }
+  input:checked + .info-text{
+    display: block;
+  }
+}
+
+
+
+
```
aka-jain: @deshraj  sure

PR #149, opened 2016-11-27
Setup ModelAdmin class for each model in Challenge app

gauthamzz:


apps/challenges/admin.py
```diff
@@ -1,3 +1,28 @@
...
+                   "enable_forum",
+                   "anonymous_leaderboard")
+    search_fields = ("title", "creator")
+
+
+@admin.register(Phase)
+class PhaseModelAdmin(TimeStampedAdmin):
+    list_display = ("name", "leaderboard_public", "challenge")
+    list_filter = ("leaderboard_public", "challenge")
+    search_fields = ("name")
```
TroJan: search_fields needs to be a tuple or list. In python when creating a tuple with single element, it is necessary to insert comma at the last, like

```
search_fields=('name',)
````

Comment (gauthamzz) [inline apps/challenges/admin.py]: done

**Comment by gauthamzz** [comment]
fixed @TroJan take a look

**Comment by TroJan** [comment]
@gauthamzz : the diff is fine. can you please squash these commits into a single one. Also for the commit message we generally prefer that you namespace it like here it can be `challenges: Setup Admin class for each model.`

File: apps/challenges/admin.py [hunk 2]
```diff
@@ -1,3 +1,19 @@
 from django.contrib import admin
 
-# Register your models here.
+from base.admin import TimeStampedAdmin
+
+from .models import Challenge, Phase
+
+
+@admin.register(Challenge)
+class ChallengeModelAdmin(TimeStampedAdmin):
```
TroJan: can you please rename it `ChallengeAdmin`

apps/challenges/admin.py
```diff
@@ -1,3 +1,19 @@
...
+    list_display = ("title", "start_date", "end_date", "creator", "published", "enable_forum", "anonymous_leaderboard")
+    list_filter = ("creator", "published", "enable_forum", "anonymous_leaderboard")
+    search_fields = ("title", "creator")
+
+
+@admin.register(Phase)
+class PhaseModelAdmin(TimeStampedAdmin):
```
TroJan: can you rename it `PhaseAdmin`

gauthamzz:
@TroJan and @deshraj  can you review this

PR #166, opened 2016-11-28
Added autofocus to input fields

jainarunarunjain:
Just added autofocus attribute to input field. Please have a look.


aka-jain:
Great. 
@deshraj : I think we have to make a seperate directive for autofocus. But for now it is OK. Can we open a new issue for that?

PR #195, opened 2016-12-01
"created_by" field added to ParticipantTeam model

deshraj:


deshraj:
Closing it since fixed in #207

PR #214, opened 2016-12-02
added autofxiture

anweshknayak:


anweshknayak:
@deshraj please review

Issue #243, opened 2016-12-05
Change color of radio buttons to EvalAI's primary color (Dark Gray)


aka-jain:
Change color of radio buttons in Create Team page.
Refer variable.scss file for color themes. Use $dark-gray for the radio button. Currently it is of green color and needs to be changed.

jainarunarunjain:
@aka-jain Can i work on this?

aka-jain:
@jainarunarunjain sure.

aka-jain:
Closing and fixed with #248

PR #244, opened 2016-12-06
Dark gray radio button

anweshknayak:


anweshknayak:
@aka-jain is the team page showing up as of now after running gulp dev and connect ?

aka-jain:
In addition of then run npm install and bower install. Then you are good to go.

anweshknayak:
#243

PR #259, opened 2016-12-11
Added improved version of view with loader

aka-jain:
@deshraj : have a look please!


frontend/src/views/web/partials/sim-loader
```diff
@@ -0,0 +1,12 @@
...
+<div class="text-white center w-300 loader-title">{{loaderTitle}}</div>
+    <div class="loader">
+
+      <div></div>
+      <div></div>
+      <div></div>
+      <div></div>
+      <div></div>
+    </div>
+</div>
```
deshraj: @aka-jain Can you add a newline here? :stuck_out_tongue:

PR #265, opened 2016-12-18
Challenge: Get Past Challenges API Endpoint added

deshraj:


deshraj:
@TroJan can you please review this?

apps/challenges/views.py
```diff
@@ -132,3 +133,17 @@ def disable_challenge(request, pk):
         response_data = {
             'error': 'Sorry, you do not have permission to disable this challenge'}
         return Response(response_data, status=status.HTTP_401_UNAUTHORIZED)
+
+
+@api_view(['GET'])
+def get_past_challenges(request):
```
TroJan: `permission_classes` and `authentication_classes` are missing. Any specific reason?

tests/unit/challenges/test_views.py
```diff
@@ -1,14 +1,18 @@
...
 from allauth.account.models import EmailAddress
+from datetime import timedelta
 from rest_framework import status
 from rest_framework.test import APITestCase, APIClient
 
 from challenges.models import Challenge
 from participants.models import ParticipantTeam
 from hosts.models import ChallengeHostTeam
 
+import json
```
TroJan: this should be at the top of file.

tests/unit/challenges/test_views.py
```diff
@@ -334,3 +338,79 @@ def test_when_user_does_not_have_permission_to_disable_particular_challenge(self
...
+                    "created_by": self.challenge3.creator.created_by.pk,
+                },
+                "published": self.challenge3.published,
+                "enable_forum": self.challenge3.enable_forum,
+                "anonymous_leaderboard": self.challenge3.anonymous_leaderboard,
+            }
+        ]
+        response = self.client.get(self.url, {}, format='json')
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertJSONEqual(json.dumps(
```
TroJan: Any specific reason for using assestJSONEqual over assertEqual ?

tests/unit/challenges/test_views.py
```diff
@@ -1,14 +1,18 @@
 from django.core.urlresolvers import reverse_lazy
 from django.contrib.auth.models import User
+from django.utils import timezone
 
 from allauth.account.models import EmailAddress
+from datetime import timedelta
```
TroJan: import this above django imports

apps/challenges/views.py
deshraj: I think the challenge details should be visible to everyone so that we can index these pages. Does that make sense?

tests/unit/challenges/test_views.py
deshraj: Sure. My bad.

tests/unit/challenges/test_views.py
deshraj: Nope. Reverting it back.

Issue #256, opened 2016-12-10
Invite user to participant team api should be idempotent.


TroJan:
`invite_participant_to_team` creates a Participant object. If this api is called N number of times for a single email id, N participant objects will be created.

Later on this api will involve sending a mail to the concerned participant, which implies that this can be called N number of times(email not received due to network failure or some error).

So ideal behaviour for this api should be create Participant object once and from next time return that created object. 

Django has a function called [get_or_create](https://docs.djangoproject.com/en/1.10/ref/models/querysets/#get-or-create) which can be used here. Hook will go in `save` of the concerned serialiser

@deshraj : what say?

deshraj:
Yeah, cool. Sounds like a plan. :+1:

gauthamzz:
is this assigned ? @TroJan

TroJan:
@gauthamzz : yeah you can work on it. thanks :)

gauthamzz:
1. Im facing some issue testing the module. im getting an error  ` "detail": "Authentication credentials were not provided." ` when using the http://127.0.0.1:8000/api/participants/participant_team 
2. the url should be changed to api/participants/ participant_team/(?P<participant_team_pk>[0-9]+)/invite/(?P<participant_pk>[0-9]+)$ [name='invite_participant_to_team'] as we need both team id and participant id .. am i correct ?

TroJan:
@gauthamzz : Lets solve your issue step by step. Taking Step 1 into consideration

If you see the error message, it indicates that you are trying to access an api without providing authentication credentials. So lets us first see from where this message is coming from

If you see [here](https://github.com/Cloud-CV/EvalAI/blob/master/apps/participants/views.py#L92), then it says `TokenAuthentication` as `authentication_classes`. Error message is due this check.
Now comes how to get rid of this error message. You will need to hit this api with a valid user token. The token can be generated by you and used as a Header.

To generate token, first of you will need to find your user id(open Django admin and look into __Users__ table under __Authentication and Authorization__). Afterwards you can use this snippet to get the token

```
from django.contrib.auth.models import User
from rest_framework.authtoken.models import Token
u=User.objects.get(id=<your_user_id>)
token = Token.objects.create(user=u)
print token.key
```

You can run this snippet by firing up a django shell like

```
python manage.py shell --settings=settings.dev
```

Afterwards you use the following curl to check this api

```
[...]

PR #271, opened 2016-12-19
Invite user to participant team api should be idempotent.

gauthamzz:
#256 is solved. @TroJan is this okay ?

### DIFF
--- a/apps/participants/serializers.py
+++ b/apps/participants/serializers.py
@@ -44,5 +44,5 @@ class InviteParticipantToTeamSerializer(serializers.Serializer):
     def save(self):
         email = self.validated_data.get('email')
         return Participant.objects.get_or_create(user=User.objects.get(email=email),
-                                          status=Participant.ACCEPTED,
-                                          team=self.participant_team)
+                                                status=Participant.ACCEPTED,
+                                                team=self.participant_team)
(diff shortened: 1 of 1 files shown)


apps/participants/serializers.py
```diff
@@ -43,6 +43,6 @@ def validate_email(self, value):
 
     def save(self):
         email = self.validated_data.get('email')
-        return Participant.objects.create(user=User.objects.get(email=email),
-                                          status=Participant.ACCEPTED,
-                                          team=self.participant_team)
+        return Participant.objects.get_or_create(user=User.objects.get(email=email),
+                                                 status=Participant.ACCEPTED,
+                                                 team=self.participant_team)
```
deshraj: @gauthamzz can you please remove the extra spaces that you have added in line 47 and 48?

gauthamzz: But that's for PEP8 (tabs over spaces) . should i remove them . then travis shows error

deshraj: We prefer using spaces all over the codebase. So, please make sure that you also use spaces. :-)

If it contains spaces, then it's fine.

gauthamzz: so is there any need to change anything as of now . @deshraj just put those to align them correctly

deshraj:
I would prefer to remove the spaces in line 47 and 48 but it's just a minor change. You don't need to change it and we are good to go for now. I will merge it. 

Thanks :-)

PR #276, opened 2016-12-22
Frontend: Reset Password API integrated with frontend

deshraj:


aka-jain:
Looks good @deshraj : You can merge it :)

Issue #288, opened 2016-12-24
Create Apis for Test Environment model


TroJan:
Following apis should be created
* get list of all test environments for a particular challenge
* create a new test environment
* get details about a particular test environment
* update(partial also) a particular test environment
* delete specific test environment

dexter1691:
@TroJan - This issue is resolved completely? 
Please feel free to close this issue if it's complete.

TroJan:
For reference, Test Environment model is now ChallengePhase

PR #289, opened 2016-12-24
Challenge: Add api to fetch list of test environments.

TroJan:
Issue #288


deshraj:
@TroJan can you check https://coveralls.io/builds/9425658/source?filename=apps%2Fchallenges%2Fserializers.py#L33

It looks like 2 lines are untested from the tests. Can you please add a test case for the same?

TroJan:
@deshraj: updated the PR.

PR #301, opened 2016-12-28
[Do Not Merge] Docker: Added docker containers for easy deployment

deshraj:


docker/nodejs/Dockerfile
```diff
@@ -0,0 +1,57 @@
+FROM nginx:latest
+
+########### Nginx Configuration starts here ###########
```
dexter1691: why is there an nginx configuration in nodejs? 

Shouldn't this be a separate container with nginx linking both node.js and django specific containers.

deshraj: Yeah, that's the plan. The PR is not fully ready. I pushed so that I don't loose the code and get comments from others.

dexter1691: SG.

deshraj:
Closing this as this is duplicate as #323

Issue #297, opened 2016-12-26
Setup Throttling in drf backend


deshraj:

deshraj:
@RishabhJain2018 you showed interest in working on this. Are you still interested in working on this issue?

RishabhJain2018:
@deshraj Yeah sure.

RishabhJain2018:
@deshraj Please provide with some details and references on how to proceed with it.

dexter1691:
@RishabhJain2018:
[This resource](http://www.django-rest-framework.org/api-guide/throttling/#example) seems to be a good starting point to setup throttling. 

The motivation to setup throttling is to prevent the server from being over-burdened by certain API calls that might be resource intensive. Like the submission related endpoints.  A concrete example might be that we want to throttle the number of submissions a user can make to a particular competition per minute? So that one user doesn't keep retrying if the submission is failing. 

@deshraj: What else? Can we list them here for easy reference. 

The leaderboard and all the other things that an unsigned viewer can see will probably not update a lot and will be cached. So might be okay to not throttle those, right?

RishabhJain2018:
@dexter1691 Thank you for the resource.

yeah. It would be okay not to throttle the leaderboard. Moreover, I also want to ask about the throttling on login attempts, Do we also need to throttle that with a time limit ?

RishabhJain2018:
@deshraj @dexter1691 I also found a open source load testing tool http://locust.io/ . Please have a look. May be it would help us in some way.

RishabhJain2018:
@deshraj I think we should close this issue now?

deshraj:
Yup.

PR #305, opened 2016-12-28
Setup Throttling in drf backend.

RishabhJain2018:
#297 solved. @deshraj @dexter1691 Please review the PR.


deshraj:
Also, I think you forgot to set up the cache. For reference, see this http://www.django-rest-framework.org/api-guide/throttling/#setting-up-the-cache. We will be using Memcached backend for caching.

RishabhJain2018:
@deshraj okk sure, I wil setup the cache.

apps/challenges/views.py
```diff
@@ -19,6 +21,7 @@
 from .serializers import ChallengeSerializer, ChallengePhaseSerializer
 
 
+@throttle_classes([UserRateThrottle])
```
TroJan: can you please move this below `authentication_classes`. Idea is to follow alphabetically write down decorators. `authentication_classes` is definitely not in the order, but lets not do it for others.

Also can you make sure of this point for each and every place where `throttle_classes` decorator is used.

RishabhJain2018: @TroJan I have used `@throttle_classes` outside because the decorators work in "Inside out" manner. Now, if a user with 101th request(for throttling rate 100 requests/minute) comes then the request will be throttled initially, before checking for other decorators which can help us to save overhead on the server. Am I right ?

settings/dev.sample.py
```diff
@@ -33,3 +33,6 @@
     'apps': ['auth', 'accounts', 'analytics', 'base', 'challenges', 'hosts', 'jobs', 'participants', 'web'],
     'show_fields': True,
 }
+
+# Port number for the python-Memcached cache backend.
+CACHES['default']['LOCATION'] = '127.0.0.1:11211' # noqa: ignore=F405
```
deshraj: Please add a newline here.

settings/prod.py
```diff
@@ -53,3 +53,6 @@
 REST_FRAMEWORK_DOCS = {
     'HIDE_DOCS': True
 }
+
+# Port number for the python-memcached cache backend.
+CACHES['default']['LOCATION'] = os.environ.get("LOCATION", "") # noqa: ignore=F405
```
deshraj: Why do we need a `#noqa` here?

settings/test.py
```diff
@@ -15,3 +15,6 @@
 
 # E-Mail Settings
 EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
+
+# Port number for the python-Memcached cache backend.
+CACHES['default']['LOCATION'] = '127.0.0.1:11211' # noqa: ignore=F405
```
deshraj: Please add a newline here.

settings/dev.sample.py
RishabhJain2018: okk sure.

settings/prod.py
RishabhJain2018: It is to fix the travis issues in order to ignore the `*` imports. As we imported `CACHES` from `common.py` using the `*`.

settings/test.py
RishabhJain2018: yeah sure.

apps/challenges/views.py
deshraj: @RishabhJain2018 Yup.

settings/prod.py
deshraj: Sounds good.

deshraj:
@RishabhJain2018 Great work. :-)

PR #311, opened 2016-12-30
Dashboard footer

Ram81:


aka-jain:
@Ram81 : Thanks for helping... but this PR not looks good. Lots of unwanted files are there like main.scss~ , settings.py etc. Please make sure that you only push main files not their temporary one (like main.scss~).
Also don't push backened related files in frontend issue unless or otherwise it is mentioned.
I hope you got my points. Please make a new PR with having only needed files.
Let me know if you want more help :)

PR #323, opened 2017-01-02
Docker containers for Production Environment added

deshraj:


deshraj:
Note: 
- We have to add the environment variables in the Dockerrun.aws.json to set the configurations of RDS and other Email related Environment variables.

PR #330, opened 2017-01-04
Frontend: Create challenge view added

deshraj:


deshraj:
@aka-jain Can you use this branch and check if datetime picker is working properly or not.

aka-jain:
@deshraj : sure.

frontend/src/js/controllers/challengeCreate.js
```diff
@@ -0,0 +1,70 @@
...
+        vm.isValid = {};
+
+        // function to change password
+        vm.createChallenge = function() {
+            // vm.loginContainer = angular.element('.change-passowrd-card');
+
+            var parameters = {};
+            parameters.url = 'auth/password/change/';
+            parameters.method = 'POST';
+            parameters.data = {
```
aka-jain: @deshraj : why we are using change password in create challenge function?

deshraj: I had to change the data after integrating the datetimepicker in the view. Can you integrate the datetimepicker so that I can proceed? You can make changes to this branch only.

PR #332, opened 2017-01-05
[In-Progress] RabbitMQ: Worker scripts added

deshraj:


deshraj:
Already progressed in #376. So closing it.

Issue #338, opened 2017-01-05
Create an API endpoint for deleting a User from a Host Team


deshraj:
API endpoint: `/api/hosts/<hosts_team_id>/`
Method: `DELETE`

A user can request this API to remove himself from the host team. You can get the user from `request.user`. 

Please comment here for more details. Also, check other APIs to delete.

deshraj:
@RishabhJain2018 I have assigned this also since this is very similar to #336

RishabhJain2018:
okk sure.

PR #343, opened 2017-01-05
API endpoint for deleting a User from a Host Team.

RishabhJain2018:
#338 solved. @deshraj Please review the PR.

### DIFF
--- a/apps/hosts/urls.py
+++ b/apps/hosts/urls.py
@@ -12,6 +12,6 @@ urlpatterns = [
         name='get_challenge_host_list'),
     url(r'challenge_host_team/(?P<challenge_host_team_pk>[0-9]+)/challenge_host/(?P<pk>[0-9]+)$',
         views.challenge_host_detail, name='get_challenge_host_details'),
-    url(r'remove_challenge_host/(?P<challenge_host_team_pk>[0-9]+)$',
+    url(r'remove_self_from_challenge_host/(?P<challenge_host_team_pk>[0-9]+)$',
         views.remove_self_from_challenge_host_team, name='remove_self_from_challenge_host_team'),
 ]
--- a/apps/hosts/urls.py
+++ b/apps/hosts/urls.py
@@ -12,6 +12,6 @@ urlpatterns = [
         name='get_challenge_host_list'),
     url(r'challenge_host_team/(?P<challenge_host_team_pk>[0-9]+)/challenge_host/(?P<pk>[0-9]+)$',
         views.challenge_host_detail, name='get_challenge_host_details'),
-    url(r'(?P<challenge_host_team_pk>[0-9]+)$',
-        views.delete_challenge_host_from_team_by_self, name='delete_challenge_host_from_team_by_self'),
+    url(r'remove_challenge_host/(?P<challenge_host_team_pk>[0-9]+)$',
+        views.remove_self_from_challenge_host_team, name='remove_self_from_challenge_host_team'),
 ]
--- a/apps/hosts/urls.py
+++ b/apps/hosts/urls.py
@@ -13,6 +13,5 @@ urlpatterns = [
     url(r'challenge_host_team/(?P<challenge_host_team_pk>[0-9]+)/challenge_host/(?P<pk>[0-9]+)$',
         views.challenge_host_detail, name='get_challenge_host_details'),
     url(r'(?P<challenge_host_team_pk>[0-9]+)$',
-    	views.delete_challenge_host_from_team_by_self,
-    	name='delete_challenge_host_from_team_by_self'),
+        views.delete_challenge_host_from_team_by_self, name='delete_challenge_host_from_team_by_self'),
 ]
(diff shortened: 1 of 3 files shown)


apps/hosts/views.py
```diff
@@ -189,3 +189,27 @@ def create_challenge_host_team(request):
         challenge_host.save()
         return Response(response_data, status=status.HTTP_201_CREATED)
     return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+@throttle_classes([UserRateThrottle, ])
+@api_view(['DELETE', ])
+@permission_classes((permissions.IsAuthenticated, HasVerifiedEmail))
+@authentication_classes((ExpiringTokenAuthentication, ))
+def delete_challenge_host_from_team_by_self(request, challenge_host_team_pk):
```
TroJan: can you please change the view name to `remove_self_from_challenge_host_team`. this actually signifies that the api is trying to remove a challenge host from challenge host team instead of deleting it.

apps/hosts/views.py
```diff
@@ -189,3 +189,27 @@ def create_challenge_host_team(request):
...
+    """
+    A user can remove himself from the challenge host team.
+    """
+    try:
+        challenge_host = ChallengeHostTeam.objects.get(pk=challenge_host_team_pk)
```
TroJan: variable name should be `challenge_host_team`

apps/hosts/views.py
```diff
@@ -189,3 +189,27 @@ def create_challenge_host_team(request):
...
+    except ChallengeHostTeam.DoesNotExist:
+        response_data = {'error': 'ChallengeHostTeam does not exist'}
+        return Response(response_data, status=status.HTTP_406_NOT_ACCEPTABLE)
+
+    challenge_host = ChallengeHost.objects.get(user=request.user.id)
+    if challenge_host.team_name.id == int(challenge_host_team_pk):
```
TroJan: I guess that if you update the queryset, you will also realize that there is actually no need of this if condition over here. You can simply delete the queryset.

apps/hosts/urls.py
```diff
@@ -12,5 +12,6 @@
         name='get_challenge_host_list'),
     url(r'challenge_host_team/(?P<challenge_host_team_pk>[0-9]+)/challenge_host/(?P<pk>[0-9]+)$',
         views.challenge_host_detail, name='get_challenge_host_details'),
-
+    url(r'(?P<challenge_host_team_pk>[0-9]+)$',
```
TroJan: I guess you also need to add some other params in url so that the request comes to this. your url can be used to serve get/update/delete for a particular challenge host team. so you will explicitly need to mention something like remove

apps/hosts/views.py
RishabhJain2018: okk sure.

apps/hosts/views.py
RishabhJain2018: If I change the varibale name to `challenge_host_team` then the TravisCI build fails, as it is only used to check whether a team exists or not but `challenge_host_team` isn't used anywhere in the function.

deshraj: So @RishabhJain2018 you really don't need to assign it to a variable. Just write `ChallengeHostTeam.objects.get(pk=challenge_host_team_pk)`

apps/hosts/views.py
RishabhJain2018: @TroJan Consider a situation like if a user belongs to more than one challenge host team and the user want to delete himself from one team but not from the other , now if we don't check this condition then it will automatically delete both the instances.

apps/hosts/urls.py
RishabhJain2018: yeah sure.

apps/hosts/views.py
deshraj: Rishabh see this http://pastebin.com/gWyS3Thi. This is what @TroJan wants to say. Let me know if you have concerns about this.

RishabhJain2018:

tests/unit/hosts/test_views.py
```diff
@@ -347,3 +347,39 @@ def test_create_challenge_host_and_team_with_no_data(self):
         del self.data['team_name']
         response = self.client.post(self.url, self.data)
         self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+
+class DeleteChallengeHostFromTeamHimselfTest(BaseAPITestClass):
```
deshraj: Can you please rename this to `RemoveSelfFromChallengeHostTeamTest` to maintain consistency?

apps/hosts/urls.py
```diff
@@ -12,5 +12,6 @@
         name='get_challenge_host_list'),
     url(r'challenge_host_team/(?P<challenge_host_team_pk>[0-9]+)/challenge_host/(?P<pk>[0-9]+)$',
         views.challenge_host_detail, name='get_challenge_host_details'),
-
+    url(r'remove_challenge_host/(?P<challenge_host_team_pk>[0-9]+)$',
```
deshraj: Forgot to mention that please change the prefix of url to `remove_self_from_challenge_host_team`?

deshraj: Rest all looks good.

RishabhJain2018: okk sure.

PR #348, opened 2017-01-06
Frontend: Added capability to remove user(self) from Participant Team

deshraj:


deshraj:
screencapture-localhost-8888-web-teams-1483673274160

frontend/src/views/web/partials/sidebar.html
```diff
@@ -14,15 +14,15 @@
                 </li>
                 <li>
                     <a class="waves-effect waves-dark text-light-gray ev-md-container" ui-sref="web.teams" ui-sref-active="side-active-link">
-                        <i class="fa fa-users side-link-icon"></i> Teams</a>
+                        <i class="fa fa-users side-link-icon"></i> My Participant Teams</a>
                 </li>
                 <li>
                     <a class="waves-effect waves-dark text-light-gray ev-md-container" ui-sref="web.challenge-main" ui-sref-active="side-active-link">
-                        <i class="fa fa-fire side-link-icon"></i> Challenge Page</a>
+                        <i class="fa fa-fire side-link-icon"></i> Show all Challenges</a>
```
aka-jain: Can we use "All Challenges" instead of using "Show All Challenges"?

deshraj: Sure, will change it. :-)

frontend/src/views/web/partials/sidebar.html
```diff
@@ -14,15 +14,15 @@
                 </li>
                 <li>
                     <a class="waves-effect waves-dark text-light-gray ev-md-container" ui-sref="web.teams" ui-sref-active="side-active-link">
-                        <i class="fa fa-users side-link-icon"></i> Teams</a>
+                        <i class="fa fa-users side-link-icon"></i> My Participant Teams</a>
```
aka-jain: I think we were thinking to delete the Team page from sidebar, isn't it?

aka-jain:
@deshraj : I've leave some comments, have a look please.

frontend/src/views/web/partials/sidebar.html
deshraj: Sorry, my bad.

aka-jain: Or may be it's a My Participation Team section where only team in which a particular user is a member can only be listed here, is it?

deshraj: So currently it is a view to choose the team for a particular challenge. We have to make it independent where user can see the list of teams or just create a team. 

When user will try to participate in a challenge, then he will be shown a similar page from where he can select one team from list of teams or create a new team and then proceed.

aka-jain: Agree, that will be great :)

Issue #347, opened 2017-01-05
Create an API endpoint to get all the challenges hosted by the user


deshraj:
URL: /api/challenges/get_all_challenges_hosted_by_user/
Method: GET

RishabhJain2018:
@deshraj Can I work on this ?

deshraj:
Sure. Assigning it to you.

RishabhJain2018:
@deshraj As discussed with **TroJan** in #350 , please assign this to TroJan.

deshraj:
Fixed in #503

PR #350, opened 2017-01-06
API endpoint to get all the challenges hosted by the user.

RishabhJain2018:
#347 soved. @deshraj @TroJan Please review the PR.


TroJan:
@RishabhJain2018 : thanks for your work :)

Just that I discussed with @deshraj  and realized that this pr is actually not required. There is a view called `get_all_challenges` in `apps/challenges/views.py` which can be modified and tweaked a lot so that we can handle all these types of request over there. One of the main reason for not creating a new api is that it will become difficult to propogate any change(like any change in model fiels) across all api functions. Moreover it violates DRY.

As of now I and @deshraj  are not clear with the exact generic solution for this. Hence assigning it to myself

RishabhJain2018:
@TroJan No problem at all. I will close this PR. :-)

PR #351, opened 2017-01-06
Frontend: Added the functionality of inviting other users to participant teams

deshraj:


deshraj:
screencapture-localhost-8888-web-teams-1483740376933 1

Issue #319, opened 2017-01-02
Redesign 404 page for Clientside.


aka-jain:
Need to redesign a 404 page as shown in the given attachment. 
In case of any help or requirement, let me know.
(use evalai.cloudcv.org as a link to redirect user to homepage)
![eval-404](https://cloud.githubusercontent.com/assets/11537940/21596089/36beb6e8-d15c-11e6-9c44-095a7a6c050a.png)

aka-jain:
Suggestions are also welcome related to design.
One can also use some AI related image or robot image in place of owl because our platform is based on AI and it will be good if we go for the same (evalAi) :D :)
(make sure that image shouldn't contain any copyright issue)

Ram81:
@aka-jain Can You assign this to me ?

aka-jain:
@Ram81 : let PR #314 be merged, then you can start working on this.

aka-jain:
@Ram81: You can work on this now. Please be ready with the updated code

aka-jain:
@Ram81 : PFA of image of OWL.
![owl](https://cloud.githubusercontent.com/assets/11537940/21653385/d29e1646-d2d6-11e6-92ee-5954fd0ee795.png)

Ram81:
@aka-jain  I will add it. :+1:

aka-jain:
Closing it, fix with #359

PR #353, opened 2017-01-07
Fix: Redesign 404 page for Clientside (#319)

Ram81:


Ram81:
@aka-jain  do I need to change the message of '404' to 'Error 404' because I think in test for the page it is required to have 'Error 404' ,right?

aka-jain:
@Ram81 : No need to do it here, our 404 is design oriented. If needed, will handle it later.

Ram81:
why are the checks failing then,can you tell me?

PR #354, opened 2017-01-07
tests for urls

### LINKED ISSUE #333: Add tests for testing urls in accounts app.
The tests file must be located at `tests/unit/accounts/test_urls.py`. Create one, if it doesn't exists.
For reference please use http://stackoverflow.com/questions/18987051/how-do-i-unit-test-django-urls

anweshknayak:
fixes #333


anweshknayak:
@deshraj @RishabhJain2018 Is this the right way im doing ?

deshraj:
@anweshknayak can you please open a PR only if the code is ready? For this time, it's ok but please do that only if you want to discuss something about it. :-)

anweshknayak:
@deshraj (https://github.com/Cloud-CV/EvalAI/pull/354/commits/c2984f79bfd3d42dfa4a743f6982d817d49e211d) is this the right way for testing urls ? or I have to test using resolve attribute ?

RishabhJain2018:
@anweshknayak no, it is not the right way. Please refer to the reference link . You need to check two things: 
1). a).We want to have a hard coded URL path: as if a browser or a user had typed it in.
     b). We, as humans, know which URL pattern name we expect that to match to.
     c). We know what (keyword) arguments should be extracted from the URL string.

2). Also, to test that the URL connect to the right view, we can use `resolve`.

anweshknayak:
@RishabhJain2018 using resolve attribute it is throwing exception. 
Suppose url(r'^accounts/', include('allauth.urls')).
Now trying resolve('/accounts/') throws exception as such : ( http://pastebin.com/DXQWE6TY)

RishabhJain2018:
@anweshknayak The `django-allauth` is already a [tested ](https://github.com/pennersr/django-allauth)application. so, I don't think that we should write tests for it. Moreover, for the use of `resolve`, please refer this [link](https://docs.djangoproject.com/en/1.10/ref/urlresolvers/#resolve).

anweshknayak:
@deshraj review this please

tests/unit/accounts/test_urls.py
```diff
@@ -0,0 +1,32 @@
+from django.urls import reverse_lazy
+from rest_framework import status
+from rest_framework.test import APITestCase, APIClient
+from allauth.account.models import EmailAddress
+from django.contrib.auth.models import User
+import unittest
```
deshraj: Do we need this?

deshraj: Also, please check the PEP8 styling for imports here https://docs.djangoproject.com/en/dev/internals/contributing/writing-code/coding-style/#python-style

anweshknayak: done. @deshraj

deshraj: The order of imports is still not correct. Can you please refer to the link that I shared?

anweshknayak: ohh you meant the order. sorry I was confused

deshraj: Yes. :-)

anweshknayak: @deshraj I ordered them alphabetically as said in the pep guidelines.

Issue #324, opened 2017-01-03
Change `created_by` field from integer to username in `api/hosts/challenge_host_team/` API endpoint


deshraj:

gauthamzz:
can i try?

deshraj:
Sure. Go ahead.

deshraj:
@gauthamzz are you working on it? If yes, can you please do it quicker since we need to fix it soon.

gauthamzz:
@deshraj I'm trying but something is wrong with testing it . cos of proxy . ill try to do it fast

deshraj:
Oh wait, I know this error. We recently integrated memcache backend with the codebase so you hace to include those settings. Please re-copy the dev.sample.py to dev.py since there are new changes in there.

gauthamzz:
Its working now . Can you give me a hint how to solve this ?

deshraj:
Can you tell me that what was the issue?

gauthamzz:
the created by field was returning user id instead of username

gauthamzz:
the created by field was returning user id instead of username @deshraj .
{"count":5,"next":null,"previous":null,"results":[{"id":1,"team_name":"Gautham","created_by":1},{"id":4,"team_name":"gfffff","created_by":1},{"id":5,"team_name":"dhggggggggg","created_by":1},{"id":6,"team_name":"sssssssssssssssss","created_by":1},{"id":7,"team_name":"adssssssssssss","created_by":1}]}%

needed to be changed to

{"count":5,"next":null,"previous":null,"results":[{"id":1,"team_name":"Gautham","created_by":"gautham"},{"id":4,"team_name":"gfffff","created_by":"gautham"},{"id":5,"team_name":"dhggggggggg","created_by":"gautham"},{"id":6,"team_name":"sssssssssssssssss","created_by":"gautham"},{"id":7,"team_name":"adssssssssssss","created_by":"gautham"}]}%

gauthamzz:
This was how it was supposed to look 

'default': {
'ENGINE': 'django.db.backends.postgresql_psycopg2',
'NAME': 'evalai',
'USER': 'your username',
'PASSWORD': 'your password',
'HOST': 'localhost',
'PORT': '5432',
}

but the field password is not there and field NAME is not equal to EVALAI @deshraj

gauthamzz:
close this ?

deshraj:
Thanks @gauthamzz  :-)

PR #358, opened 2017-01-07
Change `created_by` field from integer to username in `api/hosts/challenge_host_team/` API endpoint

gauthamzz:
#324 is fixed @deshraj

### DIFF
--- a/apps/challenges/views.py
+++ b/apps/challenges/views.py
@@ -33,7 +33,7 @@ def challenge_list(request, challenge_host_team_pk):
         return Response(response_data, status=status.HTTP_406_NOT_ACCEPTABLE)
 
     if request.method == 'GET':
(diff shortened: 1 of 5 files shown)


gauthamzz:
@deshraj is it okay if change the tests or will that be another pr ?

deshraj:
It should be in the same PR. :-)

gauthamzz:
@deshraj can you tell me if i am going on the right path . I am confused :-1:

deshraj:
@gauthamzz I don't think you need to change so much. Please see this https://github.com/Cloud-CV/EvalAI/blob/master/apps/participants/serializers.py#L18. It will make this task much simpler.

gauthamzz:
Thanks :+1: I was completely lost

Issue #325, opened 2017-01-03
Add a check for ChallengeHostTeam Ownership in Challenge Creation API


deshraj:
In challenge creation API (https://github.com/Cloud-CV/EvalAI/blob/master/apps/challenges/views.py#L28), we should add a check if the `<challenge_host_team_pk>` belongs to the user making the request.

gauthamzz:
I think it's related to #324 . So is it okay if I try both

deshraj:
Both are independent but if you are willing to solve both of these then you are good to go.

deshraj:
Also, please make sure that you open two different PRs for both of these issues.

gauthamzz:
Okay . I'll try .

gauthamzz:
can you help me with something when i try 
`curl -X  POST -H "Authorization: Token id" "http://localhost:8000/api/challenges/challenge_host_team/1/challenge" `
i get this error message. I was not getting earlier. 

`<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN"
"http://www.w3.org/TR/html4/loose.dtd">
<html>
<head>
<meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
<title>Notification: Gateway Timeout</title>

<style type="text/css">
body {
  font-family: Arial, Helvetica, sans-serif;
  font-size: 14px;
  color:#333333;
  background-color: #ffffff;
}
h1 {
  font-size: 18px;
  font-weight: bold;
  text-decoration: none;
  padding-top: 0px;
  color: #2970A6;
}
a:link {
    color: #2970A6;
  text-decoration: none;
}
a:hover {
    color: #2970A6;
  text-decoration: underline;
}
p.buttonlink {
  margin-bottom: 24px;
}
.copyright {
  font-size: 12px;
  color: #666666;
  margin: 5px 5px 0px 30px;

}
.details {
  font-size: 14px;
  color: #969696;
  border: none;
  padding: 20px 20px 20px 20px;
  margin: 0px 10px 10px 35px;
}

.shadow {
  border: 3px solid #9f9f9f;
  padding: 10px 25px 10px 25px;
  margin: 10px 35px 0px 30px;
  background-color: #ffffff;
  width: 600px;

  -moz-box-shadow: 3px 3px 3px #cccccc;
  -webkit-box-shadow: 3px 3px 3px #cccccc;
  box-shadow: 3px 3px 3px #cccccc;
[...]

PR #360, opened 2017-01-07
Add a check for ChallengeHostTeam Ownership in Challenge Creation API

gauthamzz:
#325 is fixed

### DIFF
--- a/apps/challenges/views.py
+++ b/apps/challenges/views.py
@@ -32,11 +32,6 @@ def challenge_list(request, challenge_host_team_pk):
         response_data = {'error': 'ChallengeHostTeam does not exist'}
         return Response(response_data, status=status.HTTP_406_NOT_ACCEPTABLE)
 
-    if challenge_host_team.created_by != request.user:
-        response_data = {
-            'error': 'Sorry, you do not have permission to this challenge'}
-        return Response(response_data, status=status.HTTP_401_UNAUTHORIZED)
-
     if request.method == 'GET':
         challenge = Challenge.objects.filter(creator=challenge_host_team)
         paginator = PageNumberPagination()
@@ -47,6 +42,12 @@ def challenge_list(request, challenge_host_team_pk):
         return paginator.get_paginated_response(response_data)
 
     elif request.method == 'POST':
+
+        if challenge_host_team.created_by != request.user:
+            response_data = {
+                'error': 'Sorry, you do not have permission to this challenge'}
+            return Response(response_data, status=status.HTTP_401_UNAUTHORIZED)
+
         serializer = ChallengeSerializer(data=request.data,
                                          context={'challenge_host_team': challenge_host_team})
         if serializer.is_valid():
--- a/apps/challenges/views.py
+++ b/apps/challenges/views.py
@@ -43,7 +43,7 @@ def challenge_list(request, challenge_host_team_pk):
 
     elif request.method == 'POST':
 
-        if challenge_host_team.created_by != request.user:
+        if ChallengeHost.objects.filter(user=request.user).filter(team_name=challenge_host_team_pk):
             response_data = {
                 'error': 'Sorry, you do not belong to this Host Team!'}
             return Response(response_data, status=status.HTTP_401_UNAUTHORIZED)
--- a/apps/challenges/views.py
+++ b/apps/challenges/views.py
@@ -43,7 +43,7 @@ def challenge_list(request, challenge_host_team_pk):
 
     elif request.method == 'POST':
 
-        if not ChallengeHost.objects.filter(user=request.user).filter(team_name=challenge_host_team_pk):
+        if not ChallengeHost.objects.filter(user=request.user, team_name_id=challenge_host_team_pk).exists():
             response_data = {
                 'error': 'Sorry, you do not belong to this Host Team!'}
             return Response(response_data, status=status.HTTP_401_UNAUTHORIZED)
--- a/apps/challenges/views.py
+++ b/apps/challenges/views.py
@@ -45,7 +45,7 @@ def challenge_list(request, challenge_host_team_pk):
 
         if challenge_host_team.created_by != request.user:
             response_data = {
-                'error': 'Sorry, you do not have permission to this challenge'}
+                'error': 'Sorry, you do not belong to this Host Team!'}
             return Response(response_data, status=status.HTTP_401_UNAUTHORIZED)
 
         serializer = ChallengeSerializer(data=request.data,
--- a/apps/challenges/views.py
+++ b/apps/challenges/views.py
@@ -43,7 +43,7 @@ def challenge_list(request, challenge_host_team_pk):
 
     elif request.method == 'POST':
 
-        if ChallengeHost.objects.filter(user=request.user).filter(team_name=challenge_host_team_pk):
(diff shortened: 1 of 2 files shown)


apps/challenges/views.py
```diff
@@ -32,6 +32,11 @@ def challenge_list(request, challenge_host_team_pk):
         response_data = {'error': 'ChallengeHostTeam does not exist'}
         return Response(response_data, status=status.HTTP_406_NOT_ACCEPTABLE)
 
+    if challenge_host_team.created_by != request.user:
+        response_data = {
+            'error': 'Sorry, you do not have permission to this challenge'}
+        return Response(response_data, status=status.HTTP_401_UNAUTHORIZED)
+
```
deshraj: @gauthamzz can you please add this check in the `POST` method condition?

gauthamzz: Its done :+1: @deshraj

apps/challenges/views.py
```diff
@@ -42,6 +42,12 @@ def challenge_list(request, challenge_host_team_pk):
         return paginator.get_paginated_response(response_data)
 
     elif request.method == 'POST':
+
+        if challenge_host_team.created_by != request.user:
```
deshraj: @gauthamzz I was thinking deeply about this and it seems like this won't cover all the use cases. For example, if a host team consists of 4 members then only the member who has created the team can only create a challenge using that team but ideally what should happen is that all of the members of that team should have the right to create a team? Does that makes sense? Let me know if you have concerns.

apps/challenges/views.py
```diff
@@ -42,6 +42,12 @@ def challenge_list(request, challenge_host_team_pk):
...
+            response_data = {
+                'error': 'Sorry, you do not have permission to this challenge'}
```
deshraj: Change the response to `Sorry, you do not belong to this Host Team!`

apps/challenges/views.py
gauthamzz: Yeah that makes more sense  :+1:  .

gauthamzz: @deshraj i have made some changes , will this work ?

gauthamzz:
@deshraj please review the PR

apps/challenges/views.py
gauthamzz: @deshraj i made a check if any object is returned from challengehostteam under the team name and user name . will that be okay ?

deshraj:
@gauthamzz please add a test case for the same when you are ready.

gauthamzz:
can you give me any hints ?

gauthamzz:
@deshraj can you help with travis

tests/unit/challenges/test_views.py
```diff
@@ -127,6 +128,11 @@ def test_create_challenge_with_no_data(self):
         response = self.client.post(self.url, self.data)
         self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
 
+    def test_check_challenge_host_team_ownership(self):
+        del self.date['team_name']
+        response = self.client.post(self.url, self.data)
+        self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
```
deshraj: @gauthamzz see this https://travis-ci.org/Cloud-CV/EvalAI/builds/189953817#L776

There is a whitespace at the end of this line. Please remove it and it should work fine.

apps/challenges/views.py
```diff
@@ -42,6 +42,12 @@ def challenge_list(request, challenge_host_team_pk):
         return paginator.get_paginated_response(response_data)
 
     elif request.method == 'POST':
+
+        if not ChallengeHost.objects.filter(user=request.user).filter(team_name=challenge_host_team_pk):
```
gauthamzz: @deshraj how do i check this in tests

tests/unit/challenges/test_views.py
```diff
@@ -115,7 +115,8 @@ def setUp(self):
             'submission_guidelines': 'Submission guidelines for new test challenge',
             'published': False,
             'enable_forum': True,
-            'anonymous_leaderboard': False
+            'anonymous_leaderboard': False,
+            'team_name': 'Test Challenge Host Team'
```
gauthamzz: @deshraj is putting team name like this wrong? i mean its giving error at line 129 and 124

tests/unit/challenges/test_views.py
gauthamzz: @deshraj its still failing . how do i fix https://travis-ci.org/Cloud-CV/EvalAI/builds/189963015#L876

apps/challenges/views.py
TroJan: i think this condition also needs to be updated. chaining filter is not bad, but it is not required here.
you can simply pass both conditions in a single `filter`

also you cannot compare `team_name` with `challenge_host_team_pk` because `team_name` is a challenge host team instance where as `challenge_host_team_pk` is integer. So you can either convert it to `team_name=challenge_host_team` or `team_name_pk=challenge_host_team_pk`. I will prefer the latter one,

one more thing which i want to point out is you can use `exists` to check if there is any row. For more reference, you can see this [link](http://stackoverflow.com/a/6869696/2534102)

gauthamzz: i think this has another problem too .I think it doesn't check for the condition in which the creator is there. I mean creator doesn't automatically get added to the team in `challengehost`. it is only in challenge table. creator will have an object in `challenge` only . so i should check both right ? @TroJan

TroJan: we do create challenge host object when ever any new challenge host team is created. you can have a look at api `create_challenge_host_team` in `hosts/views.py`

gauthamzz: @TroJan okay , can you tell me how to write tests for that .any hints

TroJan:
@gauthamzz : hint for writing test lies in your diff only. Lets call condition `user=request.user, team_name_id=challenge_host_team_pk` as __A__. The piece of code that you have written can only be tested if your `if` condition becomes true which means that there should be no `ChallengeHost` object matching condition __A__. 

Now comes the point what to modify or pass in __A__, so that it overalls returns empty set. `challenge_host_team_pk` cant be updated here, so the only thing that you can do is request this api with a user id which is not a part of challenge host team.

Eg: Take two users with id 1 and 3.
User #1 creates a Challenge Host team with id __45__. User #3 is __not__ a part of CHT #__45__, but user #3 requests the api with id of challenge host team #__45__ to create a challenge.

I hope this helps :)

deshraj:
@jainarunarunjain I think you have to modify the path here https://travis-ci.org/Cloud-CV/EvalAI/builds/190035273#L907

deshraj:
@gauthamzz do you need some help here? Please do let me know. :-)

gauthamzz:
Yeah, I'm stuck

deshraj:
Please do let us know every time when you get stuck. We would be more than happy to help. :-)

gauthamzz:
I didn't quite get what @TroJan said . I tried lot of different things ,idk what's wrong . Btw it's my first time with tests :( @deshraj

deshraj:
@gauthamzz please see the #397. I have fixed the issue that you were facing. So basically you were not creating the challenge host team and that was causing the problem. 

Closing it now since work is carried forward to #397. 

Thanks for working on this issue. :-)

PR #376, opened 2017-01-08
[In progress] Add worker to process submission and load new challenges.

TroJan:


submission_worker.py
```diff
@@ -0,0 +1,300 @@
...
+
+    phase_data_base_directory = join(challenge_data_directory, 'phase_data')
+
+    for phase in phases:
+        phase_data_directory = join(phase_data_base_directory, str(phase.id))
+        # create phase directory
+        create_dir(phase_data_directory)
+        phase_zip_file = join(phase_data_directory, 'phase_{}.zip'.format(phase.id))
+        annotation_file_url = phase.test_annotation
+        download_and_extract_file(annotation_file_url, phase_zip_file, phase_data_directory)
```
dexter1691: I think the annotation file will just be a single file containing the ground truth. In the current setup, it won't be a zip file. 
@deshraj : please confirm?

submission_worker.py
```diff
@@ -0,0 +1,300 @@
...
+    except Exception as e:
+        print 'Failed to fetch file from {}, error {}'.format(url, e)
+        # return from here or not?
+        response = None
+
+    if response and response.status_code == 200:
+        with open(download_location, 'w') as f:
+            f.write(response.content)
+        # extract zip file
+        zip_ref = zipfile.ZipFile(download_location, 'r')
```
dexter1691: Is there a possibility that the content of the response is not a zip file? Are we putting enough checks in place to ensure that this is always a zip file?

submission_worker.py
```diff
@@ -0,0 +1,300 @@
...
+        submission = Submission.objects.get(id=submission_id)
+    except Submission.DoesNotExist:
+        print 'Submission {} does not exist'.format(submission_id)
+
+    submission_input_file = submission.input_file
+    submission_data_directory = SUBMISSION_DATA_DIR.format(submission_id=submission.id)
+    # create submission directory
+    create_dir(submission_data_directory)
+
+    submission_zip_file = join(submission_data_directory, 'submission_{}.zip'.format(submission.id))
```
dexter1691: same doubt as above. Are the submission files just one prediction file or it is a zip?

submission_worker.py
```diff
@@ -0,0 +1,300 @@
...
+        submission = None
+
+    stdout_file = join(submission_data_dir, stdout_file_name)
+    stderr_file = join(submission_data_dir, stderr_file_name)
+
+    stdout = open(stdout_file, 'a+')
+    stderr = open(stderr_file, 'a+')
+
+    for prog_cmd_counter, prog_cmd in enumerate(prog_cmd_list):
+        prog_cmd = prog_cmd.replace("$program", evaluation_script_path) \
```
dexter1691: Program should be replaced by the executable call, right? It looks like this will just contain the path to the executable but not the actual executable. 

@deshraj : Are we capturing this information somewhere in the UI? Or are we assuming a standard name to the file. For ex: "python evaluate.py $input $output" ?

dexter1691:
@TroJan : Seems like a pretty solid start! I left comments to where I had some doubts or places which might require further discussion! Please feel free to ignore the ones that don't make sense. 

How are we planning to let the main server know that the execution is finished? How will we update the clients with the results?

submission_worker.py
deshraj: @TroJan we have to make sure that all the block of code is inside try and catch to catch all the exceptions.

submission_worker.py
deshraj: Just one file.

submission_worker.py
deshraj: Yeah, we have to modify this @TroJan

submission_worker.py
dexter1691: These exceptions will be tricky to handle! These issues will need to be notified to the user! 
Will need to think of a way to let these exceptions percolate to the client. (Websockets seems to be the only reasonable way)

submission_worker.py
deshraj: Currently, we are planning to have a single script with `evaluate.py` name.

submission_worker.py
dexter1691: Ideally, when the user is uploading this zip file, we should make sure that the contents belong to a zip file, and reject any other file type inside that REST API call.

deshraj: Oh yes, good idea. I will add it in the list of checks that I thought of for the submission API.

submission_worker.py
dexter1691: Okay. So the idea is to have one evaluate.py file which accepts two arguments?  The code should then probably reflect that.

TroJan:
@dexter1691 @deshraj : basic version is almost ready, right now TLE (pool of worker threads) is not handled. can you both please have a look?

submission_worker.py
```diff
@@ -0,0 +1,309 @@
...
+    stderr = open(stderr_file, 'a+')
+
+    # call `main` from globals
+    with stdout_redirect(stdout) as new_stdout, stderr_redirect(stderr) as new_stderr:      # noqa
+        EVALUATION_SCRIPTS[challenge_id].main(phase_data_dir, user_annotation_file_path)
+    stderr.close()
+    stdout.close()
+    stderr_content = open(stderr_file, 'r').read()
+    stdout_content = open(stdout_file, 'r').read()
+    with open(stdout_file, 'r') as stdout:
```
dexter1691: This is so beautiful!!!

submission_worker.py
```diff
@@ -0,0 +1,309 @@
...
+        submission = None
+
+    stdout_file = join(temp_run_dir, stdout_file_name)
+    stderr_file = join(temp_run_dir, stderr_file_name)
+
+    stdout = open(stdout_file, 'a+')
+    stderr = open(stderr_file, 'a+')
+
+    # call `main` from globals
+    with stdout_redirect(stdout) as new_stdout, stderr_redirect(stderr) as new_stderr:      # noqa
```
dexter1691: This is so beautiful.

dexter1691:
I am guessing we will add the submission finished notification functionality in another PR? 
LGTM.  I will let deshraj go through it once and approve.

submission_worker.py
```diff
@@ -0,0 +1,309 @@
...
+    q_params['start_date__lt'] = timezone.now()
+    q_params['end_date__gt'] = timezone.now()
+
+    # make sure that the challenge base directory exists
+    create_dir_as_python_package(CHALLENGE_DATA_BASE_DIR)
+
+    active_challenges = Challenge.objects.filter(**q_params)
+
+    for challenge in active_challenges:
+        phases = challenge.challengephase_set.all()
```
deshraj: I am not sure about `challengephase_set`. Can you please tell me about this?

submission_worker.py
```diff
@@ -0,0 +1,309 @@
...
+    connection = pika.BlockingConnection(pika.ConnectionParameters(
+        host='localhost'))
+
+    channel = connection.channel()
+    channel.exchange_declare(exchange='evalai_submissions', type='topic')
+
+    # name can be a combination of hostname + process id
+    # host name : to easily identify that the worker is running on which instance
+    # process id : to add uniqueness in case more than one worker is running on the same instance
+    add_challenge_queue_name = socket.gethostname() + '_' + str(os.getpid())
```
deshraj: Can we use string formatting here?

submission_worker.py
```diff
@@ -0,0 +1,309 @@
...
+import requests
+import shutil
+import signal
+import socket
+import sys
+import time
+import yaml
+import zipfile
+
+from os.path import join
```
TroJan: `join` is used many a time, importing here to avoid writing `os.path.join` again and again.

submission_worker.py
```diff
@@ -0,0 +1,309 @@
...
+from subprocess import Popen, call
+
+from django.conf import settings
+from django.utils import timezone
+from django.core.files.base import ContentFile
+
```
TroJan: sure

submission_worker.py
```diff
@@ -0,0 +1,309 @@
...
+def create_dir_as_python_package(directory):
+    '''
+        Create a directory and then makes it a python
+        package by creating `__init__.py` file.
+    '''
+    create_dir(directory)
+    init_file_path = join(directory, '__init__.py')
+    with open(init_file_path, 'w') as init_file:        # noqa
+        # create empty file
+        pass
```
TroJan: to create an empty file.

submission_worker.py
```diff
@@ -0,0 +1,309 @@
...
+def download_and_extract_file(url, download_location, extract_location):
+    '''
+        * Function to extract downlaod a zip file, extract it and then removes the zip file.
+        * `download_location` should include name of file as well.
+    '''
+    try:
+        response = requests.get(url)
+    except Exception as e:
+        print 'Failed to fetch file from {}, error {}'.format(url, e)
+        # return from here or not?
```
TroJan: yeah, my bad, i forgot to remove.

submission_worker.py
TroJan: ChallengePhase does Foreign Key on Challenge, with no related_name, hence when fetching phase for a particular challenge I need to use `challengephase_set`.

submission_worker.py
```diff
@@ -0,0 +1,309 @@
...
+    # hard coded, should be fetched from challenge model
+    # execution_time_limit = 5 * 60
+
+    stdout_file_name = 'temp_stdout.txt'
+    stderr_file_name = 'temp_stderr.txt'
+
+    try:
+        submission = Submission.objects.get(id=submission_id)
+    except Submission.DoesNotExist:
+        submission = None
```
TroJan: Message should only be in queue only if its submission object has been saved. Message will be sent by an internal api, hence i think we can handle this case in the api which is responsible for creating and sending message.

TroJan:
@deshraj : updated the pr, have a look  :)

submission_worker.py
deshraj: Can we not add a `related_name` there? Would that be a problem?

submission_worker.py
deshraj: I see. Yeah, this sound like a plan to me.

deshraj:
@TroJan before merging it into master, can we move the submission_worker.py into `scripts/worker` direcotry?

PR #384, opened 2017-01-08
Correct path for custom 404 and 500 error pages

jainarunarunjain:
@deshraj Please review this PR


settings/common.py
```diff
@@ -83,7 +83,7 @@
 
 ROOT_URLCONF = 'evalai.urls'
 
-TEMPLATE_PATH = os.path.join(BASE_DIR, 'frontend')
+TEMPLATE_PATH = os.path.join(BASE_DIR, 'frontend/dist')
```
deshraj: Can you please rewrite this as `TEMPLATE_PATH = os.path.join(BASE_DIR, 'frontend', 'dist')`

jainarunarunjain: @deshraj sure

jainarunarunjain:
@deshraj Can you please tell me why this build still fail.

deshraj:
You have to change the path here https://github.com/Cloud-CV/EvalAI/blob/master/evalai/views.py#L5

deshraj:
Also, you can run tests on your local machine using the command `py.test`. So, there is no need to push and then wait for this to complete.

PR #389, opened 2017-01-09
Challenges: classmethods added to Challenge and ChallengePhase Model

deshraj:


apps/challenges/models.py
```diff
@@ -54,3 +87,22 @@ class ChallengePhase(TimeStampedModel):
...
+    def get_end_date(self):
+        """Returns the end date of Challenge"""
+        return self.end_date
+
+    @property
+    def is_active(self):
+        """Returns if the challenge is active or not"""
+        if self.end_date > timezone.now():
+            return True
+        return False
```
dexter1691: are we planning to put get methods for each field? 
If yes, then he getter method list seems incomplete. 
If not, then we should probably get rid of all get functions, right?

dexter1691: @deshraj: Can we add a check that end_date should always be greater than start_date? 
Also write a corresponding test case for that!

tests/unit/challenges/test_models.py
```diff
@@ -0,0 +1,142 @@
...
+    def setUp(self):
+        super(ChalengeTestCase, self).setUp()
+
+        try:
+            os.makedirs('/tmp/evalai')
+        except OSError:
+            pass
+
+        with self.settings(MEDIA_ROOT='/tmp/evalai'):
+            self.challenge.image = SimpleUploadedFile('test_sample_file.jpg',
```
dexter1691: Probably more rigorous tests here would be nice! For example: 
Can we check if random characters return an error or not? 
Can we check if the length of the URL is okay? 

We should probably put a check for folder name being not too long! and image name being not too long!

dexter1691: ^Those are probably for views.py though. So feel free to ignore these for models.py and make sure we include them in other tests.

apps/challenges/models.py
deshraj: No, we are not planning to have get method for each field. We will have methods for those fields which we need to use somewhere and some other methods that will devise some metric using the fields already present such as `is_active()`. 

Sure, I will add a test for that.

Issue #387, opened 2017-01-08
Create the job submission API


deshraj:

RishabhJain2018:
@deshraj I think we should close this?

deshraj:
Fixed in #396. Closing it.

PR #396, opened 2017-01-10
Jobs: Added API endpoint for creating submission to a Challenge Phase

deshraj:


apps/jobs/models.py
```diff
@@ -36,8 +40,10 @@ class Submission(TimeStampedModel):
         (FINISHED, FINISHED),
     )
 
-    participant_team = models.ForeignKey(ParticipantTeam, related_name='submissions')
-    challenge_phase = models.ForeignKey(ChallengePhase, related_name='submissions')
+    participant_team = models.ForeignKey(
```
dexter1691: can we make this user? That will store the participant team information as well, right? 
It will require one extra join though to query submission per team. Might be good to add both then? 

Saying this because then we will loose out information regarding which user submitted. Not super relevant info but still.

apps/jobs/urls.py
```diff
@@ -1,6 +1,9 @@
 from django.conf.urls import url
-# from .views import ()
 
+from . import views
 
 urlpatterns = [
+    url(r'challenge/(?P<challange_id>[0-9]+)/'
+        r'challenge_phase/(?P<challange_phase_id>[0-9]+)/',
```
dexter1691: Shouldn't this be appended with a submission end_point?

dexter1691: How do you update / delete / get the submission given submission id

apps/jobs/views.py
```diff
@@ -1,3 +1,82 @@
...
+
+from accounts.permissions import HasVerifiedEmail
+
+from participants.utils import (
+    get_participant_team_id_of_a_user_for_a_challenge,
+    check_user_participated_in_challenge,)
+
+from challenges.models import (
+    ChallengePhase,
+    Challenge,)
```
dexter1691: extra comma?

apps/jobs/views.py
```diff
@@ -1,3 +1,82 @@
...
+from participants.models import (ParticipantTeam, Participant)
```
dexter1691: or no extra comma?

apps/participants/utils.py
```diff
@@ -0,0 +1,28 @@
...
+def check_user_participated_in_challenge(user, challenge_id):
+    """Returns boolean if the user has participated in a particular challenge"""
+    participant_teams = Participant.object.filter(user=user).values_list('teams', flat=True)
+    return Challenge.objects.filter(pk=challenge_id, participant_teams__in=participant_teams).exists()
+
+
+def get_participant_team_id_of_a_user_for_a_challenge(user, challenge_id):
+    """Returns the participant team object for a particular user for a particular challenge"""
+    participant_teams = Participant.object.filter(user=user).values_list('teams', flat=True)
+    for participant_team in participant_teams:
```
dexter1691: There should be a better way to do this using joins.

apps/participants/utils.py
```diff
@@ -0,0 +1,28 @@
...
+
+def check_participant_team_participated_in_challenge(participant_team_id, challenge_id):
+    """Returns boolean if the Participant Team participated in particular Challenge"""
+    return Challenge.objects.filter(pk=challenge_id, participant_team__id=participant_team_id).exists()
+
+
+def check_user_participated_in_challenge(user, challenge_id):
+    """Returns boolean if the user has participated in a particular challenge"""
+    participant_teams = Participant.object.filter(user=user).values_list('teams', flat=True)
+    return Challenge.objects.filter(pk=challenge_id, participant_teams__in=participant_teams).exists()
```
dexter1691: This should essentially use the information from get_participant_team_id_of_a_user_for_a_challenge function. 
If participant_team is provided, use the info otherwise call the function to get that info based on user. 
Should also rename these functions to make them concise.

dexter1691:
I am assuming this is gonna handle only the file upload part! actual update of the submission status will be done through channels?

apps/participants/utils.py
deshraj: Yup, I am figuring that out.

apps/jobs/models.py
```diff
@@ -58,3 +64,57 @@ def __unicode__(self):
...
+                submissions_done_today_count = Submission.objects.filter(
+                    challenge_phase__challenge=self.challenge_phase.challenge,
+                    participant_team=self.participant_team,
+                    challenge_phase=self.challenge_phase,
+                    submitted_at__gte=datetime.date.today()).count()
+
+                if ((submissions_done_today_count + 1 - failed_count > self.challenge_phase.max_submissions_per_day) or
+                        (self.challenge_phase.max_submissions_per_day == 0)):
+                    print 'PERMISSION DENIED'
+                    raise PermissionDenied(
```
dexter1691: we are not handling this exception in the views.py L78 I guess. We should handle this exception gracefully and send a proper error message.

apps/jobs/views.py
deshraj: This is the convention that we should ideally follow and we have been maintaining this throughout the project.

dexter1691: Not in this file though. There are a couple of places where there is no comma. Right?

deshraj: Sure, will look into it.

apps/jobs/models.py
```diff
@@ -58,3 +64,57 @@ def __unicode__(self):
...
+                self.submission_number = sub_num + 1
+            else:
+                self.submission_number = 1
+
+            failed_count = Submission.objects.filter(
+                challenge_phase=self.challenge_phase,
+                participant_team=self.participant_team,
+                status=Submission.FAILED).count()
+
+            offset_submission_count = self.submission_number - failed_count
```
dexter1691: can we rename this to successful_count?
Also we are going to use success_count more, so might be better to explicitly save that and not save failed_count?

apps/jobs/urls.py
deshraj: Yeah, we will add that functionality soon. This PR mainly focuses on creating the Submission  only since this is a major feature.

apps/jobs/models.py
deshraj: Cool I will add another field in the model named `submitted_by` to store the information of the user and that will not affect the business logic. I will just need to store an additional field.

apps/jobs/models.py
deshraj: Sure.

apps/jobs/models.py
deshraj: I am not sure what kind of exception we can get there? Can you please elaborate more?

apps/jobs/urls.py
dexter1691: Okay, but it still feels odd to not have a submission end_point appended after challenge_phase_id. The post request should be sent to /challenge/<challenge_id>/challenge_phase<phase_id>/submission right?

deshraj: So, since it is in the `jobs` app, so the url would be `/api/jobs/challenge/<challenge_id>/challenge_phase<phase_id>/submission`. Does that sound good?

dexter1691: sg

apps/jobs/models.py
dexter1691: If permission is denied then on save, it will throw an error, right? This I guess will call 500 Internal error. We should catch this exception and throw a 401 error saying that max number of submissions have been made.

apps/jobs/views.py
```diff
@@ -1,3 +1,82 @@
...
+                                       permission_classes,
+                                       throttle_classes,)
+from rest_framework.response import Response
+from rest_framework_expiring_authtoken.authentication import (
+    ExpiringTokenAuthentication,)
+from rest_framework.throttling import UserRateThrottle
+
+from accounts.permissions import HasVerifiedEmail
+
+from participants.utils import (
```
TroJan: can you please reorder imports alphabetically ?

apps/jobs/views.py
```diff
@@ -1,3 +1,82 @@
...
+@throttle_classes([UserRateThrottle])
+@api_view(['POST'])
+@permission_classes((permissions.IsAuthenticated, HasVerifiedEmail))
+@authentication_classes((ExpiringTokenAuthentication,))
+def challenge_submission(request, challenge_id, challenge_phase_id):
+    """API Endpoint for making a submission to a challenge"""
+    # check if the challenge exists or not
+    try:
+        challenge = Challenge.objects.get(pk=challenge_id)
+        if not challenge.is_active():
```
TroJan: what if take `is_active` check out of `try-catch`

apps/jobs/views.py
```diff
@@ -1,3 +1,82 @@
...
+    # check if the challenge phase exists or not
+    try:
+        challenge_phase = ChallengePhase.objects.get(
+            pk=challenge_phase_id, challenge=challenge)
+    except ChallengePhase.DoesNotExist:
+        response_data = {'error': 'Challenge Phase does not exist!'}
+        return Response(response_data, status=status.HTTP_401_UNAUTHORIZED)
+
+    # check if challenge is public and accepting solutions
+    if not challenge.is_public:
```
TroJan: why querying for phase, when challenge is not active, should be after `challenge.is_public` check

apps/jobs/models.py
deshraj: So, we are not storing the successful or failure counts since we can calculate that by seeing the status of the submissions made by the user.

apps/jobs/models.py
deshraj: No, it will give a 403 response so we are good to go with this.

apps/jobs/models.py
```diff
@@ -58,3 +64,57 @@ def __unicode__(self):
...
+            if hasattr(self.challenge_phase, 'max_submissions_per_day'):
+                submissions_done_today_count = Submission.objects.filter(
+                    challenge_phase__challenge=self.challenge_phase.challenge,
+                    participant_team=self.participant_team,
+                    challenge_phase=self.challenge_phase,
+                    submitted_at__gte=datetime.date.today()).count()
+
+                if ((submissions_done_today_count + 1 - failed_count > self.challenge_phase.max_submissions_per_day) or
+                        (self.challenge_phase.max_submissions_per_day == 0)):
+                    print 'PERMISSION DENIED'
```
TroJan: also this print statement can be made more meaningful, so that it becomes easier for us to debug any issue(in case if it happens).
One more thing which I am curious about is why are we not using python logging module?

apps/jobs/models.py
```diff
@@ -58,3 +64,57 @@ def __unicode__(self):
...
+                    raise PermissionDenied(
+                        "The maximum number of submissions this day have been reached.")
```
TroJan: I think it should be `The maximum number of submissions that can be made today has reached` or `The maximum number of submission for today has been reached`.

deshraj:
Fixed #387

apps/jobs/models.py
```diff
@@ -58,3 +64,57 @@ def __unicode__(self):
...
+            else:
```
TroJan: I really didn't get this else here. What is it for? Can you explain me more clearly?

apps/jobs/models.py
deshraj: Yeah we will replace all the debugging statements with logging once the pr is ready.

apps/jobs/models.py
deshraj: I just realized that it's not required here according to our use case. I will remove it.

apps/jobs/models.py
dexter1691: Hmm. Ok for now but minor point is the stack trace will miss the actual endpoint that is called. Can you plz add a simple /TODO  just so that we remember to check.

deshraj:
@TroJan can you please review the PR again? I have added the tests.

apps/participants/utils.py
```diff
@@ -0,0 +1,28 @@
...
+    return Challenge.objects.filter(pk=challenge_id, participant_team__id=participant_team_id).exists()
+
+
+def has_user_participated_in_challenge(user, challenge_id):
+    """Returns boolean if the user has participated in a particular challenge"""
+    participant_teams = Participant.objects.filter(user=user).values_list('teams', flat=True)
+    return Challenge.objects.filter(pk=challenge_id, participant_teams__in=participant_teams).exists()
+
+
+def get_participant_team_id_of_user_for_a_challenge(user, challenge_id):
```
TroJan: see if [this](http://stackoverflow.com/a/2364277/2534102) helps here ?

apps/jobs/views.py
```diff
@@ -1,3 +1,87 @@
...
+@api_view(['POST'])
+@permission_classes((permissions.IsAuthenticated, HasVerifiedEmail))
+@authentication_classes((ExpiringTokenAuthentication,))
+def challenge_submission(request, challenge_id, challenge_phase_id):
+    """API Endpoint for making a submission to a challenge"""
+    # check if the challenge exists or not
+    try:
+        challenge = Challenge.objects.get(pk=challenge_id)
+    except Challenge.DoesNotExist:
+        response_data = {'error': 'Challenge does not exist!'}
```
deshraj: Lets remove the exclamation mark. We have been following that convention so far.

apps/jobs/views.py
```diff
@@ -1 +1,74 @@
...
+        response_data = {
+            'error': 'Sorry, cannot accept submissions since challenge phase is not public!'}
+        return Response(response_data, status=status.HTTP_406_NOT_ACCEPTABLE)
+
+    participant_team_id = get_participant_team_id_of_user_for_a_challenge(
+        request.user, challenge_id)
+    try:
+        participant_team = ParticipantTeam.objects.get(pk=participant_team_id)
+    except ParticipantTeam.DoesNotExist:
+        response_data = {'error': 'You haven\'t participated in the challenge'}
```
deshraj: I have used single quotes since we have been following this convention so far.

apps/participants/utils.py
deshraj: I am pretty sure that there is a better way of doing this. Will look into it. For now, I think we are good with this.

deshraj:
@TroJan can you please check and merge it so that we can start integrating the worker? w

TroJan:
@deshraj : looks good, can be merged now

Issue #390, opened 2017-01-09
Increase the code quality in landscape. 


RishabhJain2018:
@deshraj I found the code quality is low in landscape.io . Please refer to this [link](https://landscape.io/github/Cloud-CV/EvalAI/547/messages/smell).

Can I work on this to fix this??

deshraj:
Oh this looks like a good initiative. Do you want to work on this? I was going to open an issue for the same.

RishabhJain2018:
@deshraj yeah sure.

souravsingh:
@deshraj I am interested in working on the issue if no one else is working

deshraj:
@souravsingh: Sorry, @RishabhJain2018 is already working on this.

RishabhJain2018:
@deshraj I think we should close this?

deshraj:
Thanks. :-)

Issue #327, opened 2017-01-04
Add Docstring to all the API Endpoints


deshraj:

souravsingh:
@deshraj I would like to work on the issue.

deshraj:
@souravsingh Sorry, this one is already assigned to @TroJan. Feel free to pick any unassigned issue. :-)

TroJan:
@souravsingh : you can work on this, I am little busy with other issues.
@deshraj : please re assign it.

deshraj:
Oh Ok. @souravsingh If you are interested then you can work on this issue. :-)

dexter1691:
@souravsingh: Status update? Are you still working on this issue?

souravsingh:
@dexter1691 I am working on it

deshraj:
@souravsingh any updates on this? I think it has been really long time since you have been working on this. If you are not working on this, then please let me know so that I can assign it to someone else.

souravsingh:
@deshraj I got caught up with some commitments. I will finish this up right away.

deshraj:
@souravsingh looks like you are busy and hence I am unassigning this issue.

souravsingh:
@deshraj I almost forgot the issue. I will send a WIP PR right away.

deshraj:
Oh Ok. Thanks for letting me know. :-)

souravsingh:
I apologize for keeping this pending

deshraj:
Looks like @souravsingh is not able to work on this. Anyone interested in working on this issue is welcome. :)

RishabhJain2018:
@deshraj Can I work on this?

deshraj:
Sure. Assigning it to you.

gauthamzz:
is anyone working on this ?

RishabhJain2018:
@gauthamzz No, I am not working on this but @souravsingh is working on this.

deshraj:
Closing this for now. Will reopen it once there comes a requirement.

PR #398, opened 2017-01-10
Fixes Landscape configuration.

RishabhJain2018:
#390 solved. @deshraj Please review the PR.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,10 +1,20 @@
 [MESSAGES CONTROL]
 # Disable Messages
-disable=C1001,R0903,W0613,W0702,E1101,D100,C0330,W0223,C0111,W0232,C0103,R0201
+# C0103: Invalid %s name "%s".
+# C0111: Missing %s docstring.
+# C0330: Regarding hanging indentation.
+# C1001: Old-style class defined.
+# D100:  Missing docstring in public module.
+# E1101: %s %r has no %r member.
+# R0201: Method could be a function.
+# R0903: Too few public methods.
+# W0223: Method %r is abstract in class %r but is not overridden.
+# W0232:  Class has no __init__ method.
+# W0613: Unused argument.
+# W0702: No exception type(s) specified.
+disable=C0103,C0111,C0330,C1001,D100,E1101,R0201,R0903,W0223,W0232,W0613,W0702
 
 [FORMAT]
-
-# Maximum number of characters on a single line.
 max-line-length=120
 
 [DESIGN]
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,13 @@
+[MESSAGES CONTROL]
+# Disable Messages
+disable=C1001,R0903,W0613,W0702,E1101,D100,C0330,W0223,C0111,W0232,C0103,R0201
+
+[FORMAT]
+
+# Maximum number of characters on a single line.
+max-line-length=120
+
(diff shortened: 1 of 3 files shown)


RishabhJain2018:
@deshraj For now, I have ignored the `docstrings` in all the files. When the issue #327 is solved, then please enable it from the `.landscape.yml` and `.pylintrc` files.

TroJan:
@RishabhJain2018 : what if we break this pr per app, problem is we usually avoid so much code changes(this involves more than 15 files) in a single app.

Doing changes per app will make it neat, easy for us to review also

RishabhJain2018:
@TroJan yeah sure, we can do that. Moreover, I would like to suggest that If within this PR only, I will remove all the present commits & push the per app basis commits. Will it do any good?

dexter1691:
@TroJan: Thanks!!!
Can we please break these down into smaller more meaningful commits!

RishabhJain2018:
@dexter1691 yeah sure. I am breaking down this PR into smaller per app basis commits.

RishabhJain2018:
@dexter1691 @TroJan Please review the PR.

deshraj:
@RishabhJain2018 can you please divide this PR into different PRs per app?

deshraj:
Also, do update your fork before pushing.

RishabhJain2018:
@deshraj yeah sure.

.pylintrc
```diff
@@ -0,0 +1,407 @@
+[MASTER]
```
deshraj: @RishabhJain2018 is this file is really required here?

RishabhJain2018: @deshraj yeah, This file is for the pylint configuration. Like if we have to disable a pylint warning on complete project then we can do it using `disable` option within this file. Moreover, it is to make pylint checks according to our requirements, like `max-line-length` [here](https://github.com/Cloud-CV/EvalAI/pull/398/files#diff-d6d3d73911fd3ad6f39e4027f763b307R291) is set to `120`, whereas the default is `80`. There are many other features also which can be enabled or disabled.

deshraj: I think we are already using `flake8`. So, this would be just a duplicate thing to use for the same purpose.

RishabhJain2018: @deshraj Please have a look at the comparison of `flake8` and `pylint` [here](https://python.libhunt.com/project/pylint/vs/flake8). Pylint is a good, handy, convenient and wonderful tool to be used.

.pylintrc
```diff
@@ -0,0 +1,13 @@
+[MESSAGES CONTROL]
+# Disable Messages
+disable=C1001,R0903,W0613,W0702,E1101,D100,C0330,W0223,C0111,W0232,C0103,R0201
```
deshraj: @RishabhJain2018 can you add comment to show what all these mean?

.pylintrc
```diff
@@ -0,0 +1,13 @@
...
+
+[FORMAT]
+
+# Maximum number of characters on a single line.
```
deshraj: Please remove this line.

.pylintrc
RishabhJain2018: yeah sure.

.pylintrc
RishabhJain2018: okk

deshraj:
Looks good. Great work. :-)

PR #373, opened 2017-01-08
is_public field in Phase model created.

RishabhJain2018:
#364 solved with migrations added. @deshraj Please review the PR.

PR #403, opened 2017-01-10
Challenges: upload_to path fixed for test_annotation field

deshraj:
This PR corrects the mistake done in PR #373 where the path of `test_annotations` was modified. Undoing that in this PR.


deshraj:
Merging this PR since this is blocking from building another feature.

Issue #406, opened 2017-01-11
Add a new status option in Submission model


deshraj:
For reference, see this https://github.com/Cloud-CV/EvalAI/blob/master/apps/jobs/models.py#L25

We want to add a new status option with name `SUBMITTING = "submitting"`

Ram81:
@deshraj  as I am not working on any issue currently can I work on this?

deshraj:
Sure, go ahead. Assigning it to you. :-)

Ram81:
@deshraj  I just have to add the option name submitting & add it in status_Options in the Submission model ,am I right ? or do I have to change any other code?

deshraj:
You would also need to add an entry similar to this https://github.com/Cloud-CV/EvalAI/blob/master/apps/jobs/models.py#L32

Ram81:
Okay Understood @deshraj will do it & let you know?

deshraj:
Sounds good.

Ram81:
@deshraj  sorry for this I will just remove directive.js from it !

TroJan:
@deshraj : please make sure that migrations(named please) are also there in the PR.

deshraj:
Sure, will take of it.

TroJan:
@Ram81 : thanks for the quick fix :)

PR #407, opened 2017-01-11
Fix: Add a new status option in Submission model (#406)

Ram81:


frontend/src/js/directives/directives.js
```diff
@@ -289,4 +289,4 @@
         }
```
TroJan: can you please remove this file ?

Issue #439, opened 2017-01-14
Profile Section is not fixed when page is scrolled


Ram81:
When we open My Profile link from dashboard & try to scroll the position of profile section isn't fixed.

Ram81:
@aka-jain I will work on this for now.

aka-jain:
@Ram81 : Sure.

Ram81:
@aka-jain  should I make the profile section fixed like dashboard-sidebar when opened or some other change if you want?

aka-jain:
Yes, give it a fixed position for now, and integrate it with inner scroll in Porfile section.

Ram81:
"integrate it with inner scroll in Porfile section." didn't quite get that?

aka-jain:
@Ram81 : Make profile section scrollable in case when screen's height becomes less then the height of the content in Profile Bar.

Ram81:
@aka-jain  should I Integrate inner scroll to profile section in this issue or do you want me to close this issue & wait for next version of the issue?

PR #440, opened 2017-01-14
Fix: Profile Section is not fixed when scrolled(#439)

Ram81:
I've added a fixed position to it if any changes please suggest @aka-jain


aka-jain:
@Ram81 : I'm merging this for now. Will fix it in another version. Thanks for the efforts :)

PR #451, opened 2017-01-15
Jobs: Add publisher to send submission message.

TroJan:
@deshraj @dexter1691 : please review

### DIFF
--- a/apps/jobs/sender.py
+++ b/apps/jobs/sender.py
@@ -1,0 +1,29 @@
+import json
+import pika
+
+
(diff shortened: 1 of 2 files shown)


TroJan:
things that i will be updating after this PR is merged is to move all the connection parameters(also exchange and queue names) out of the codebase to settings.

One of the thing that concerns me here is that for each submission message a new connection will be created, Ideally we could have created a single persistent connection and over it all the message would have been sent. But for v1 as discussed with @deshraj , we are going with this approach only.

PR #454, opened 2017-01-15
correct path for custom error pages

jainarunarunjain:

### DIFF
--- a/evalai/views.py
+++ b/evalai/views.py
@@ -2,14 +2,14 @@ from django.shortcuts import render
 
 
 def page_not_found(request):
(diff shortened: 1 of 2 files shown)


deshraj:
@jainarunarunjain Are the errors still there on the terminal now?

jainarunarunjain:
@deshraj No,  there is no error in terminal

deshraj:
So, you are getting travis build error because travis doesn't run `gulp dev` and hence there is not `dist` directory. As discussed before, lets create another 404 and 500 page for django in `evalai/templates`.

jainarunarunjain:
@deshraj Sure

PR #455, opened 2017-01-15
EvalAI: URL Pattern for passwrod reset confirm modified

deshraj:


deshraj:
Modifying the URL Pattern since the email sent for password reset has to be in that format where `auth` is the parent for all the authentication related stuff.

TroJan:
looks good to me

PR #456, opened 2017-01-15
Settings: MEDIA ROOT added for storing media files

deshraj:


.gitignore
```diff
@@ -82,3 +82,7 @@ bower_components/
 
 # Challenge data files
 challenge_data/*
+
+# Media Files
+media/
```
TroJan: can you explain why is this directory ignored ?

settings/dev.sample.py
```diff
@@ -39,3 +39,5 @@
         'BACKEND': 'django.core.cache.backends.dummy.DummyCache',
     }
 }
+
+MEDIA_URL = "/media/"
```
TroJan: what about other settings?

.gitignore
deshraj: For development, all the media files will go in this directory and hence we don't want anyone to add the media directory to staging area.

settings/dev.sample.py
deshraj: I don't think we need any other settings. Can you please elaborate a bit more?

Issue #457, opened 2017-01-15
Add support for getting all challenges in `get_all_challenge` api.


TroJan:
value of `challenge_time` can be either of `ALL`, `PAST`, `PRESET` or `FUTURE`.

deshraj:
Fixed in #458

PR #458, opened 2017-01-15
Challenge: Add support to get all challenge in get_all_challenge api.

TroJan:
Now get_all_challenge accpets either of ALL, FUTURE, PAST or PRESENT
as url arguments to return corresponding challenges.

Issue #457


deshraj:
Looks Good.

PR #460, opened 2017-01-15
Urls: Add Media root and url.

TroJan:


deshraj:
Closing since already fixed in #514

PR #479, opened 2017-01-16
Adding Sentry and File Logging for production.

dexter1691:
Log all debug data to a file because storage is cheap.
Log only errors on Sentry because only 10,000 event per month for free :(


deshraj:
LGTM.

PR #493, opened 2017-01-17
Create Participant Serializer to show the Participant details

jayantsa:


dexter1691:
@jayantsa : Can you please fix the pep8 errors. 
To see those errors run flake8. Look at the travis CI details to see the command and the error messages.

deshraj:
@jayantsa please see  the errors here https://travis-ci.org/Cloud-CV/EvalAI/builds/192543422

Let me know if you don't understand some part.

jayantsa:
@deshraj  Please Review and suggest to pass coverage checks

apps/participants/serializers.py
```diff
@@ -49,6 +49,22 @@ def save(self):
...
 
 
+class ParticipantSerializer(serializers.ModelSerializer):
+
+    member_name = serializers.SerializerMethodField()
+    member_id = serializers.SerializerMethodField()
+
+    class Meta:
+        model = Participant
+        fields = ('member_id', 'member_name')
```
deshraj: You should include all the fields in the `Participant` model.

apps/participants/serializers.py
```diff
@@ -49,6 +49,22 @@ def save(self):
...
+
+    def get_member_name(self, obj):
```
deshraj: After including all the fields, we won't need this method.

apps/participants/serializers.py
```diff
@@ -49,6 +49,22 @@ def save(self):
...
+        return obj.user.username
+
+    def get_member_id(self, obj):
```
deshraj: Same for this method.

apps/participants/urls.py
```diff
@@ -17,4 +17,7 @@
     url(r'remove_self_from_participant_team/(?P<participant_team_pk>[0-9]+)$',
         views.remove_self_from_participant_team,
         name='remove_self_from_participant_team'),
+    url(r'participants_list_based_on_team/(?P<participant_team_pk>[0-9]+)$',
```
deshraj: Please change `participants_list_based_on_team` to `participants_list_based_on_participant_team`

apps/participants/serializers.py
jayantsa: @deshraj 
If we include all the fields in Participant model by:
fields=('user','status'.'team')
then serializer returns just id of user and team as output

deshraj:
@jayantsa I realized that we don't need an endpoint for this. What we can do is that create a `ParticipantSerializer` and then inject it in the `ParticipantTeamSerializer` and it will do the work. Otherwise we would have to make two different API calls which doesn't make sense. 

Please let me know if I am not clear.

jayantsa:
@deshraj Please Review. 
Unable to change **api/participants/participants_list_based_on_team/<pk>** 
to  **api/participants/participants_list_based_on_participant_team/1**  
as it is directing to ParticipantTeamDetails . Please Check I think it's an issue

deshraj:
@jayantsa I think you missed my comment.

jayantsa:
@deshraj ok ,I will make those changes

jayantsa:
@deshraj If we inject ParticipantSerializer in ParticipantTeamSerializer, I think it will create an issue when Post request is made from from frontend at the time of creation of Team because at that time are not adding Participant to teams

deshraj:
I see. You are correct. What we can do is create a new serializer that is combination of ParticipantSerializer and ParticipantTeamSerializer and use it in the endpoint where we want to show the details of Team and its participants. Does that make sense?

jayantsa:
@deshraj Please Review . Added Serializer and now **api/participants/participant_team/1**
gives details of team with its members on GET request

apps/participants/serializers.py
```diff
@@ -49,6 +49,38 @@ def save(self):
...
+        return obj.user.id
+
+
+# Serializer for Participant Teams and Participant Combined
+class ParticipantsBasedOnParticipantTeamSerializer(serializers.ModelSerializer):
```
deshraj: Lets rename the serializer to `ParticipantTeamDetailSerializer `

jayantsa:
@deshraj Changed Name of Serializer. Please review

deshraj:
Great work @jayantsa :-)

PR #495, opened 2017-01-17
Challenges: Fixes the case when challenge to paritcipate mapping is M2M

deshraj:
Currently, one case was missing in this endpoint i.e when the participant team has participated in multiple challenges. The API returned only one challenge for a participant team but the mapping is many to many.


deshraj:
@dexter1691 @TroJan Please review

deshraj:
Merging it for now since this is a blocking the submission feature on frontend.

PR #498, opened 2017-01-17
Challenge: Add api to get challenge given participant team id or challenge host team id.

TroJan:


apps/challenges/urls.py
```diff
@@ -21,5 +21,7 @@
         name='get_all_challenges'),
     url(r'challenge/(?P<pk>[0-9]+)/',
         views.get_challenge_by_pk, name='get_challenge_by_pk'),
+    url(r'challenge$', views.get_challenges_based_on_teams,
```
deshraj: Lets remove `challenge`

TroJan: I have kept challenge intentionally, so that till this commit all apis are like

```
* challenges/challenge/:challenge_id
* challenges/challenge/present
* challenges/challenge?participant_team=1
```

TroJan: afterwards we can remove challenge in one commit if `challenges/challenge` is redundant.

deshraj: Yeah, exactly thats what I meant. But sure, we can do it later.

deshraj:
Overall looks good to me. Merging it.

PR #500, opened 2017-01-17
Participants: Update get participant team ids function.

TroJan:
Creates a new function which returns participant team ids for a particular user


TroJan:
@deshraj : please review.

deshraj:
Looks good.

Issue #511, opened 2017-01-17
Show the Backend error messages from submission API


deshraj:
Currently we are showing the hard coded response when we get an error from the backend on submission. We want to show the exact backend error messages on the client side.

RishabhJain2018:
@deshraj Can I work on this issue?

deshraj:
Sure sure. Assigning it to you. :-)

deshraj:
Fixed in #520

PR #516, opened 2017-01-18
Backend Error Messages: Submission Error Backend Messages Shown.

RishabhJain2018:
#511 solved. @deshraj @aka-jain @spyshiv Please review the PR.

### DIFF
--- a/frontend/src/js/controllers/challengeCtrl.js
+++ b/frontend/src/js/controllers/challengeCtrl.js
@@ -207,7 +207,7 @@
                                     onError: function(response) {
                                         var status = response.status;
                                         var error = response.data;
-                                        utilities.storeData('emailError', error.detail);
+                                        utilities.storeData('emailError', error.error);
                                         $state.go('web.permission-denied');
                                         utilities.hideLoader();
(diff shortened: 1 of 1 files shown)


aka-jain:
@RishabhJain2018 I think this has to be done on backside to submission API not on frontend side.
@deshraj Am I right? It would be good if you give a brief here.

deshraj:
@aka-jain So the issue was to show the error messages thrown from the backend instead of showing the hard coded message on frontend if some error occurs during the submission.

RishabhJain2018:
@deshraj @aka-jain The error was to access the dictionary's item which is coming in response to frontend from backend. So, we can do both ways either change the `key` in the api in backend or change the response object's `key` in frontend. So, what should we do ?

aka-jain:
@RishabhJain2018 , Wait I'm working on your PR.

aka-jain:
@RishabhJain2018 : I've to close this PR. Have to do major changes here. Thanks for the efforts :)

RishabhJain2018:
@aka-jain No problem. :-)

Issue #522, opened 2017-01-18
Change name of memcached location environment Variable


deshraj:

deshraj:
We want to change the name of the environment variable from `LOCATION` to `MEMCACHED_LOCATION`

RishabhJain2018:
@deshraj Can I work on this issue?

deshraj:
Sure, if you want to work on this, then please do it quickly. You need to make changes in `settings/common.py` and `settings/prod.py`

deshraj:
Fixed in #523

PR #523, opened 2017-01-18
Name of memcached location environment Variable Modified.

RishabhJain2018:
#522 solved. @deshraj Please review the PR.


deshraj:
Please check `settings/dev.py` also.

RishabhJain2018:
@deshraj No, I don't think so that we have to change there as we are using `DummyCache Backend` for development. Please refer to this [link](https://github.com/Cloud-CV/EvalAI/blob/master/settings/dev.sample.py#L39).

Issue #528, opened 2017-01-19
Change ParticipantTeamSerializer to ParticipantTeamDetailSerializer in participant_team_list view


jayantsa:

gauthamzz:
Can i try this ?

jayantsa:
@gauthamzz  Already done that . Was an issue to solve a previous one

deshraj:
Fixed in #529

PR #529, opened 2017-01-19
changed Serializer and test for same

jayantsa:
solves #528


jayantsa:
@deshraj Please Review

jayantsa:
@deshraj Please review this is for #528

PR #549, opened 2017-01-21
Pagination for Participant team view.

aka-jain:
@spyshiv @deshraj Please review!


deshraj:
@aka-jain does this fix the selection of participant view in Challenge Page?

deshraj:
Oh never mind, Just checked on the local machine and seems to work well for the Participant Team View. Good Job. :-)

PR #553, opened 2017-01-21
Correct Path Added in Migration Scripts.

RishabhJain2018:
Added Correct Path Added in Migration Scripts.


deshraj:
Thanks for fixing this @RishabhJain2018 :-)

PR #555, opened 2017-01-22
Challenge Phase Code Name and Challenge Id made unique together

deshraj:


deshraj:
@TroJan Please review

Issue #557, opened 2017-01-22
Change placeholder when Inviting other users


deshraj:
Currently the placeholder is set to `deshraj@cloudcv.org` . We need to change it to `foobar@cloudcv.org`

Ram81:
can I do this too @deshraj with #560?

aka-jain:
Closing it. Fix with #570

PR #570, opened 2017-01-22
Fix: Change Placeholder when Inviting other Users(#557)

Ram81:
@deshraj please review!


frontend/src/js/controllers/challengeHostTeamsCtrl.js
```diff
@@ -198,7 +198,7 @@
             var confirm = $mdDialog.prompt()
                 .title('Invite others to this Team')
                 .textContent('Enter the email address of the person')
-                .placeholder('deshraj@cloudcv.org')
+                .placeholder('foobar@cloudcv.org')
```
aka-jain: @Ram81 : please replace this text to "email" only. There is no need to add dummy email here.

Ram81:
@aka-jain  please review

PR #592, opened 2017-01-24
Fixed Error message in Sign up view

anweshknayak:
@aka-jain @spyshiv this should work ? I cannot test it since gulp task is breaking

### DIFF
--- a/frontend/src/js/controllers/authCtrl.js
+++ b/frontend/src/js/controllers/authCtrl.js
@@ -56,6 +56,8 @@
 
             //reset error msg
             vm.wrnMsg = {};
+            vm.isFormError = false;
+            vm.FormError = {};
         }
 
         // Function to signup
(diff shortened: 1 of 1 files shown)


aka-jain:
@anweshknayak Please make your code runnable, so that it will be easy for you and us to understand. 
@spyshiv I think reset function will not work with form errors., wil you please check it once.

spyshiv:
@aka-jain: We have to reset variables on route change that will be permanent solution. The approach of calling reset function on click will conflict Since we can't initiate `ng-click` and `ui-sref` at a time.

aka-jain:
Agree @spyshiv , we have to undo these changes.

anweshknayak:
@aka-jain @spyshiv what is to be done about this ?

aka-jain:
@anweshknayak Need to close this PR. We have to do some changes regarding this.

PR #597, opened 2017-01-24
Fix: Route and View created for Terms and Conditions page (#566)

Ram81:
@deshraj  please review!

PR #598, opened 2017-01-24
About Us: About us Page created.

RishabhJain2018:
fixes. #598  @aka-jain @spyshiv @deshraj  Please review the PR.

Moreover, @aka-jain Do I also need to create a separate `about-us` page for the footer `About-Us` link in the `landing page`?

### DIFF
--- a/frontend/src/views/web/dashboard-about-us.html
+++ b/frontend/src/views/web/dashboard-about-us.html
@@ -1,0 +1,7 @@
+<div class="ev-sm-container en-view">
+    <div class="row">
+        <div class="col s12">
+            Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. It has survived not only five centuries, but also the leap into electronic typesetting, remaining essentially unchanged. It was popularised in the 1960s with the release of Letraset sheets containing Lorem Ipsum passages, and more recently with desktop publishing software like Aldus PageMaker including versions of Lorem Ipsum.
+        </div>
+    </div>
+</div>
--- a/frontend/src/views/web/dashboard-about-us.html
+++ b/frontend/src/views/web/dashboard-about-us.html
@@ -1,7 +1,0 @@
-<div class="ev-sm-container en-view">
-    <div class="row">
-        <div class="col s12">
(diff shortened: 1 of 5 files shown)


aka-jain:
@RishabhJain2018 try to make a seperate route for About u page. Because it is not based on Authentication so just make a url as "localhost:8888/about-us" and make a new view for it. Getting?

RishabhJain2018:
@aka-jain Yeah, earlier I thought so, but with that we need to include all the `js ,css etc.`  files again & complete page will be reloaded when the user will click `About Us`. but in the above case as I have done we only need to make a `about us` html file for `landing page` whose parent will be `auth` for dashboard `About us` link the parent will be `web`. What are your views about this?

aka-jain:
No @RishabhJain2018 , index file is the Grandparent of all routes, so it will work fine. BAsically localhost:8888 is itself an index file, which placed at topmost level. 
You can check Ram's PR at https://github.com/Cloud-CV/EvalAI/pull/597
I hope it maybe helpful for you :)

RishabhJain2018:
Ohh..okk @aka-jain sure.

aka-jain:
@RishabhJain2018 There will be only one file for about us page. You are using 2 same files for dashboard and landing. Please make it one because it isn't dependent on auth or without login, user can also  access it. So make it one and redirect to that html from dashboard as well as landing page footer.

frontend/src/views/web/dashboard-about-us.html
```diff
@@ -0,0 +1,7 @@
+<div class="ev-sm-container en-view">
+    <div class="row">
+        <div class="col s12">
+            Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. It has survived not only five centuries, but also the leap into electronic typesetting, remaining essentially unchanged. It was popularised in the 1960s with the release of Letraset sheets containing Lorem Ipsum passages, and more recently with desktop publishing software like Aldus PageMaker including versions of Lorem Ipsum.
+        </div>
+    </div>
+</div>
```
aka-jain: @RishabhJain2018 I think there is no need to use this file. Please remove it.

RishabhJain2018: yeah sure.

RishabhJain2018:
@aka-jain Please review the PR.

Issue #559, opened 2017-01-22
Show a default logo when there is no logo for a challenge


deshraj:
When there is no logo for a particular challenge, then we need to show a default logo of EvalAI there. 
@aka-jain can you give a sample logo that needs to be shown if the original logo is not ready yet?

Ram81:
@deshraj as I am done with my previous work can I do this too ?

gauthamzz:
I would like to try this

Ram81:
@deshraj anyone doing this issue?

deshraj:
Sure, go ahead. You can use the CloudCV logo to display when there in no logo for a particular challenge.

Ram81:
@deshraj  you can close this one fixed in #611

aka-jain:
Fixed with #611

PR #611, opened 2017-01-29
Fix: Default Logo shown when there is no logo for challenge (#559)

Ram81:
@deshraj  @aka-jain  please review!


frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -38,6 +38,11 @@
                 var response = response.data;
                 vm.page = response;
                 vm.isActive = response.is_active;
+                
+                if(vm.page.image == null){
+                	vm.page.image = "dist/images/logo.png";
```
aka-jain: @Ram81 Do we have an image named as "logo.png"? I don't think we have it.

Ram81: @aka-jain  yes there is it is the cloudcv logo image!

PR #633, opened 2017-02-05
Solves Login-Signup Toggle and Host Team Delete API Endpoint problem

yadavankit:
@aka-jain This is working good for me. Have a look :)

### DIFF
--- a/apps/hosts/views.py
+++ b/apps/hosts/views.py
@@ -199,7 +199,7 @@ def remove_self_from_challenge_host_team(request, challenge_host_team_pk):
         response_data = {'error': 'ChallengeHostTeam does not exist'}
         return Response(response_data, status=status.HTTP_406_NOT_ACCEPTABLE)
     try:
-        challenge_host = ChallengeHost.objects.filter(user=request.user.id, team_name__pk=challenge_host_team_pk)
+        challenge_host = ChallengeHostTeam.objects.filter(created_by=request.user.id, pk=challenge_host_team_pk)
         challenge_host.delete()
         return Response(status=status.HTTP_204_NO_CONTENT)
     except:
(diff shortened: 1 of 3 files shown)


apps/hosts/views.py
```diff
@@ -199,7 +199,7 @@ def remove_self_from_challenge_host_team(request, challenge_host_team_pk):
         response_data = {'error': 'ChallengeHostTeam does not exist'}
         return Response(response_data, status=status.HTTP_406_NOT_ACCEPTABLE)
     try:
-        challenge_host = ChallengeHost.objects.filter(user=request.user.id, team_name__pk=challenge_host_team_pk)
+        challenge_host = ChallengeHostTeam.objects.filter(created_by=request.user.id, pk=challenge_host_team_pk)
```
deshraj: @yadavankit since you are filtering the `ChallengeHostTeam`, so it will delete the `ChallengeHostTeam` object instead of `ChallengeHost`.

frontend/src/js/controllers/authCtrl.js
```diff
@@ -56,6 +56,9 @@
 
             //reset error msg
             vm.wrnMsg = {};
+
+            //switch off form errors
+            vm.isFormError = false;
```
spyshiv: LGTM :+1:

apps/hosts/views.py
yadavankit: @deshraj this was some real desperate work from my side 😂  Coming to the problem, as of now, we show the teams `created_by` user in challenge-host-teams page. Since, the user still (after his removal from `ChallengeHost`) remains the `created_by` of `ChallengeHostTeam` this is happening. Well, will be reverting back this commit :)

deshraj: I just commented in another issue thread. Please look at that thread.

PR #634, opened 2017-02-05
fix api end point remove host team

gauthamzz:


deshraj:
@gauthamzz as discussed with you on DM, I am closing this.

PR #648, opened 2017-02-08
Frontend (Fix): Remove unnecessary gitignore inside frontend directory

spyshiv:
* Remove unnecessary `.gitignore` (inside `frontend/`) 
* Remove `frontend/index.html` since `base.html` will build it dynamically
* Add `frontend/index.html` in `.gitignore` (of root directory)

@aka-jain / @deshraj Please Merge this.


frontend/.gitignore
```diff
@@ -1,8 +0,0 @@
-# files/directories to be ignored
-
-.sass-cache/
-
-node_modules/
-bower_components/
-
-dist/
```
aka-jain: @spyshiv We need dist/ I think?

spyshiv: @aka-jain : See the .gitignore of root directory. 
there is `dist/` (line 14) that's working for the `frontend/dist/` .
However run project in locally and see this if everything is fine.

.gitignore
```diff
@@ -79,6 +79,7 @@ fabfile/
 .sass-cache/
 node_modules/
 bower_components/
+frontend/index.html
```
spyshiv: @aka-jain : Should i add `frontend/dist/` here to make difference between frontend and backend??

aka-jain: Yep, do this please! It will be helpful and more meaningful for other contributors.

spyshiv:
@aka-jain : Done

Issue #352, opened 2017-01-07
Integrate Eslint with the frontend code


deshraj:

deshraj:
For reference, see this http://eslint.org/docs/user-guide/configuring

RishabhJain2018:
@deshraj Can I work on this??

RishabhJain2018:
@deshraj should I create the configuration file at `EvalAI/.eslintrc.json` this location?

aka-jain:
@RishabhJain2018 : Do some research and find the best way to integrate Eslint by your own.

RishabhJain2018:
@aka-jain okk sure.

deshraj:
Assignin to @aka-jain now.

dexter1691:
@RishabhJain2018: This resource might be very useful: https://github.com/airbnb/javascript
It's the gold standard guidelines used by a lot of javascript repositories.

aka-jain:
@dexter1691 : Airbnb guidelines are really awesome but we are already using John Papa Guidelines for our Javascript (Angular1), also currently we are not working on ES6 version which is highly recommendable for production now a days so I think at this point this repo may not be useful for us as many guidlines are related to ES6. But to make code quality good, John Papa guidelines are also fair enough for JS and Angular. https://github.com/johnpapa/angular-styleguide/blob/master/a1/README.md

@RishabhJain2018 : You can go thorugh the Airbnb repo, it will be very helpful to understand the structure of Javascript :) And for our project, try to go through ES5 section.

dexter1691:
@aka-jain : This is done? Should we close?

aka-jain:
@dexter1691 : Need to do some improvements on this. I'm working on it.

spyshiv:
@aka-jain : Assigning this to me since you have other tasks to do.

aka-jain:
@spyshiv Sure.

aka-jain:
@spyshiv Is it done with Gulp improvement?

spyshiv:
@aka-jain : Almost done. There is seperate branch `js_linting` for this purpose. Need minor improvements in it.
However checkout to that branch and run `gulp lint` to see the errors and warnings.

aka-jain:
Fixed with #650

PR #650, opened 2017-02-08
Frontend(js linting): Add js linting packages

spyshiv:
Solution for #352 
* Integrated js linting task with gulp

## How to run
* `gulp dev:runserver`
or 
* `gulp lint`

It is integrated with gulp so on development when `gulp dev:runserver` runs then lint task will run and errors and warnings can be seen on the terminal.

Since there are too many minor errors so for now everything will work as it is even after showing errors and warnings but after fixing these errors i will rewrite runserver code so that server will not run until the linting errors remain.

@aka-jain @deshraj : Please review.


aka-jain:
Agree @spyshiv , great work. Will create an issue for this. Do we need to add it in readme?

PR #657, opened 2017-02-09
Fix(gulp inject): Add rootslash in static path injection

spyshiv:
@deshraj @aka-jain Please merge this


aka-jain:
Nice catch @spyshiv. Merging it.

PR #660, opened 2017-02-09
Worker: Support for dataset_splits added in worker

deshraj:


deshraj:
@dexter1691 @TroJan Please review the PR.

scripts/workers/submission_worker.py
```diff
@@ -41,7 +41,7 @@
 os.environ.setdefault('DJANGO_SETTINGS_MODULE', DJANGO_SETTINGS_MODULE)
 django.setup()
 
-from challenges.models import Challenge, ChallengePhase     # noqa
+from challenges.models import Challenge, ChallengePhase, ChallengePhaseSplit, DatasetSplit, LeaderboardData  # noqa
```
TroJan: can you please span this import to multiple lines ?

scripts/workers/submission_worker.py
```diff
@@ -277,12 +283,61 @@ def run_submission(challenge_id, challenge_phase, submission_id, submission, use
...
-        submission_output = EVALUATION_SCRIPTS[challenge_id].evaluate(annotation_file_path,
-                                                                      user_annotation_file_path,
-                                                                      challenge_phase.codename,)
+    try:
+        successful_submission_flag = True
+        with stdout_redirect(stdout) as new_stdout, stderr_redirect(stderr) as new_stderr:      # noqa
+            submission_output = EVALUATION_SCRIPTS[challenge_id].evaluate(annotation_file_path,
+                                                                          user_annotation_file_path,
+                                                                          challenge_phase.codename,)
+
```
TroJan: write a proper comment explaining the criteria of marking a submission as successful. This will help any one digging or updating code in future.

scripts/workers/submission_worker.py
```diff
@@ -277,12 +283,61 @@ def run_submission(challenge_id, challenge_phase, submission_id, submission, use
...
+                leaderboard_data = LeaderboardData()
+                leaderboard_data.challenge_phase_split = challenge_phase_split
+                leaderboard_data.submission = submission
+                leaderboard_data.leaderboard = challenge_phase_split.leaderboard
+                leaderboard_data.result = split_result[dataset_split.codename]
+
+                leaderboard_data_list.append(leaderboard_data)
+
+            # Check to see if length of leaderboard_data_list matches with length of submission_output['result']
```
TroJan: writing comments is actually good, but I will prefer that you write the reason for the check instead of explaining what the code is doing. Explaining a piece of code by writing a comment states that the code is not readable. Here in your case the code is good enough to read, so it will be great if you can write the reason explaining "WHY".

scripts/workers/submission_worker.py
```diff
@@ -277,12 +283,61 @@ def run_submission(challenge_id, challenge_phase, submission_id, submission, use
...
+
+        # Once the submission_output is processed, then save the submission object with appropriate status
+        else:
+            successful_submission_flag = False
+
+    except:
+        stderr.write(traceback.format_exc())
+        successful_submission_flag = False
+
+    set_submission_status(submission, successful_submission_flag)
```
TroJan: Since this is one time activity, do we need a function for this?

scripts/workers/submission_worker.py
```diff
@@ -277,12 +283,61 @@ def run_submission(challenge_id, challenge_phase, submission_id, submission, use
...
+
+            leaderboard_data_list = []
+            for split_result in submission_output['result']:
+
+                # Check if the dataset_split exists for the codename in the result
+                try:
+                    split_code_name = split_result.items()[0][0]  # get split_code_name that is the key of the result
+                    dataset_split = DatasetSplit.objects.get(codename=split_code_name)
+                except:
+                    stderr.write(traceback.format_exc())
```
TroJan: what if apart from the traceback, we can also write an informatory message on stderr, so that it becomes more clear on seeing the error file.

scripts/workers/submission_worker.py
```diff
@@ -277,12 +283,61 @@ def run_submission(challenge_id, challenge_phase, submission_id, submission, use
...
+                    successful_submission_flag = False
+                    break
+
+                leaderboard_data = LeaderboardData()
+                leaderboard_data.challenge_phase_split = challenge_phase_split
+                leaderboard_data.submission = submission
+                leaderboard_data.leaderboard = challenge_phase_split.leaderboard
+                leaderboard_data.result = split_result[dataset_split.codename]
```
TroJan: Can you please use `.get` here?

PR #671, opened 2017-02-13
Frontend(urls routing): Prevent 404 error on trailing slash at the end of urls

spyshiv:
Fix for getting 404 error page on adding trailing slash in urls

@deshraj Please review and test


deshraj:
Tested on my local machine and working perfectly. Good job ;)

Issue #428, opened 2017-01-13
Instance is None while saving the file in Submission Model


deshraj:
In Submission model, when the model is saved then the path where the `input_file` is saved is `submission_files/submission_None/` but ideally the `None` should be replaced with the id of the submission.

deshraj:
@TroJan can you please prioritize this over all other issues assigned to you? It seems like this the super critical thing for us for now.

TroJan:
Fixed via #681

PR #681, opened 2017-02-16
Jobs: Handle case when instance is None for input_file.

TroJan:
When submission was saved, input_file was saved in a folder named
submission_None. This handles the case for it by using signals.


TroJan:
Issue #428

PR #682, opened 2017-02-16
API endpoint for team member details

yadavankit:
Adds functionality for our-team page. API is available at `/api/web/team/` `GET` request.


deshraj:
@yadavankit, Also please write unit tests for the view that you have added.

yadavankit:
@deshraj will be working on unit test now :)

apps/web/models.py
```diff
@@ -33,11 +35,13 @@ class Team(models.Model):
 
     name = models.CharField(max_length=100)
     email = models.EmailField(max_length=70)
+    description = RichTextField(null=True)
```
deshraj: You can simply keep it text field instead of using RichtextField. We don't want to use ckeditor just for the purpose of showing Team details.

apps/web/models.py
```diff
@@ -33,11 +35,13 @@ class Team(models.Model):
...
     headshot = models.ImageField(upload_to="headshots", null=True, blank=True)
     visible = models.BooleanField(default=True)
     github_url = models.URLField(max_length=200)
     linkedin_url = models.URLField(max_length=200)
     personal_website = models.URLField(max_length=200)
+    bg_image = models.ImageField(upload_to="bg-images", null=True, blank=True)
```
deshraj: Please rename this to `background_Image`

apps/web/models.py
yadavankit: I guess @aka-jain mentioned explicitly for a text editor in admin, so, I used ckeditor. Well, will be reverting if not needed :)

apps/web/models.py
yadavankit: Sure 👍

deshraj: Edit: field name should be `background_image`

apps/web/models.py
deshraj: Angular can parse the `<i>` tags. So, there is no point in keeping this. Moreover, this is not a client side facing feature and hence we should not add extra unnecessary dependency.

yadavankit:
@deshraj please review :)

apps/web/views.py
```diff
@@ -55,3 +57,16 @@ def contact_us(request):
...
             return Response(response_data, status=status.HTTP_201_CREATED)
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+@throttle_classes([AnonRateThrottle])
+@api_view(['GET'])
+@permission_classes((permissions.AllowAny,))
+def our_team(request):
+    team = Team.objects.all()
+    if not team:
```
deshraj: @yadavankit I don't think we need this check here. If there will be no team then an empty queryset will be returned and that is totally fine.

apps/web/views.py
```diff
@@ -55,3 +57,16 @@ def contact_us(request):
             response_data = {'message': 'Your message has been successfully recorded. We will contact you shortly.'}
             return Response(response_data, status=status.HTTP_201_CREATED)
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+@throttle_classes([AnonRateThrottle])
+@api_view(['GET'])
+@permission_classes((permissions.AllowAny,))
+def our_team(request):
+    team = Team.objects.all()
```
deshraj: Please make it `teams`

tests/unit/web/test_views.py
```diff
@@ -32,7 +33,7 @@ def setUp(self):
         self.data = {
             'name': 'New Message',
             'email': 'newuser@test.com',
-            'message': 'New Message for Conatct'
+            'message': 'New Message for Contact'
```
deshraj: You can open a new PR for this typo fix. Please remove this from this PR since its not a part of this.

yadavankit: will do it the first thing in morning :)

yadavankit: @deshraj review please :)

yadavankit:
@deshraj changes done

Issue #714, opened 2017-02-22
Footer in About Us page is not aligned 


Ram81:
![footer](https://cloud.githubusercontent.com/assets/16323427/23216486/214f297a-f93c-11e6-893f-4d5f5fd6ab03.png)

Ram81:
@aka-jain I must be fixed at bottom!

deshraj:
Thanks for opening this @Ram81 :) Feel free to fix this.

deshraj:
Fixed in #715

PR #715, opened 2017-02-22
Fix: Footer Aligned for About Us Page (#714)

Ram81:
@deshraj please review


deshraj:
Very Quick 😄 

Fixes #714

PR #721, opened 2017-02-22
Jobs: Path fixture for the submission related files.

### LINKED ISSUE #652: Enforce unique random file name for files being uploaded
Currently, the files being uploaded are having simple name and we don't want them to serve to anyone who knows the pattern. Hence, we should store the files with name as uuid that will to secure it to some extent. 

For reference, see this http://stackoverflow.com/questions/2673647/enforce-unique-upload-file-names-using-django?answertab=active#tab-top

RishabhJain2018:
Fixes #652 . @deshraj @TroJan Please review the PR.


apps/jobs/models.py
```diff
@@ -70,9 +76,9 @@ class Submission(TimeStampedModel):
     started_at = models.DateTimeField(null=True, blank=True)
     completed_at = models.DateTimeField(null=True, blank=True)
     when_made_public = models.DateTimeField(null=True, blank=True)
-    input_file = models.FileField(upload_to=RandomFileName("submission_files/submission"))
-    stdout_file = models.FileField(upload_to=RandomFileName("submission_files/submission"), null=True, blank=True)
-    stderr_file = models.FileField(upload_to=RandomFileName("submission_files/submission"), null=True, blank=True)
+    input_file = models.FileField(upload_to=get_upload_path)
```
TroJan: Can we please remove `get_upload_path` and add some hook in `RandomFileName` class.

RishabhJain2018: @TroJan I think if we try to add a hook in `RandomFileName` then we may not be able to use it to other places, like we have used it[ here](https://github.com/Cloud-CV/EvalAI/blob/master/apps/challenges/models.py#L35). What are your views about it?

TroJan: @RishabhJain2018 : will update your shortly over this

RishabhJain2018: @TroJan Yeah Sure.

TroJan:
@RishabhJain2018:

what if we update the `RandomFileName.__call__` to something like this

```
extension = os.path.splitext(filename)[1]
if 'id' in self.path and instance.pk:
            self.path = self.path.format(id=instance.pk)
filename = '{}{}'.format(uuid.uuid4(), extension)
filename = os.path.join(self.path, filename)
return filename
```


and then in the model field, call it by

```
input_file = models.FileField(upload_to=RandomFileName("submission_files/submission_{id}"))
```

Can you check for all the case if it is working fine?  Let me know if this does not works

RishabhJain2018:
@TroJan Yeah, it is working fine for all the cases. 
I have updated the PR according to this. Thanks for guiding. :-)

deshraj:
@TroJan can you please try this on your local machine and confirm once?

TroJan:
Looks ok to me, @deshraj : your input ?

deshraj:
Looks good to me.

Issue #733, opened 2017-02-24
Reset password page doesn’t reset when moved from one view to another


deshraj:
## How to replicate this issue?

1. Click on Forgot password
2. Submit email
3. Click `Back to Log In`
4. Click again on `Forgot Password`

Ram81:
@deshraj  can I take this up?

deshraj:
Yep, sure.

Ram81:
@deshraj fixed in #735 you can close this

spyshiv:
Closing this.

PR #735, opened 2017-02-24
Fix: Reset Password does not work when switching (#733)

Ram81:
@deshraj please review


frontend/src/js/controllers/authCtrl.js
```diff
@@ -59,6 +59,9 @@
 
             //switch off form errors
             vm.isFormError = false;
+            
+            //reset form when link sent for reset password
+            vm.isMail=true;
```
spyshiv: @Ram81: Nice catch. 
Put one space before and after `=` .

Ram81:
@spyshiv  Done

Ram81:
@spyshiv Done

PR #748, opened 2017-03-02
Docs: Add verbose setup docs.

TroJan:
To update user and password in sample settings `postgres` and `postgres`


docs/setup.md
```diff
@@ -0,0 +1,120 @@
+## How to setup
+
+This section guides first-time contributor through installing the EvalAI development environment on Ubuntu
```
deshraj: It would be great if we can mention the 'recommended' version of Ubuntu for the contributors.

Issue #744, opened 2017-03-01
Frontend: Changing route doesn't scroll to top in the new page


spyshiv:
Example:
1. Open home page and go to footer
2. In footer click on the `about us` link

You can see the page doesn't scroll to top. See the screenshot
![screenshot from 2017-03-02 02 04 20](https://cloud.githubusercontent.com/assets/7015220/23480065/94a3367c-feec-11e6-8a55-7129a7c69e3d.png)

aka-jain:
@spyshiv I'm on it.

spyshiv:
cool :+1:

aka-jain:
Task is open, I'm working on some other issue.

spyshiv:
@aka-jain : Cool :+1:

gauthamzz:
I would like to try this

spyshiv:
@gauthamzz Sure assigning to you

aka-jain:
@gauthamzz are you still on it?

gauthamzz:
@aka-jain  i am i need one more day , i have an exam tmmrw.then im free .Is that fine ?

aka-jain:
Let's free this issue till then. Let me know whenever you get free. Will assign it to you again if no one shows interest in it.

spyshiv:
@anweshknayak : can you fix this ?? Let me know if you are interested.

anweshknayak:
@spyshiv Sure. I am interested.

spyshiv:
@anweshknayak: Let see how quick you are.... Assigning this to you.

PR #776, opened 2017-03-03
Frontend: Changing route doesn't scroll to top in the new page

gauthamzz:
#744 is solved @spyshiv please review

### DIFF
--- a/frontend/src/views/web/about-us.html
+++ b/frontend/src/views/web/about-us.html
@@ -1,7 +1,7 @@
 <div class="white-bg view-container">
     <main-header></main-header>
     <!-- EvalAI into -->
-    <div class="grad-container">
+    <div class="grad-container" smooth-scroll>
         <div class="row">
             <div class="col s12 m8 l8">
                     <br><br>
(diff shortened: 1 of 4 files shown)


aka-jain:
@gauthamzz I think you took this issue in a wrong way. @spyshiv need your thoughts here.

gauthamzz:
@aka-jain idk im confused , you just need to scroll back to top automatically when that page is loaded right ?

aka-jain:
So do we have this directive already installed "smooth-scroll"?

gauthamzz:
i saw that https://github.com/d-oliveros/ngSmoothScroll was already installed , so i used its functionality. I thought it might be easier. Should i change it ?

anweshknayak:
Yeah in https://github.com/Cloud-CV/EvalAI/blob/master/bower.json#L26

aka-jain:
@gauthamzz that's not a correct way to solve this. You have to make a function in run block of Angular so that every time the route changes the page scroll to it's topmost position. Smooth scroll is just a hack and if we go with it then we have to take care to add it in each views which might be a problem.

aka-jain:
And that smooth scroll we used to scroll to a particular part of a page.

aka-jain:
Fixed with #813

PR #790, opened 2017-03-04
Moves hardcoded connection params from worker to settings

### LINKED ISSUE #783: Move connection params out of worker to settings
We are using RabbitMQ to publish messages for submission and execute them. RabbitMQ needs some connection parameters like host name, exchange name, queue name, etc.

To consume message from the queue, a worker is written. This workers lives at `scripts/workers/submission_worker.py`.  

Presently the connection as well as exchange parameters are hardcoded in that file. Ideally they should be present in environment variables. But as of now we can avoid keeping them in environment variables and instead leverage Django settings.

So this task involves moving the connection parameters to settings. Concerned parameters are 
* host
* exchange name
* exchange type
* submission queue name
* submission queue parameters

As of now all the parameters can live in base settings. Also please make sure that you define a dict something like `RABBITMQ_PARAMETERS` and within it multiple keys for host, exchange and all.
So final value of variable should look like

```
RABBITMQ_PARAMETERS = {
    'HOST': 'value',
    'EXCHANGE_NAME': 'value'
    'SUBMISSION_QUEUE': {
        'NAME': 'value',
        ....
    }
}
```

mayank-agarwal-96:
Fixes #783 
The connection params are now moved to settings/common.py .

### DIFF
--- a/scripts/workers/submission_worker.py
+++ b/scripts/workers/submission_worker.py
@@ -58,7 +58,6 @@ PHASE_ANNOTATION_FILE_PATH = join(PHASE_DATA_DIR, '{annotation_file}')
 SUBMISSION_DATA_DIR = join(SUBMISSION_DATA_BASE_DIR, 'submission_{submission_id}')
 SUBMISSION_INPUT_FILE_PATH = join(SUBMISSION_DATA_DIR, '{input_file}')
 CHALLENGE_IMPORT_STRING = 'challenge_data.challenge_{challenge_id}'
-RABBITMQ_PARAMETERS = settings.RABBITMQ_PARAMETERS
 EVALUATION_SCRIPTS = {}
 
 # map of challenge id : phase id : phase annotation file name
@@ -446,12 +445,12 @@ def main():
 
     load_active_challenges()
     connection = pika.BlockingConnection(pika.ConnectionParameters(
-        host=RABBITMQ_PARAMETERS['HOST']))
+        host=settings.RABBITMQ_PARAMETERS['HOST']))
 
     channel = connection.channel()
     channel.exchange_declare(
-        exchange=RABBITMQ_PARAMETERS['EVALAI_EXCHANGE']['NAME'],
-        type=RABBITMQ_PARAMETERS['EVALAI_EXCHANGE']['TYPE'])
+        exchange=settings.RABBITMQ_PARAMETERS['EVALAI_EXCHANGE']['NAME'],
+        type=settings.RABBITMQ_PARAMETERS['EVALAI_EXCHANGE']['TYPE'])
 
     # name can be a combination of hostname + process id
     # host name : to easily identify that the worker is running on which instance
@@ -460,7 +459,7 @@ def main():
                                                                 process_id=str(os.getpid()))
 
     channel.queue_declare(
-        queue=RABBITMQ_PARAMETERS['SUBMISSION_QUEUE'],
+        queue=settings.RABBITMQ_PARAMETERS['SUBMISSION_QUEUE'],
         durable=True)
 
     # reason for using `exclusive` instead of `autodelete` is that
@@ -473,15 +472,15 @@ def main():
     create_dir_as_python_package(SUBMISSION_DATA_BASE_DIR)
 
     channel.queue_bind(
-        exchange=RABBITMQ_PARAMETERS['EVALAI_EXCHANGE']['NAME'],
-        queue=RABBITMQ_PARAMETERS['SUBMISSION_QUEUE'],
+        exchange=settings.RABBITMQ_PARAMETERS['EVALAI_EXCHANGE']['NAME'],
+        queue=settings.RABBITMQ_PARAMETERS['SUBMISSION_QUEUE'],
         routing_key='submission.*.*')
     channel.basic_consume(
         process_submission_callback,
-        queue=RABBITMQ_PARAMETERS['SUBMISSION_QUEUE'])
+        queue=settings.RABBITMQ_PARAMETERS['SUBMISSION_QUEUE'])
 
     channel.queue_bind(
-        exchange=RABBITMQ_PARAMETERS['EVALAI_EXCHANGE']['NAME'],
+        exchange=settings.RABBITMQ_PARAMETERS['EVALAI_EXCHANGE']['NAME'],
         queue=add_challenge_queue_name, routing_key='challenge.add.*')
     channel.basic_consume(add_challenge_callback, queue=add_challenge_queue_name)
 
--- a/scripts/workers/submission_worker.py
+++ b/scripts/workers/submission_worker.py
@@ -477,10 +477,12 @@ def main():
         exchange=RABBITMQ_PARAMETERS.get('EXCHANGE_NAME', None),
         queue=RABBITMQ_PARAMETERS.get('SUBMISSION_QUEUE', None).get('NAME', None),
         routing_key='submission.*.*')
-    channel.basic_consume(process_submission_callback,
+    channel.basic_consume(
+        process_submission_callback,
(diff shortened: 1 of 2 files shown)


mayank-agarwal-96:
@TroJan @deshraj  flake8 is giving out 

> E402 module level import not at top of file for the statement 

`from settings.common import RABBITMQ_PARAMETERS
`
But this should be imported after django.setup() has been called. What can I do about it?

mayank-agarwal-96:
@TroJan @aka-jain @deshraj Please review.

scripts/workers/submission_worker.py
```diff
@@ -446,20 +446,23 @@ def main():
     sys.path.append(COMPUTE_DIRECTORY_PATH)
 
     load_active_challenges()
-
     connection = pika.BlockingConnection(pika.ConnectionParameters(
-        host='localhost'))
+        host=RABBITMQ_PARAMETERS.get('HOST', None)))
```
TroJan: We cannot use None as host, So `HOST` key must exist in `RABBITMQ_PARAMETERS`.

Also you can safely remove `None` everywhere.

scripts/workers/submission_worker.py
```diff
@@ -446,20 +446,23 @@ def main():
...
 
     channel = connection.channel()
-    channel.exchange_declare(exchange='evalai_submissions', type='topic')
+    channel.exchange_declare(
+        exchange=RABBITMQ_PARAMETERS.get('EXCHANGE_NAME', None),
```
TroJan: I will prefer that you make key as `EVALAI_EXCHANGE` and then `NAME` and `TYPE` further to nest that dict.

scripts/workers/submission_worker.py
```diff
@@ -446,20 +446,23 @@ def main():
...
     # name can be a combination of hostname + process id
     # host name : to easily identify that the worker is running on which instance
     # process id : to add uniqueness in case more than one worker is running on the same instance
     add_challenge_queue_name = '{hostname}_{process_id}'.format(hostname=socket.gethostname(),
                                                                 process_id=str(os.getpid()))
 
-    channel.queue_declare(queue='submission_task_queue', durable=True)
+    channel.queue_declare(
+        queue=RABBITMQ_PARAMETERS.get('SUBMISSION_QUEUE', None).get('NAME', None),
+        durable=RABBITMQ_PARAMETERS.get('SUBMISSION_QUEUE', None).get('DURABLE', None))
```
TroJan: No need of taking durable key from settings, this is something which will vary from queue to queue, so this much of arguments can be kept in worker only.

scripts/workers/submission_worker.py
```diff
@@ -470,10 +473,17 @@ def main():
...
+        exchange=RABBITMQ_PARAMETERS.get('EXCHANGE_NAME', None),
+        queue=RABBITMQ_PARAMETERS.get('SUBMISSION_QUEUE', None).get('NAME', None),
+        routing_key='submission.*.*')
+    channel.basic_consume(
+        process_submission_callback,
+        queue=RABBITMQ_PARAMETERS.get('SUBMISSION_QUEUE', None).get('NAME', None))
+
+    channel.queue_bind(
+        exchange=RABBITMQ_PARAMETERS.get('EXCHANGE_NAME', None),
+        queue=add_challenge_queue_name, routing_key='challenge.add.*')
```
TroJan: We use `.get` generally when we are not sure that the key will exist in dictionary or not. If we see the case here, this is something which is totally in our hand, so we can safely use `RABBITMQ_PARAMTERS['SUBMISSION_QUEUE']`

scripts/workers/submission_worker.py
```diff
@@ -48,7 +48,7 @@
                                LeaderboardData) # noqa
 
 from jobs.models import Submission          # noqa
-
+from settings.common import RABBITMQ_PARAMETERS        # noqa
```
TroJan: Can you try `from django.conf import settings` and see if it works?

To refer `RABBITMQ_PARAMETERS`, you should use `settings.RABBITMQ_PARAMETERS`. Let me know if it does not works.

settings/common.py
```diff
@@ -261,3 +261,13 @@
         'BACKEND': 'django.core.cache.backends.memcached.MemcachedCache',
     }
 }
+
+RABBITMQ_PARAMETERS = {
```
TroJan: The whole object will change accordingly to the above review comments.

mayank-agarwal-96: Thank you for a detailed review. I am working on changes.

scripts/workers/submission_worker.py
mayank-agarwal-96: shall I add a check for this? It will check if the host key exists or not.

settings/common.py
mayank-agarwal-96: I have made the changes.

scripts/workers/submission_worker.py
```diff
@@ -59,7 +58,7 @@
 SUBMISSION_DATA_DIR = join(SUBMISSION_DATA_BASE_DIR, 'submission_{submission_id}')
 SUBMISSION_INPUT_FILE_PATH = join(SUBMISSION_DATA_DIR, '{input_file}')
 CHALLENGE_IMPORT_STRING = 'challenge_data.challenge_{challenge_id}'
-
+RABBITMQ_PARAMETERS = settings.RABBITMQ_PARAMETERS
```
TroJan: I am actually thinking here is it required ? We can simply refer `settings.RABBITMQ_PARAMETERS`

mayank-agarwal-96: I thought it would be neat to define a variable for it. I will change it and use settings.RABBITMQ_PARAMETERS instead.

mayank-agarwal-96:
@TroJan Changed it!

TroJan:
@mayank-agarwal-96 : Just being curious here, can you tell me if you faced any difficulty in running the worker locally ? I actually need to make the docs more verbose so your input will help me a lot.

TroJan:
@mayank-agarwal-96 : thanks for the fix. I will really appreciate if you can help me with that docs part.

mayank-agarwal-96:
@TroJan Thanks for the merge. I will be glad to help you with the docs part. Please tell me what to do :smile:

TroJan:
@mayank-agarwal-96 : So any challenges that you faced while running the worker in your development environment ?

mayank-agarwal-96:
@TroJan Rabbitmq was not installed on my system and it is not mentioned in the setup guide to install rabbitmq-server. That was the only problem I faced.

TroJan:
@mayank-agarwal-96 : see if you can install rabbitmq by following this [doc](https://github.com/Cloud-CV/EvalAI/blob/master/docs/setup.md)

mayank-agarwal-96:
@TroJan I have already installed rabbitmq. I worked it out while trying to run the worker script.

TroJan:
@mayank-agarwal-96 : anything else where you got struck ?

mayank-agarwal-96:
@TroJan I was stuck on postgresql a bit but I was able to work it out because I had used it before. We can add a bit about postgresql user and password.

TroJan:
@mayank-agarwal-96 : sure thanks a lot :)

mayank-agarwal-96:
@TroJan Welcome :) Also I remember when I tried running python manage.py shell command, raven and django-storages were not installed. I think that is because we are installing requirements/dev.txt .

TroJan:
@mayank-agarwal-96 : the reason for this is by default we have set `DJANGO_SETTINGS_MODULE` environment variable to `settings.prod`. You can see reference of this variable in `manage.py` and `evalai/wsgi.py`

So to run every manage.py command in development environment, you explicitly need to override it.
Just add `--settings=settings.dev` in the command, like

```
python manage.py shell --settings=settings.dev
```

mayank-agarwal-96:
@TroJan Thanks :) Got it!

Issue #782, opened 2017-03-04
Make fetching pull request locally easy


TroJan:
This involves writing a bash script which will help reviewer to locally fetch a pull request and test it thoroughly on its local machine.

For reference you can consider github [checking out pull request](https://help.github.com/articles/checking-out-pull-requests-locally/#modifying-an-inactive-pull-request-locally) documentation

Details about the script:

**Path**: scripts/tools (create this folder if it does not exist)
Script will take 2 command line arguments 
* Id of pull request
* remote(default value is upstream)

So script will be run like

```
./scripts/tools/fetch-pull-request <pull_request_id> <remote>
```

The checkout branch for review should have name something like `review-<pull_request_id>`.

gauthamzz:
can i try this ?

TroJan:
@gauthamzz : sure assigning this to you :)

TroJan:
@gauthamzz : thanks for the fix. :)

PR #803, opened 2017-03-06
Make fetching pull request locally easy 

gauthamzz:
#782 is solved.
@TroJan please review


scripts/tools/fetch-pull-request
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
```
TroJan: You can use bash [set builtin](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html#The-Set-Builtin). I will prefer `-x` and `-e`. `-x` prints a trace of command so there wont be any need to write print statements.

scripts/tools/fetch-pull-request
```diff
@@ -0,0 +1,19 @@
...
+if [ $# -eq 0 ]
+    then
+        echo "Id of pull request not supplied"
+        exit 0
+fi
+
+remote=${2-upstream}
+echo "Setting remote value as $remote"
```
TroJan: When you use `set -x`, then there is no need to write any echo statements.

scripts/tools/fetch-pull-request
```diff
@@ -0,0 +1,19 @@
...
+
+#fetching pull request
+fetch="git fetch $remote pull/$1/head:review-$1"
+echo "Exectuting $fetch"
+eval $fetch
```
TroJan: Also you directly commands instead of taking them in variable and then using eval. For eg, 
```
git fetch $remote pull/$1/head:review-$1
```

scripts/tools/fetch-pull-request
```diff
@@ -0,0 +1,19 @@
...
+
+#switiching to branch review-<pull_request_id>
+checkout="git checkout review-$1"
+echo "Exectuting $checkout"
+eval $checkout
```
TroJan: Same goes here also, Directly execute the commands instead of using eval.

scripts/tools/fetch-pull-request
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
+if [ $# -eq 0 ]
+    then
```
TroJan: can we write `then` on the same line as `if`. Also `;` after if condition ends.

scripts/tools/fetch-pull-request
```diff
@@ -0,0 +1,19 @@
...
+        echo "Id of pull request not supplied"
```
TroJan: Can you please make the message as `Pull request id not given.`. Also add a help message like to execute this script give 2 arguments.

scripts/tools/fetch-pull-request
```diff
@@ -0,0 +1,19 @@
...
+        exit 0
```
TroJan: We should exit with a non zero status here, so that error can be signified.

scripts/tools/fetch-pull-request
```diff
@@ -0,0 +1,19 @@
...
+fi
+
+remote=${2-upstream}
```
TroJan: I am not fully sure, but shouldn't it be `${2:-upstream}` ?

gauthamzz: Both will work i think

scripts/tools/fetch-pull-request
gauthamzz: should i keep the echo if we are keeping the set -x then we will have messages like 

`+ '[' 0 -eq 0 ']'
+ echo 'Pull request id not given'
Pull request id not given`

scripts/tools/fetch-pull-request
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+if [ "$1" == "--help" ]; then
+    printf "Usage: fetch-pull-request [--help] pull_request_id [remote=upstream] \n\n Locally fetch a pull request and test it. \n Remote is set to upstream by default.\n"
+    exit 1
+elif [ $# -eq 0 ]; then
+    echo "Pull request id not given."
+    exit 1
+else
+    set -e -x
+    remote=${2:-upstream}
```
gauthamzz: @TroJan should i add a check if upstream exists also in this ?

scripts/tools/fetch-pull-request
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+if [ "$1" == "--help" ]; then
+    printf "Usage: fetch-pull-request [--help] pull_request_id [remote=upstream] \n\n Locally fetch a pull request and test it. \n Remote is set to upstream by default.\n"
+    exit 1
+elif [ $# -eq 0 ]; then
+    echo "Pull request id not given."
+    exit 1
+else
+    set -e -x
```
gauthamzz: i thought set -x -e would be better if we add it only after the main part of code is being implemented or it shows if statements,echo and exit also ,should i change it ?

deshraj:
Also, we would prefer using `_` instead of `-` for the file name. Please change accordingly.

TroJan:
@gauthamzz : LGTM. Can you please update the file name as @deshraj has suggested ?(use underscore instead of hyphen)

gauthamzz:
@TroJan requested changes done :)

PR #808, opened 2017-03-06
Login page redesigned with rocket container

chan24:
Redesigned login page with _rocket container_. Retaining the same bg but reused the _rocketcontainer_

 module into the page.


chan24:
@aka-jain Review my submission.

aka-jain:
@chan24 thanks for the efforts but the result is not looking as per expectation.
Dude you have to make the whole page like about us page. 
Form is not left aligned, and container should be of fixed height. 
Let me close this PR, I think I should make some new issue before that. I'm closing this PR for now.

Issue #815, opened 2017-03-07
Redesign auth view like we have other pages.


aka-jain:
Redesign auth views, namely:
Reset Password
Email Confirmation

@deshraj add other auth views too if I left any.

Ram81:
@aka-jain can I take this one?

aka-jain:
Sure @Ram81.

deshraj:
@aka-jain is it different from #996?

aka-jain:
@deshraj description updated!

deshraj:
Fixed in #999. Closing it.

PR #816, opened 2017-03-07
Fix: Redesign auth view(#815)

Ram81:
@aka-jain please review!
I've done some changes to login & signup page but there is still work to do in design of both of them,
and also I wanted to know do you want the footer to be there in these pages?

### DIFF
--- a/frontend/src/views/web/auth.html
+++ b/frontend/src/views/web/auth.html
@@ -1,7 +1,7 @@
 <div class="white-bg view-container flex-body">
 	<main-header></main-header>
 	<!-- EvalAI into -->
-	<div class="ev-container auth-container content">
+	<div class="ev-container grad-container content">
 		<!--auth container -->
 		<div class="ev-head-small">
 			<div class="row">
--- a/frontend/src/views/web/auth.html
+++ b/frontend/src/views/web/auth.html
@@ -1,21 +1,28 @@
-<!-- dynamic header for home page-->
-<dyn-header></dyn-header>
-<!-- background hero section -->
-<section class="bg-section">
-    <div class="bg-image"></div>
-</section>
-<!-- background hero section -->
-<section class="bg-section">
-    <div class="bg-image"></div>
-</section>
-<eval-loader></eval-loader>
-<!--auth container -->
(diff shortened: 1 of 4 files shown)


aka-jain:
@Ram81 please use classes that are already present in landing and about us page. Don't change styles of auth CSS.

Ram81:
@aka-jain  I needed to change the auth-container class otherwise the alignment was not as we expected do you want me to use grad-container class instead of auth-container like we did in about-us page?

aka-jain:
@Ram81 yes please, use those classes.

Ram81:
@aka-jain but if I use grad-conatiner class then alignment of the login & signup view to be loaded & also the spacemen container gets disturbed please suggest what should I do?

aka-jain:
That's why I've created a new issue for auth design. So you have to redesign the auth view again and use classes as we used in about us page. Do that first please.

Ram81:
@aka-jain I've used existing css classes only now & what about the footer?

aka-jain:
@Ram81 Add that too.

frontend/src/views/web/auth.html
```diff
@@ -1,21 +1,28 @@
...
-            <div class="reg-alert col s12 m4 offset-m4 align-left text-highlight w-300">{{auth.regMsg}}</div>
-            <!-- render login/signup form here -->
-            <div class="ui-view"></div>
-        </div>
-    </div>
-</section>
+<div class="white-bg view-container flex-body">
+	<main-header></main-header>
+	<!-- EvalAI into -->
+	<div class="ev-container grad-container content">
```
aka-jain: @Ram81 make a seperate div for `content` class.
p.s. just for readeability.

Ram81: @aka-jain sure

Ram81:
@aka-jain please review!

deshraj:
@Ram81 can you please share screenshot?

Ram81:
log

aka-jain:
@Ram81 That's not looking good. Let me redesign this page. Will share the auth design with you soon. 
Thanks for the efforts. I'm closing this PR.

PR #833, opened 2017-03-14
Docs: Add doc of how submission and submission worker works.

TroJan:
@deshraj


deshraj:
@TroJan let me know when you are done with this. We can iterate over this and build a solid documentation for the workers.

PR #846, opened 2017-03-19
Update readme.

AnshulBasia:
These were mainly the issues I faced when I was setting up just using the readme. I believe, readme should be self sufficient for setting up the environment. Please have a look. Thanks!


deshraj:
Thanks for adding this information @AnshulBasia. :)

PR #869, opened 2017-03-24
Jobs: Add output_metadata field to Submission model

deshraj:


deshraj:
Closing this PR since realized that this can be achieved by a much simpler method without adding a new field in the table.

PR #881, opened 2017-03-27
IsChallengeCreator decorator

### LINKED ISSUE #878: Use Decorator to check permission in Disable Challenge API
Currently for the disable challenge API, there is an explicit condition added to check if the user is creator of the challenge or not (https://github.com/Cloud-CV/EvalAI/blob/master/apps/challenges/views.py#L160). This has to be removed and the `isChallengeCreator` decorator has to be added to this view. 

The decorator can be found here https://github.com/Cloud-CV/EvalAI/blob/master/apps/challenges/permissions.py#L6

varunagrawal:
Fixes #878 

Also provides Python 3 compatibility for `scripts/workers/submission_worker.py`.


apps/challenges/migrations/0019_added_leaderboard_data_table.py
```diff
@@ -10,7 +10,7 @@
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('jobs', '0004_submission_output'),
+        ('jobs', '0001_squashed_0005_upload_unique_random_filename'),
```
deshraj: @varunagrawal This is not the part of this PR. Can you please open a separate PR for this?

apps/challenges/permissions.py
```diff
@@ -21,3 +24,31 @@ def has_permission(self, request, view):
                 return False
         else:
             return False
+
+
+def is_challenge_creator(func):
```
deshraj: I don't think there is a need for the decorator since it is already written on line 9 of the same file. Can you please use the same decorator?

apps/challenges/migrations/0019_added_leaderboard_data_table.py
varunagrawal: Alrighty.

apps/challenges/permissions.py
varunagrawal: Ah I misunderstood the PR. I see I have to use the django-rest-framework permissions decorator. Sorry about that.

varunagrawal:
@deshraj I have updated the PR. Thanks for bearing with my noobness on this project. :sweat_smile:

deshraj:
No worries. Glad that you are interested in contributing to the project.

apps/challenges/permissions.py
deshraj: No issues. :)

scripts/workers/submission_worker.py
```diff
@@ -102,7 +102,7 @@ def download_and_extract_file(url, download_location):
     try:
         response = requests.get(url)
     except Exception as e:
-        print 'Failed to fetch file from {}, error {}'.format(url, e)
+        print('Failed to fetch file from {}, error {}'.format(url, e))
```
deshraj: We haven't added support for Python 3.x as of now. So, can you please remove these changes related to Python 3.x

apps/challenges/urls.py
```diff
@@ -10,7 +10,7 @@
         views.challenge_detail, name='get_challenge_detail'),
     url(r'challenge/(?P<challenge_pk>[0-9]+)/participant_team/(?P<participant_team_pk>[0-9]+)',
         views.add_participant_team_to_challenge, name='add_participant_team_to_challenge'),
-    url(r'challenge/(?P<pk>[0-9]+)/disable',
+    url(r'challenge/(?P<challenge_pk>[0-9]+)/disable',
```
deshraj: There is no need to change this. Can you please revert this?

apps/challenges/urls.py
```diff
@@ -19,7 +19,7 @@
     # `A-Za-z` because it accepts either of `all, future, past or present` in either case
     url(r'challenge/(?P<challenge_time>[A-Za-z]+)$', views.get_all_challenges,
         name='get_all_challenges'),
-    url(r'challenge/(?P<pk>[0-9]+)/',
+    url(r'challenge/(?P<challenge_pk>[0-9]+)/',
```
deshraj: Can you please revert this also?

apps/challenges/views.py
```diff
@@ -147,8 +147,8 @@ def add_participant_team_to_challenge(request, challenge_pk, participant_team_pk
 @api_view(['POST'])
 @permission_classes((permissions.IsAuthenticated, HasVerifiedEmail))
 @authentication_classes((ExpiringTokenAuthentication,))
-def disable_challenge(request, pk):
-
+@permission_classes((IsChallengeCreator,))
+def disable_challenge(request, challenge_pk):
```
deshraj: Please revert this line to original.

scripts/workers/submission_worker.py
varunagrawal: Done.

apps/challenges/urls.py
varunagrawal: Done

apps/challenges/urls.py
varunagrawal: Done

apps/challenges/views.py
varunagrawal: Done. Sorry about that. I assumed a consistent naming scheme would make life easier.

apps/challenges/views.py
```diff
@@ -147,8 +147,8 @@ def add_participant_team_to_challenge(request, challenge_pk, participant_team_pk
 @api_view(['POST'])
 @permission_classes((permissions.IsAuthenticated, HasVerifiedEmail))
 @authentication_classes((ExpiringTokenAuthentication,))
+@permission_classes((IsChallengeCreator,))
 def disable_challenge(request, pk):
```
deshraj: @varunagrawal can you please add a test case for this view in which you test if the user is not the creator of the challenge then he is not allowed to disable the challenge. 

Let me know if you need some help with this.

varunagrawal: Yup was working on that right now. 

Also, there are two `permission_classes` decorators. Do you want me to merge the two into one?

varunagrawal: Actually nevermind. Noob question.

varunagrawal:
@deshraj I have a couple of questions. So when I use the permission decorator, it always returns `403 FORBIDDEN` instead of the `401 UNAUTHORIZED` error. Is this normal?

And for the test case where the challenge does not exist, the test case expects `406 NOT ACCEPTABLE`, but since the challenge does not exist, the permission returns false and thus I get a `403 FORBIDDEN` once again.

This is what the permission looks like now. I have added "POST" as a valid method as well as error handling when the challenge PK is invalid.
```python
class IsChallengeCreator(permissions.BasePermission):
    """
    Permission class for edit/delete a challenge.
    """
    message = "Sorry, you are not allowed to perform this operation!"

    def has_permission(self, request, view):

        if request.method in permissions.SAFE_METHODS:
            return True
        elif request.method in ['DELETE', 'PATCH', 'PUT', 'POST']:
            try:
                challenge = Challenge.objects.get(pk=request.parser_context['kwargs']['pk'])
            except Challenge.DoesNotExist:
                return False

            if request.user.id == challenge.creator.created_by.id:
                return True
            else:
                return False
        else:
            return False
```

Can you please confirm if this is indeed the expected behavior?

varunagrawal:
@deshraj I've added the updated tests and the modified permission. Can you please review the code to make sure my understanding behind this is correct?

varunagrawal:
So after some digging around, it seems that the django_rest_framework raised the PermissionDenied exception when the permission decorator returns false. This exception leads to a `403 FORBIDDEN` error by the framework itself and I can't seem to update that in a clean way.

apps/challenges/permissions.py
```diff
@@ -13,8 +13,12 @@ def has_permission(self, request, view):
 
         if request.method in permissions.SAFE_METHODS:
             return True
-        elif request.method in ['DELETE', 'PATCH', 'PUT']:
-            challenge = Challenge.objects.get(pk=request.parser_context['kwargs']['pk'])
+        elif request.method in ['DELETE', 'PATCH', 'PUT', 'POST']:
```
deshraj: @varunagrawal There is no need to make changes in the existing permissions :)

deshraj:
> So after some digging around, it seems that the django_rest_framework raised the PermissionDenied exception when the permission decorator returns false. This exception leads to a 403 FORBIDDEN error by the framework itself and I can't seem to update that in a clean way.

Yup, you are correct. If the user is not the Challenge Host of that particular challenge, then 403 is raised and that is what should happen actually.

apps/challenges/permissions.py
varunagrawal: If the POST method isn't allowed the first test fails where it is a legitimate disable request via a POST method :confused:

deshraj: Ah my bad. You are right.

tests/unit/challenges/test_views.py
```diff
@@ -443,17 +462,29 @@ def test_particular_challenge_for_disable_does_not_exist(self):
...
         response = self.client.post(self.url, {})
-        self.assertEqual(response.status_code, status.HTTP_406_NOT_ACCEPTABLE)
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
 
     def test_when_user_does_not_have_permission_to_disable_particular_challenge(self):
         self.url = reverse_lazy('challenges:disable_challenge',
                                 kwargs={'pk': self.challenge2.pk})
         expected = {
-            'error': 'Sorry, you do not have permission to disable this challenge'
+            'detail': 'Sorry, you are not allowed to perform this operation!'
```
deshraj: Can you please replace the key with `error`. This is because we follow a consistent error dictionary across the project.

tests/unit/challenges/test_views.py
```diff
@@ -443,17 +462,29 @@ def test_particular_challenge_for_disable_does_not_exist(self):
...
-        self.assertEqual(response.status_code, status.HTTP_401_UNAUTHORIZED)
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+    def test_disable_challenge_when_user_is_not_creator(self):
+        self.url = reverse_lazy('challenges:disable_challenge',
+                                kwargs={'pk': self.challenge2.pk})
+        expected = {
+            'detail': 'Sorry, you are not allowed to perform this operation!'
+        }
+        alt_client = APIClient(enforce_csrf_checks=True)
```
deshraj: There is no need to create another `APIClient` object since you can use the one created globally at https://github.com/Cloud-CV/EvalAI/pull/881/files/41ef119d19d112a2e8a53dc86e0461a403d9c181#diff-7b5c4a15a7006ac909ae8faa7b5d7e65R66

deshraj:
Rest all looks good. Great job :)

tests/unit/challenges/test_views.py
varunagrawal: The `detail` key is from the framework and alas I have no control over it. I can change the test case to only check the value instead of the whole dictionary. Would that be a good solution?

tests/unit/challenges/test_views.py
varunagrawal: If I understand this correctly, then I should add `self.user` as a `ChallengeHost` with the `self.challenge_host_team1` team, so `self.user` now has access to the challenge but is not the creator. Does this sound correct to you?

varunagrawal: Else, I'll have to set the user credentials of `self.user2` in the global `APIClient` and then reset it to the `self.user` object at the end of the test.

deshraj: > If I understand this correctly, then I should add self.user as a ChallengeHost with the self.challenge_host_team1 team, so self.user now has access to the challenge but is not the creator. Does this sound correct to you?

Yep, this is correct. Just make sure that the user is not the creator of the team. Rest all sounds good.

varunagrawal:
Updated.

deshraj:
Looks good. Great job @varunagrawal :)

varunagrawal:
Wouldn't have been possible without your guidance, @deshraj!

PR #882, opened 2017-03-27
Updated migration file to account for prior migration merge

varunagrawal:
Fix for the missing migration error in `jobs` caused due to the migration merge.

### DIFF
--- a/apps/challenges/migrations/0019_added_leaderboard_data_table.py
+++ b/apps/challenges/migrations/0019_added_leaderboard_data_table.py
@@ -10,7 +10,7 @@ import django.db.models.deletion
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('jobs', '0004_submission_output'),
+        ('jobs', '0001_squashed_0005_upload_unique_random_filename'),
         ('challenges', '0018_added_challenge_phase_split_table'),
     ]
 


varunagrawal:
@deshraj can you please review this?

deshraj:
I haven't reviewed this. Will check and merge it soon. Thanks for your patience.

TroJan:
@deshraj @varunagrawal : Anyone point me out clearly why is the change needed?

varunagrawal:
The migration step when setting up the project fails due to the missing migration dependency.

TroJan:
@varunagrawal : I did a fresh setup of the database part and faced no such error. Do you have any snapshot of it so that I can dig what am I missing?

varunagrawal:
You can find it right here: https://github.com/Cloud-CV/EvalAI/issues/878

You may have to drop the complete database when you setup from scratch. It may be that since Django saves migrations within the `django_migrations` table, you are not seeing this.

TroJan:
@varunagrawal : As stated earlier I did a fresh setup of the database(which means that I created my database again after dropping the existing one). @deshraj also pointed out that he is not facing this [error](https://github.com/Cloud-CV/EvalAI/issues/878#issuecomment-289349082).

Hence closing this as of now, will re-open if anyone else faces the same issue again. Thanks :)

varunagrawal:
@TroJan just did a clean install again. Still facing this issue. I am following the exact instructions in the README. Am I missing something?

varunagrawal:
I understood the problem. This is an issue with Python 3. Python 2 is able to correctly parse the changes and setup the migrations but it seems that Python 3 is having a hard time doing so. This will be problematic in the next quarter when Python 3 migration is decided upon.

TroJan:
@varunagrawal : Yeah we will definitely look into this when Python 3 support is being added. Thanks :)

deshraj:
@varunagrawal Glad that you finally figured out the problem. This will help us to  tackle future bugs due to Python 3.x

PR #883, opened 2017-03-27
Show the team name

varunagrawal:
Currently in the Django admin, when we wish to select the ChallengeHostTeam, the admin interface shows us `ChallengeHostTeam object` instead of the team name. This can be confusing for users who are a part of multiple teams.

This PR fixes this issue.

### DIFF
--- a/apps/hosts/models.py
+++ b/apps/hosts/models.py
@@ -17,6 +17,9 @@ class ChallengeHostTeam(TimeStampedModel):
     def __unicode__(self):
         return "%s: %s" % (self.team_name, self.created_by)
 
+    def __str__(self):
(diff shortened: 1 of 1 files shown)


deshraj:
@varunagrawal It looks like you are using Python 3.x but we haven't really added support for 3.x. So, I cannot merge this PR. It would be great if you can use Python 2.x for this project. 
We currently don't have any plans for supporting Python 3.x till the next quarter. We will let you know once we have this on our roadmap and then we can work this out.

Thanks :)

varunagrawal:
Yup, I realized that when running tests. I have set up a new venv with Python2. Thanks for the clarifications.

PR #905, opened 2017-04-03
Frontend: Add support for making submissions public on leaderboard

deshraj:


deshraj:
screen shot 2017-04-03 at 10 53 26 am

PR #911, opened 2017-04-05
Frontend: Fix business logic for showing std-err column

deshraj:


TroJan:
@deshraj "business logic" ?

PR #912, opened 2017-04-05
Challenges: Logs change in Challenge and Challenge Phase model when file fields are updated.

### LINKED ISSUE #839: Log change in Challenge and Challenge Phase model when file fields are updated
__Problem__

We need a way to figure out if the `evaluation_script`field value is changed for Challenge Model or not. Same goes for `test_annotation` field in ChallengePhase Model.

__Please be more clear__ ?

When ever a model is saved, its saved function is called. So this task involves figuring out a way in save model to detect if the model is updated or added. If the model is updated, then further it should be figured out whether the `evaluation_script` is changed for Challenge model or not.

If `evaluation_script` of Challenge model is changed, then simply print a message stating the change. This print statement will later on act as a logging statement in case we need to debug something.

The above requested thing goes same for `test_annotation` field of ChallengePhase model.

__Why__ ?

Submission worker when starts loads all `evaluation_script` and `test_annotation` file in advance. In case if the Challenge Host updates any of the file, there is no way to communicate this to the worker

__Impact__ ?

It actually guarantees that whenever any submission is evaluated, it is evaluated against the write set of parameters.

__How__ ?

As of now, I think it can be done via following ways

* Either overriding the save method of each model and adding a hook over there to check if the instance is being updated or added. May be there is a flag or something which tells whether the instance is added or updated

* Either by [signals](https://docs.djangoproject.com/en/1.10/topics/signals/). Reason I am asking to use signal because we need to carry out the same set of action for more than one models(presently _Challenge_ and _ChallengePhase_). Doing it by signal will help us achieve more modularized code.

* Your suggestion ??

__TakeAways__

About how django model is saved, how signal works, and a brief intro about RabbitMQ and submission worker

Also I will prefer that __before coding__, its better to __discuss__ the approach over here only, so that we can get an idea and update if something is not right.

RishabhJain2018:
fixes #839 . @deshraj @TroJan Please review the PR.

### DIFF
--- a/frontend/src/js/controllers/challengeCtrl.js
+++ b/frontend/src/js/controllers/challengeCtrl.js
@@ -503,8 +503,8 @@
             vm.phaseId = phaseId;
 
             var all_phases = vm.phases.results;
-            for(var i=0; i<vm.phases.results.length; i++){
-                if (all_phases[i].id == phaseId){
+            for (var i = 0; i < vm.phases.results.length; i++) {
+                if (all_phases[i].id == phaseId) {
                     vm.currentPhaseLeaderboardPublic = all_phases[i].leaderboard_public;
                     break;
                 }
@@ -547,7 +547,7 @@
                     var details = response.data;
                     vm.submissionResult = details;
 
-                    for (var i = 0; i < details.results.length; i++){
+                    for (var i = 0; i < details.results.length; i++) {
                         vm.submissionVisibility[details.results[i].id] = details.results[i].is_public;
                     }
 
@@ -651,7 +651,7 @@
                 vm.stopFetchingSubmissions();
                 vm.poller = $interval(function() {
                     var parameters = {};
-                    parameters.url = "jobs/challenge/" + vm.challengeId + "/challenge_phase/" + vm.phaseId + "/submission/";
+                    parameters.url = "jobs/challenge/" + vm.challengeId + "/challenge_phase/" + vm.phaseId + "/submission/?page=" + Math.ceil(vm.currentPage);
                     parameters.method = 'GET';
                     parameters.data = {};
                     parameters.token = userKey;
@@ -660,7 +660,7 @@
                             var details = response.data;
 
                             // Set the is_public flag corresponding to each submission
-                            for (var i = 0; i < details.results.length; i++){
+                            for (var i = 0; i < details.results.length; i++) {
                                 vm.submissionVisibility[details.results[i].id] = details.results[i].is_public;
                             }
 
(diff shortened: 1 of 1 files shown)


apps/challenges/models.py
```diff
@@ -12,6 +14,11 @@
 class Challenge(TimeStampedModel):
 
     """Model representing a hosted Challenge"""
+
+    def __init__(self, *args, **kwargs):
+        super(Challenge, self).__init__(*args, **kwargs)
+        self.__original_evaluation_script = self.evaluation_script
```
TroJan: Is there any particular reason for using double underscore ? As far as I know, using a single underscore would have conveyed the meaning. What are your thoughts on this @deshraj ?

apps/challenges/models.py
```diff
@@ -73,6 +80,14 @@ def is_active(self):
         return False
 
 
+@receiver(signals.post_save, sender=Challenge)
```
TroJan: The reason for using signal was a common function which can be applied across all models. This function is ok but I will really be more happy if you could somehow dig and see if `create_post_challenge` and `create_post_challengephase` can be made a single function.

apps/challenges/models.py
```diff
@@ -73,6 +80,14 @@ def is_active(self):
...
+def create_post_challenge(sender, instance, created, **kwargs):
+    if instance._Challenge__original_evaluation_script is False:
+        print "First time evaluation script added !"
```
TroJan: Can we use `logger` here. I know these print statement will be removed in the future, but as of now lets try following practices and use logger.

apps/challenges/models.py
```diff
@@ -98,7 +118,7 @@ class ChallengePhase(TimeStampedModel):
     challenge = models.ForeignKey('Challenge')
     is_public = models.BooleanField(default=False)
     is_submission_public = models.BooleanField(default=False)
-    test_annotation = models.FileField(upload_to=RandomFileName("test_annotations"))
+    test_annotation = models.FileField(upload_to=RandomFileName("test_annotations"), default=False)
```
TroJan: Can you explain me why default=False is required here?

Moreover if this is required, I will prefer that you create a separate PR for it

apps/challenges/models.py
deshraj: Yeah, let's follow the convention that we have been following i.e use single underscores.

RishabhJain2018: @deshraj @TroJan Actually, I used that to be specific about the class private member names. Please refer [here](http://stackoverflow.com/questions/1301346/what-is-the-meaning-of-a-single-and-a-double-underscore-before-an-object-name). Moreover, I didn't know about the convention we have been following. I will update the PR accordingly.

apps/challenges/models.py
RishabhJain2018: @TroJan Yeah sure I will work on it.

apps/challenges/models.py
RishabhJain2018: @TroJan I added it as I am checking the condition of first time upload of a `test_annotation`. Please refer [here](https://github.com/Cloud-CV/EvalAI/pull/912/files#diff-06139830cb2f143920edd3015bb7b1baR153).
Can I create a new issue for it ?

apps/challenges/models.py
deshraj: I think we are good with making it `semiprivate` instead of making it `superprivate`. So, we would go with single underscore.

apps/challenges/models.py
deshraj: Yeah please create a new issue and lets get that part merged first before merging this one.

deshraj:
Ummm...I don't see a need to close this PR.

RishabhJain2018:
ohh...I accidentally closed it. I will reopen it once I will incorporate all the changes of the above discussion.

deshraj:
Sounds good.

PR #919, opened 2017-04-08
Docs: Add doc of how submission and submission worker works.

TroJan:


TroJan:
@deshraj : we can iterate over this and see what all more needs to be added.

deshraj:
Sure. Sounds good to me.

PR #928, opened 2017-04-11
[REVIEW PENDING] Challenges: Logs change in Challenge and Challenge Phase model when file fields are updated.

### LINKED ISSUE #839: Log change in Challenge and Challenge Phase model when file fields are updated

RishabhJain2018:
fixes #839.
This PR also incorporates the reviews from PR #912 .

### DIFF
--- a/apps/base/models.py
+++ b/apps/base/models.py
@@ -21,26 +21,22 @@ class TimeStampedModel(models.Model):
         app_label = 'base'
 
 
-"""
-The decorator below is used to pass model field names to django signals function.
-"""
-
-
-def extra_args(field_name, *args, **kwargs):
-    def extra_args_decorator(f, *args, **kwargs):
-        def extra_args_wrapper(sender, instance, **kwargs):
+def model_field_name(field_name, *args, **kwargs):
+    """
+    The decorator is used to pass model field names to create_post_model_field function for logging change.
+    """
+    def model_field_name_decorator(f, *args, **kwargs):
+        def model_field_name_wrapper(sender, instance, **kwargs):
             f(sender, instance, field_name=field_name, **kwargs)
-        return extra_args_wrapper
-    return extra_args_decorator
-
-
-"""
-The function below is used to log the change when any attribute of model field instance is changed.
-"""
+        return model_field_name_wrapper
+    return model_field_name_decorator
 
 
 def create_post_model_field(sender, instance, field_name, **kwargs):
-    if getattr(instance, "_original_{}".format(field_name)) is False:
-        logger.info("{} is added first time !".format(field_name))
+    """
+    When any model field value changes, it is used to log the change.
+    """
+    if getattr(instance, '_original_{}'.format(field_name)) is False:
+        logger.info('{} for {} is added first time !'.format(field_name, instance.pk))
     else:
-        logger.info("{} is changed !".format(field_name))
+        logger.info('{} for {} changed !'.format(field_name, instance.pk))
--- a/apps/base/models.py
+++ b/apps/base/models.py
@@ -1,6 +1,11 @@
 from __future__ import unicode_literals
 
+import logging
+
 from django.db import models
+
+# Get an instance of a logger
+logger = logging.getLogger(__name__)
 
 
 class TimeStampedModel(models.Model):
@@ -26,6 +31,6 @@ def extra_args(fragment_name, *args, **kwargs):
 
 def create_post_model_field(sender, instance, fragment_name, **kwargs):
     if getattr(instance, "_original_" + str(fragment_name)) is False:
-        print str(fragment_name) + " is added first time !"
+        logger.info(str(fragment_name) + " is added first time !")
     else:
-        print str(fragment_name) + " is changed !"
+        logger.info(str(fragment_name) + " is changed !")
--- a/apps/base/models.py
+++ b/apps/base/models.py
@@ -21,16 +21,16 @@ class TimeStampedModel(models.Model):
         app_label = 'base'
 
 
-def extra_args(fragment_name, *args, **kwargs):
-    def extra_args_inner1(f, *args, **kwargs):
-        def extra_args_inner2(sender, instance, **kwargs):
-            f(sender, instance, fragment_name=fragment_name, **kwargs)
-        return extra_args_inner2
-    return extra_args_inner1
+def extra_args(field_name, *args, **kwargs):
+    def extra_args_decorator(f, *args, **kwargs):
+        def extra_args_wrapper(sender, instance, **kwargs):
+            f(sender, instance, field_name=field_name, **kwargs)
+        return extra_args_wrapper
+    return extra_args_decorator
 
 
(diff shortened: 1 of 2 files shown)


RishabhJain2018:

apps/base/models.py
```diff
@@ -14,3 +19,18 @@ class TimeStampedModel(models.Model):
...
+def extra_args(fragment_name, *args, **kwargs):
+    def extra_args_inner1(f, *args, **kwargs):
+        def extra_args_inner2(sender, instance, **kwargs):
+            f(sender, instance, fragment_name=fragment_name, **kwargs)
+        return extra_args_inner2
+    return extra_args_inner1
+
+
+def create_post_model_field(sender, instance, fragment_name, **kwargs):
+    if getattr(instance, "_original_" + str(fragment_name)) is False:
```
deshraj: Convert `str(fragment_name)` -> `fragment_name`. There is no need to use `str()`

deshraj: Also, please use strong formatting instead of `+`

apps/base/models.py
```diff
@@ -14,3 +19,18 @@ class TimeStampedModel(models.Model):
     class Meta:
         abstract = True
         app_label = 'base'
+
+
+def extra_args(fragment_name, *args, **kwargs):
+    def extra_args_inner1(f, *args, **kwargs):
```
TroJan: Can we make function names better?

apps/base/models.py
```diff
@@ -14,3 +19,18 @@ class TimeStampedModel(models.Model):
...
+        def extra_args_inner2(sender, instance, **kwargs):
+            f(sender, instance, fragment_name=fragment_name, **kwargs)
```
TroJan: Instead of `fragment_name` what if we take parameter as `field_name`

TroJan:
@RishabhJain2018 : overall looks nice, Have left some minor comments for you to work upon. :)

RishabhJain2018:

apps/base/models.py
RishabhJain2018: yeah sure.

deshraj:
@RishabhJain2018 can you please write tests for this? We are not comfortable in merging this without having tests for the models.

RishabhJain2018:
@TroJan @deshraj Please guide me through how to write test cases for `django-signals`.

deshraj:
@taranjeet can you review this and get this done? Also, it would be great if you can figure out if there is a need to write test cases for this or not.

taranjeet:
@deshraj : sure will look into this.

apps/base/models.py
```diff
@@ -14,3 +19,28 @@ class TimeStampedModel(models.Model):
...
+def extra_args(field_name, *args, **kwargs):
+    def extra_args_decorator(f, *args, **kwargs):
+        def extra_args_wrapper(sender, instance, **kwargs):
+            f(sender, instance, field_name=field_name, **kwargs)
+        return extra_args_wrapper
+    return extra_args_decorator
+
+
+"""
+The function below is used to log the change when any attribute of model field instance is changed.
```
taranjeet: Can we shift this doc string inside the function? Also it will be great if we can update it to

```
when any model field value changes
```

apps/base/models.py
```diff
@@ -14,3 +19,28 @@ class TimeStampedModel(models.Model):
...
+"""
+
+
+def create_post_model_field(sender, instance, field_name, **kwargs):
+    if getattr(instance, "_original_{}".format(field_name)) is False:
```
taranjeet: Can we please use single quotes everywhere?

apps/base/models.py
```diff
@@ -14,3 +19,28 @@ class TimeStampedModel(models.Model):
     class Meta:
         abstract = True
         app_label = 'base'
+
+
+"""
```
taranjeet: This docstring can be moved within the function only.

taranjeet: Also docstring should be updated. Here `django signals function` is a bit confusing. Can we find something else for it?

apps/base/models.py
```diff
@@ -14,3 +19,28 @@ class TimeStampedModel(models.Model):
...
+The decorator below is used to pass model field names to django signals function.
+"""
+
+
+def extra_args(field_name, *args, **kwargs):
```
taranjeet: One more thing, can we use some other function name. `extra_args` and its related functions - these names are fine, just that they dont give any impression of what they are meant for.

apps/base/models.py
```diff
@@ -14,3 +19,28 @@ class TimeStampedModel(models.Model):
...
+
+
+"""
+The function below is used to log the change when any attribute of model field instance is changed.
+"""
+
+
+def create_post_model_field(sender, instance, field_name, **kwargs):
+    if getattr(instance, "_original_{}".format(field_name)) is False:
+        logger.info("{} is added first time !".format(field_name))
```
taranjeet: Though this message is just an informatory message, but since we are getting information over here, so what if we log `instance.pk` in the log statements ?

apps/challenges/models.py
```diff
@@ -74,6 +80,10 @@ def is_active(self):
         return False
 
 
+signals.post_save.connect(extra_args(field_name="evaluation_script")
```
taranjeet: Also can we please make sure to use single quotes everywhere?

RishabhJain2018:

taranjeet:
@RishabhJain2018 : thanks for the fix :)

PR #938, opened 2017-04-16
Added lint to start of dev:runserver task

aka-jain:
@spyshiv Have a look please!


spyshiv:
LGTM :) . 
great catch @aka-jain

PR #942, opened 2017-04-21
Bug Fix: Fixes unicode error on django admin

deshraj:


deshraj:
Currently, since we have `__str__()` method for some of the model definitions and it causes `UnicodeEncodeError : 'ascii' codec can't encode characters in position 0-7: ordinal not in range(128)` and we have 2.X Python version on the production server. So, changing the method name to `__unicode__()`.

PR #944, opened 2017-04-22
Frontend: Modify navigation bar options

deshraj:


deshraj:
Part of #945. Closing it.

PR #945, opened 2017-04-22
Frontend: Modify Getting Involved Page

deshraj:


dexter1691:
In general, whenever you are committing lots of text, try to split the text into multiple lines so that minor changes to the text can be identified by git as line diffs and reduce conflicts. Because if the entire para is one line, even if you make change to a character the whole line is considered to be changed.

PR #949, opened 2017-04-22
Fix #948: User is allowed to participate in a challenge with one team

### LINKED ISSUE #948: Teams: A user can participate in a Challenge through multiple teams
#### Steps to reproduce
1. User X participated in Challenge C with team T1
2. User Y created a team T2
3. User Y invited User X to team T2
4. User Y can participate in challenge C with team T2

### Expected behavior
Step 4 should not be allowed since user X has already participated in the challenge with Team T1. And hence user Y should be asked to choose a new team to participate in the challenge.

deshraj:
Fixes #948


dexter1691:
Does this handle the modification part too? If the user who had already participated tries to be added in another team, it should also fail. 

For now, let's not allow users to be added into the team even if he deletes his previous team. So basically check for all users from deleted teams / existing teams. Will keep the logic simple

PR #966, opened 2017-04-25
[REVIEW PENDING] Jobs:API to show number of remaining submissions on submission page.

### LINKED ISSUE #914: Show number of remaining submissions on Submission Page
### USE CASE

Whenever a user tries to make a submission to a particular phase, then we should show these details: 
- Number of submissions he can do for that particular day
- Number of overall remaining submissions he can do in total

### TODO

- [ ] Add support of query params where the keys can be `date`(date in particular format DD-MM-YYYY comes here)
- [ ] On Challenge Submission view, calculate the remaining daily submissions and remaining total submissions using the results of challenge phase API and submission API on the client side and bind it with the view. This will keep getting updated as the user chooses a different phase.
- [ ] Also show the time remaining so that the user knows that when he can make his next submission to that particular phase.

Please let me know if something is unclear.

RishabhJain2018:
fixes #914 . @deshraj @TroJan Please review the PR.


apps/jobs/views.py
```diff
@@ -229,3 +231,72 @@ def leaderboard(request, challenge_phase_split_id):
...
+    try:
+        ParticipantTeam.objects.get(pk=participant_team_id)
+    except ParticipantTeam.DoesNotExist:
+        response_data = {'error': 'You haven\'t participated in the challenge'}
+        return Response(response_data, status=status.HTTP_403_FORBIDDEN)
+
+    max_submission_per_day = ChallengePhase.objects.get(
+            pk=challenge_phase_id).max_submissions_per_day
+
+    max_submission = ChallengePhase.objects.get(
```
deshraj: Please don't query the database again. :)

apps/jobs/views.py
```diff
@@ -229,3 +231,72 @@ def leaderboard(request, challenge_phase_split_id):
...
+            pk=challenge_phase_id).max_submissions
+
+    submissions_done_today_count = Submission.objects.filter(
+        challenge_phase__challenge=challenge_id,
+        participant_team=participant_team_id,
+        challenge_phase=challenge_phase_id,
+        submitted_at__gte=datetime.date.today()).count()
+
+    failed_count = Submission.objects.filter(
```
deshraj: `failed_count` -> `failed_submissions_count`

apps/jobs/views.py
```diff
@@ -229,3 +231,72 @@ def leaderboard(request, challenge_phase_split_id):
...
+        challenge_phase=challenge_phase_id,
+        participant_team=participant_team_id,
+        challenge_phase__challenge=challenge_id,
+        status=Submission.FAILED,
+        submitted_at__gte=datetime.date.today()).count()
+
+    if (submissions_done_today_count - failed_count) >= max_submission_per_day or max_submission_per_day == 0:
+        dtnow = datetime.datetime.now()
```
deshraj: Please use meaningful variable names.

apps/jobs/views.py
```diff
@@ -229,3 +231,72 @@ def leaderboard(request, challenge_phase_split_id):
...
+        dttomorrow = dtnow + datetime.timedelta(days=1)
+        midnight = dttomorrow.replace(hour=0, minute=0, second=0)
+        remaining_time = midnight - dtnow
```
deshraj: A better way of calculating would be 

```python
from django.utils import timezone
from datetime import timedelta, datetime, time

now = timezone.now()
date_tomorrow = today.date() + timedelta(1)
midnight =  datetime.combine(date_tomorrow, time())
time_left  = midnight - now
# Finally user time_left accordingly in your code 
```

This will make sure that you are using the `timezone` module that is Timezone Aware.

apps/jobs/views.py
```diff
@@ -229,3 +231,72 @@ def leaderboard(request, challenge_phase_split_id):
...
+        response_data = {'message': 'You have exhausted the submission limit for today',
```
deshraj: Change message to `You have exhausted today's submission limit`

apps/jobs/views.py
```diff
@@ -229,3 +231,72 @@ def leaderboard(request, challenge_phase_split_id):
...
+                         'remaining_time': remaining_time.total_seconds()
+                         }
+        return Response(response_data, status=status.HTTP_200_OK)
+    else:
+        remaining_submission_per_day = max_submission_per_day - (submissions_done_today_count - failed_count)
+        remaining_submission = max_submission - (submissions_done_today_count - failed_count)
```
deshraj: `remaining_submission` -> `remaining_submission_count`

apps/jobs/views.py
```diff
@@ -229,3 +231,72 @@ def leaderboard(request, challenge_phase_split_id):
...
+        dtnow = datetime.datetime.now()
+        dttomorrow = dtnow + datetime.timedelta(days=1)
+        midnight = dttomorrow.replace(hour=0, minute=0, second=0)
+        remaining_time = midnight - dtnow
+        response_data = {'message': 'You have exhausted the submission limit for today',
+                         'remaining_time': remaining_time.total_seconds()
+                         }
+        return Response(response_data, status=status.HTTP_200_OK)
+    else:
+        remaining_submission_per_day = max_submission_per_day - (submissions_done_today_count - failed_count)
```
deshraj: `remaining_submission_per_day` -> `remaining_submissions_today_count`

deshraj:
@RishabhJain2018 any updates on this?

RishabhJain2018:
@deshraj I apologise for the delay. I will fix the requested changes asap.

deshraj:
No worries. Just please keep me posted about the same.

apps/jobs/views.py
```diff
@@ -229,3 +231,72 @@ def leaderboard(request, challenge_phase_split_id):
...
+@throttle_classes([UserRateThrottle])
+@api_view(['GET'])
+@permission_classes((permissions.IsAuthenticated, HasVerifiedEmail))
+@authentication_classes((ExpiringTokenAuthentication,))
+def remaining_submission(request, challenge_phase_id, participant_team_id, challenge_id):
+    try:
+        Challenge.objects.get(pk=challenge_id)
+    except Challenge.DoesNotExist:
+        response_data = {'error': 'Challenge does not exist'}
+        return Response(response_data, status=status.HTTP_400_BAD_REQUEST)
```
RishabhJain2018: yean sure. :+1:

apps/jobs/urls.py
```diff
@@ -9,6 +9,10 @@
     url(r'challenge/(?P<challenge_id>[0-9]+)/'
         r'challenge_phase/(?P<challenge_phase_id>[0-9]+)/submission/',
         views.challenge_submission, name='challenge_submission'),
+    url(r'challenge/(?P<challenge_id>[0-9]+)/'
+        r'challenge_phase/(?P<challenge_phase_id>[0-9]+)/participant_team/(?P<participant_team_id>[0-9]+)/'
+        r'get_remaining_submission/',
+        views.remaining_submission, name='remaining_submission'),
```
TroJan: @deshraj : May I know why are we using hyphen instead of underscore in urls. As far as all the urls are concerned in the project, I think we are using underscore.

apps/jobs/views.py
```diff
@@ -229,3 +231,72 @@ def leaderboard(request, challenge_phase_split_id):
...
+            pk=challenge_phase_id).max_submissions_per_day
+
+    max_submission = ChallengePhase.objects.get(
+            pk=challenge_phase_id).max_submissions
+
+    submissions_done_today_count = Submission.objects.filter(
+        challenge_phase__challenge=challenge_id,
+        participant_team=participant_team_id,
+        challenge_phase=challenge_phase_id,
+        submitted_at__gte=datetime.date.today()).count()
```
TroJan: Though we are running the production server in UTC, but still I will prefer that we should `timezone` module from utc, so that no one faces date time time related warning.

apps/jobs/urls.py
deshraj: I realized that we have been using underscores but soon we have to refactor the URLs to use hyphens instead of underscores. I have been reading about naming the URL endpoints and we should stick to the best practices.

TroJan: Can you point me to one such link(apart from that SEO one) ?

deshraj: See this http://stackoverflow.com/questions/10302179/hyphen-underscore-or-camelcase-as-word-delimiter-in-uris

Please read the first three answers.

TroJan: Okay nice, @RishabhJain2018 ignore these changes then.

@deshraj: Lets keep this in mind that we need to port all urls to follow hyphen scheme.

RishabhJain2018:

apps/jobs/views.py
```diff
@@ -51,22 +56,25 @@ def challenge_submission(request, challenge_id, challenge_phase_id):
         return Response(response_data, status=status.HTTP_400_BAD_REQUEST)
 
     if request.method == 'GET':
-        # getting participant team object for the user for a particular challenge.
+        # getting participant team object for the user for a particular
```
TroJan: I think these changes are because of running autopep8. Can you make sure that these are removed from the PR ?

apps/jobs/views.py
```diff
@@ -224,8 +238,78 @@ def leaderboard(request, challenge_phase_split_id):
...
     response_data = result_page
     return paginator.get_paginated_response(response_data)
+
+
+@throttle_classes([UserRateThrottle])
+@api_view(['GET'])
+@permission_classes((permissions.IsAuthenticated, HasVerifiedEmail))
+@authentication_classes((ExpiringTokenAuthentication,))
+def get_remaining_submissions(request, challenge_phase_pk, participant_team_pk, challenge_pk):
+    challenge = get_challenge_object(challenge_pk)
```
TroJan: @RishabhJain2018 : I feel that the below parts need some more thoughts and probably rewrite from you. Can you please go through it and see if it adheres to the concerned task

apps/jobs/utils.py
```diff
@@ -0,0 +1,32 @@
...
+    Challenge,
+)
+from participants.models import (ParticipantTeam,)
+
+
+def get_challenge_object(challenge_pk):
+    try:
+        Challenge.objects.get(pk=challenge_pk)
+    except Challenge.DoesNotExist:
+        response_data = {'error': 'Challenge does not exist'}
```
TroJan: String should be something like `{model_name} {pk} does not exist`. Clearing this out so that no confusion arises later on!

apps/jobs/utils.py
```diff
@@ -0,0 +1,32 @@
...
+        return Response(response_data, status=status.HTTP_400_BAD_REQUEST)
+
+
+def get_challenge_phase_object(challenge_phase_pk):
```
RishabhJain2018: @TroJan Yeah sure. I think for this we can follow a convention `get_{app_name}_{model_name}_object`. What do you say ?

apps/jobs/views.py
RishabhJain2018: @TroJan Yeah sure.

apps/jobs/utils.py
TroJan: I dont think there is need of `app_name` here since we are already namespacing this functions by keeping them in `utils.py` of each app. That is actually more than enough. So with this anyone wanting to use will write something like

`from apps.challenges.utils import get_challenge_object`

One more thing to keep in mind is that these functions should simply return a object or raise an exception if obejct is not found. In any case I dont think it makes sense to return Response object as that will bring a hassle to the callee of keeping explicit checks of value being returned. This may be debatable @deshraj : your thoughts required here.

deshraj: I think we can raise an API exception whenever the object is not found. We do the similar thing while saving the submission (see https://github.com/Cloud-CV/EvalAI/blob/master/apps/jobs/models.py#L157). This looks like a neat approach for me and we can proceed with this.

TroJan: So function will look something like
```
def get_challenge_object(pk):
    try:
        challenge=Challenge.objects.get(pk=pk)
        return challenge
    except Challenge.DoesNotExist as e:
       raise e
```

If this is right, Can you tell me what code will the callee have to write @deshraj ?

deshraj:
```
challenge_obj = get_challenge_object(<pk>)
```

I think if there would be an exception, then the API exception will be thrown and it won't proceed further. Please let me know if I am wrong or missing something here.

TroJan: Yeah this is fine.

TroJan: Also, I think these utils functions should not be a part of this PR. They should go in a separate PR on per app basis(only raise PR for functions which are required as of now in this PR).

TroJan: @RishabhJain2018 : Since with time these functions will grow, so we can create these functions dynamically(can be kept in base app) and then in utils of each app, call that function to get `get_*_object` function.

This way we prevent ourselves from complaints of code climate also 😄

RishabhJain2018: @TroJan I didn't get your last point of creating these functions dynamically ?

TroJan: See if [this](https://stackoverflow.com/a/13184536/2534102) helps. What I want is basically a outer functions which receives argument like name of model(may be something else) and then it creates a function(which does the object exist part and raising exception one).

RishabhJain2018: Yeah.I got it.

RishabhJain2018:

apps/jobs/urls.py
```diff
@@ -9,6 +9,10 @@
     url(r'challenge/(?P<challenge_id>[0-9]+)/'
```
taranjeet: The title of the PR says "api to show number of remaining submissions" but it also involves changes related to other apis also. Most of the changes are because of running autopep8. Can you please remove those changes so that this PR is only concerned with remaining submission api?

RishabhJain2018: ok sure.

RishabhJain2018:
@deshraj @taranjeet please review the PR.

apps/jobs/views.py
```diff
@@ -228,3 +232,65 @@ def leaderboard(request, challenge_phase_split_id):
...
+        remaining_submissions_today_count = (max_submission_per_day -
+                                             (submissions_done_today_count -
+                                              failed_submissions_count)
+                                             )
+        remaining_submission_count = max_submission - \
+            (submissions_done_today_count - failed_submissions_count)
+        response_data = {'remaining_submissions_today_count': remaining_submissions_today_count,
+                         'remaining_submissions': remaining_submission_count
+                         }
+        return Response(response_data, status=status.HTTP_200_OK)
```
deshraj: Please add a newline here.

apps/jobs/views.py
```diff
@@ -228,3 +232,65 @@ def leaderboard(request, challenge_phase_split_id):
...
+        participant_team=participant_team_pk,
+        challenge_phase__challenge=challenge_pk,
+        status=Submission.FAILED,
+        submitted_at__gte=timezone.now().date()).count()
+
+    if ((submissions_done_today_count - failed_submissions_count) >= max_submission_per_day
+            or (max_submission_per_day == 0)):
+        date_time_now = timezone.now()
+        date_time_tomorrow = date_time_now.date() + datetime.timedelta(1)
+        utc = pytz.UTC
```
deshraj: We should use the timezone from the settings file instead of using from pytz.

taranjeet:
@RishabhJain2018 : let us know when the changes are done, so that we can review it. :)

apps/jobs/views.py
RishabhJain2018: @taranjeet I didn't get what exactly @deshraj wanted to say? I mean how we can get time from settings module as we are using `postgres` as database. Please refer here: https://docs.djangoproject.com/en/1.10/ref/settings/#time-zone

apps/jobs/views.py
```diff
@@ -228,3 +231,65 @@ def leaderboard(request, challenge_phase_split_id):
...
+    failed_submissions_count = Submission.objects.filter(
+        challenge_phase=challenge_phase_pk,
+        participant_team=participant_team_pk,
+        challenge_phase__challenge=challenge_pk,
+        status=Submission.FAILED,
+        submitted_at__gte=timezone.now().date()).count()
+
+    if ((submissions_done_today_count - failed_submissions_count) >= max_submission_per_day
+            or (max_submission_per_day == 0)):
+        date_time_now = timezone.now()
```
taranjeet: Can you also explain this part?

Also please write in comment of what and why are you doing this. Trying keeping the `what` part to the point.

tests/unit/jobs/test_views.py
```diff
@@ -371,3 +372,152 @@ def test_get_challenge_submissions(self):
...
+        response = self.client.get(self.url, {})
+        self.assertEqual(response.data, expected)
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+    def test_get_remaining_submission(self):
+        self.url = reverse_lazy('jobs:get_remaining_submissions',
+                                kwargs={'challenge_phase_pk': self.challenge_phase.pk,
+                                        'challenge_pk': self.challenge.pk})
+
+        submissions_done_today_count = Submission.objects.filter(
```
taranjeet: Since these are test cases, so ideally no logic should be written here. test cases should be aware of what the input will be and what will be the output. In your case, it is pretty obvious that there is a single failed and successful submission. That knowledge is enough for writing test cases.

Can you please update the test case also?

tests/unit/jobs/test_views.py
```diff
@@ -371,3 +372,152 @@ def test_get_challenge_submissions(self):
...
+        response = self.client.get(self.url, {})
+        self.assertEqual(response.data, expected)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+    def test_get_remaining_submission_time_when_limit_is_exhausted(self):
+        self.url = reverse_lazy('jobs:get_remaining_submissions',
+                                kwargs={'challenge_phase_pk': self.challenge_phase.pk,
+                                        'challenge_pk': self.challenge.pk})
+
+        submissions_done_today_count = Submission.objects.filter(
```
taranjeet: Above comment goes here also

apps/jobs/views.py
```diff
@@ -228,3 +231,65 @@ def leaderboard(request, challenge_phase_split_id):
...
+    challenge_model = get_model_object(Challenge)
+    challenge_model(challenge_pk)
+
+    challenge_phase_model = get_model_object(ChallengePhase)
+    challenge_phase = challenge_phase_model(challenge_phase_pk)
+
+    participant_team_pk = get_participant_team_id_of_user_for_a_challenge(
+        request.user, challenge_pk)
+
+    try:
```
RishabhJain2018: Yes, this check validates that participant model exists. But we are not returning any response if the participant team for the user doesn't exists, so I added this `try-except` part. Please let me know if I am wrong somewhere?

apps/jobs/views.py
RishabhJain2018: Yes, In this part I am calculating the remaining time of the participant when the limit has been exhausted. It is done in the following way:
1. If the submissions that are done today - failed submissions is greater than the given maximum submissions per day then
2. variable date_time_now will take the utc time of that instant when this condition is true.
3. Now , I am calculating the midnight time & then subtracting the date_time_now from the midnight time to get the remaining time.
Please check if something is still not clear.

RishabhJain2018: Moreover, I will add comment for this too :+1:

apps/jobs/views.py
taranjeet: `None` is being returned. Can we add a check on it?

RishabhJain2018: check in the sense that we should return a response from there only, right? This will reduce an extra query to be executed on the database.

taranjeet: yeah, also write a comment stating that check for existence of participant team is deduced from above function. You can rephase the comment in your own words.

RishabhJain2018: yeah sure.

apps/jobs/views.py
```diff
@@ -228,3 +231,65 @@ def leaderboard(request, challenge_phase_split_id):
...
+    except ParticipantTeam.DoesNotExist:
+        response_data = {'error': 'You haven\'t participated in the challenge'}
+        return Response(response_data, status=status.HTTP_403_FORBIDDEN)
+
+    max_submission_per_day = challenge_phase.max_submissions_per_day
+
+    max_submission = challenge_phase.max_submissions
+
+    submissions_done_today_count = Submission.objects.filter(
+        challenge_phase__challenge=challenge_pk,
```
RishabhJain2018: I didn't get the question. Please elaborate a bit?

taranjeet: To extract submission_done_today_count, you are passing `challenge_pk` in the params where as for failed_submission_count, you are passing `challenge_phase_pk`. May I know the reason for such ?

RishabhJain2018:
```
    submissions_done_today_count = Submission.objects.filter(
        challenge_phase__challenge=challenge_pk,
        participant_team=participant_team_pk,
        challenge_phase=challenge_phase_pk,
        submitted_at__gte=timezone.now().date()).count()

```
I guess the confusion is due to reordering of the params in both queries. Both the queries are almost same with a addition of` status` field in the below query.

    ```
failed_submissions_count = Submission.objects.filter(
        challenge_phase=challenge_phase_pk,
        participant_team=participant_team_pk,
        challenge_phase__challenge=challenge_pk,
        status=Submission.FAILED,
        submitted_at__gte=timezone.now().date()).count()

```
Please check once again.

Comment (taranjeet) [inline apps/jobs/views.py]: Ohh I just saw that. Please re-order the fields, so that anyone else does not have the same confusion

File: tests/unit/jobs/test_views.py [hunk 23]
```diff
@@ -371,3 +371,113 @@ def test_get_challenge_submissions(self):
...
+    def test_get_remaining_submission_when_challenge_does_not_exist(self):
+        self.url = reverse_lazy('jobs:get_remaining_submissions',
+                                kwargs={'challenge_phase_pk': self.challenge_phase.pk,
+                                        'challenge_pk': self.challenge.pk})
+
+        expected = {
+            'detail': 'Challenge {} does not exist'.format(self.challenge.pk)
+        }
+
+        self.challenge.delete()
```
taranjeet: I dont think there is a need to delete this challenge. Simply ask for `self.challenge.pk + 1` in the url, I think it will be sufficient.

tests/unit/jobs/test_views.py
```diff
@@ -371,3 +371,113 @@ def test_get_challenge_submissions(self):
...
+    def test_get_remaining_submission_when_challenge_phase_does_not_exist(self):
+        self.url = reverse_lazy('jobs:get_remaining_submissions',
+                                kwargs={'challenge_phase_pk': self.challenge_phase.pk,
+                                        'challenge_pk': self.challenge.pk})
+
+        expected = {
+            'detail': 'ChallengePhase {} does not exist'.format(self.challenge_phase.pk)
+        }
+
+        self.challenge_phase.delete()
```
taranjeet: Above comment goes for here also.

tests/unit/jobs/test_views.py
```diff
@@ -371,3 +371,113 @@ def test_get_challenge_submissions(self):
...
+        self.challenge.save()
+        response = self.client.get(self.url, {})
+        self.assertEqual(response.data, expected)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+    def test_get_remaining_submission_time_when_limit_is_exhausted(self):
+        self.url = reverse_lazy('jobs:get_remaining_submissions',
+                                kwargs={'challenge_phase_pk': self.challenge_phase.pk,
+                                        'challenge_pk': self.challenge.pk})
+        setattr(self.challenge_phase, 'max_submissions_per_day', '1')
```
taranjeet: I think 1 should be an integer.

PR #978, opened 2017-04-30
Remove unnecessary print statements from views and tests

TroJan:
@deshraj


deshraj:
My bad. I will take care of this from next time. Thanks for fixing. :)

PR #979, opened 2017-04-30
submission_worker.py: Add logging statements instead of print statements

TroJan:
@deshraj


deshraj:
Thanks, this looks good to me. :)  Although I think we would have to change the logging configuration in the settings really soon.

TroJan:
@deshraj : yeah but as of now this is actually better than print statements atleast. :)

deshraj:
Yeah, agree.

PR #981, opened 2017-04-30
Submission Worker: Proper error handling in case

TroJan:
@deshraj : this message prevents worker from blocking the queue(not acking the message) when a message of Submission which does not exist is published.


deshraj:
LGTM. Merging this. :)

Issue #1000, opened 2017-05-13
Replace %s with string format {} across all str functions of models


TroJan:
This is a pretty simple task. Some part of the code base contains `%s` function(used for string formatting). It involves replacing all the occurrences of such `%s` with string format function `{}` everywhere in models of each app.

For example one such case can be found in `challenges/models.py`, __LeaderBoard__ model.
The function right now looks like 
```
def __unicode__(self):
        return "%s" % (self.id)
```

After replacing `%s` with string format function, the above function will looks something like

```
def __unicode__(self):
    return '{}'.format(self.id)
```

Note: Only replace in models of each app, not across the whole project as of now.

isht3:
I would be happy to help! Is anyone working on this?

deshraj:
No, I don't think so. Please go ahead. :)

isht3:
Okay, on it. Why are we doing it? Is the latter method more efficient than the other?

deshraj:
For more reference, please see this http://stackoverflow.com/questions/5082452/python-string-formatting-vs-format

isht3:
Understood. Thanks. :+1:

PR #1001, opened 2017-05-13
Fix: Replace %s with string format {} across all str functions of app models

isht3:
reference to issue #1000


TroJan:
@isht3 : thanks for the quick fix. :)

isht3:
@TroJan Anytime. :smile:

Issue #1002, opened 2017-05-13
Add test cases for model string functions where ever missing.


TroJan:
Each model has a `__unicode__` function. This function is present in many of the models. 

Presently test cases are missing for some model's `unicode` functions. This task addresses the same issue.

It first involves figuring out a list of all the models where its `__unicode__` function is present and its corresponding test case is not written. Then multiple PR can be raised per app which adds the tests for the same.

One such example of `unicode` function test is in `tests/unit/challenges/test_models.py`,  with the snippet

```
def test__str__(self):
        self.assertEqual(self.challenge.title, self.challenge.__str__())
```

TroJan:
@isht3 :  Are you willing to work on this ? This will give you a fair idea of how the models are organised and placed in apps.

isht3:
@TroJan I would love to!

TroJan:
@isht3 : Sure, go ahead. Let me know if you need any help

isht3:
@TroJan A PR for each app would suffice, right?

isht3:
If we haven't declared a `__str__()` method in the class, then how does `self.challenge.__str__()` work?

TroJan:
@isht3 : yeah a PR for each app would do.

So there are three functions namely `str`, `unicode` and `repr`. First str is looked(infact str is the first choice for python 3), then `unicode`(in python 2 this is the first choice) and then `repr`.

So in our case when `__str__` is called, python fallbacks to pickup `unicode`.

I hope this clears your doubt

isht3:
Thanks, now its cleared! :+1:

TroJan:
@isht3 : Can you list down all the apps where these changes are required. challenges app is done.

isht3:
@TroJan In `Challenges`, `Leaderboard` model is to be completed, I found that a little hard because a lot of things like `Submission` and other classes had to be set up to test that data, and I'm working on it now.

On `Accounts`, there's this one small issue which is bothering me, while setting up `ProfileTestCase` and running the tests, this piece of code, 
```
    def setUp(self):
        super(ProfileTestCase, self).setUp()
        self.profile = Profile.objects.create(
                user=self.user,
                contact_number="2983191839",
                affiliation="testing"
            )
```

throws the following error.
```
IntegrityError: duplicate key value violates unique constraint "user_profile_user_id_key"
DETAIL:  Key (user_id)=(3) already exists.

```
[...]

PR #1007, opened 2017-05-15
Add tests for __unicode__ function in models

isht3:
tests for __unicode__ have been added for the following classes:
1. challenges.DatasetSplit
2. challenges.Leaderboard
3. challenges.ChallengePhaseSplit

In response to #1002, the last function, `LeaderboardData` needs `Submission` and other classes to be present. So ill do a PR for that class soon enough along with the rest of them.

### DIFF
--- a/tests/unit/challenges/test_models.py
+++ b/tests/unit/challenges/test_models.py
@@ -1,3 +1,4 @@
+import json
 import os
 import shutil
 
@@ -8,7 +9,7 @@ from django.test import TestCase
 from django.contrib.auth.models import User
 from django.utils import timezone
 
-from challenges.models import Challenge, ChallengePhase
+from challenges.models import Challenge, ChallengePhase, ChallengePhaseSplit, DatasetSplit, Leaderboard
 from hosts.models import ChallengeHostTeam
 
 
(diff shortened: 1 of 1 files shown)


TroJan:
@isht3 : Thanks for the fix. Will do review by tonight. :)

isht3:
@TroJan Do let me know if any changes should be made, and Ill start work on the rest after that. :smile:

tests/unit/challenges/test_models.py
```diff
@@ -8,7 +9,7 @@
 from django.contrib.auth.models import User
 from django.utils import timezone
 
-from challenges.models import Challenge, ChallengePhase
+from challenges.models import Challenge, DatasetSplit, ChallengePhase, Leaderboard, ChallengePhaseSplit
```
TroJan: Can we please order these imports alphabetically ?

tests/unit/challenges/test_models.py
```diff
@@ -96,6 +97,29 @@ def test_get_end_date(self):
                          self.challenge.get_end_date())
 
 
+class DatasetSplitTestCase(BaseTestCase):
+
+    def setUp(self):
+        super(DatasetSplitTestCase, self).setUp()
```
TroJan: The idea behind making a temporary directory is to upload all files related to a model here. After everything is done this directory can simply be removed. Also mostly os manages the temp directory allocation and removal by themselves.

But if we see DatasetSplit model, there is no file field.  So I think you can remove this part from here. Same goes for LeaderBoard model.

tests/unit/challenges/test_models.py
```diff
@@ -142,3 +166,68 @@ def test_get_start_date(self):
...
+            visibility=ChallengePhaseSplit.PUBLIC
+            )
+
+    def tearDown(self):
+        shutil.rmtree('/tmp/evalai')
+
+    def test__str__(self):
+        challenge_phase_name = self.challenge_phase_split.challenge_phase.name
+        dataset_split_name = self.challenge_phase_split.dataset_split.name
+        string = challenge_phase_name + " : " + dataset_split_name
```
TroJan: Here instead of joining the string, we can use format function like ```'{}:{}'.format(challenge_phase_name, dataset_split_name)```. Also can we please make the name more meaningful(string is kind of generic)

isht3:
@TroJan Done. :smiley:

TroJan:
@isht3 : thanks for the fix :)

Issue #1012, opened 2017-05-20
Group all auth related files into single folder named as Auth


aka-jain:
Pages that are part of Auth view should be grouped into a single folder under 'web' folder and named as 'auth'.

aka-jain:
Fixed with #1014

PR #1014, opened 2017-05-20
Grouped auth files. 

aka-jain:
From now onwards, Auth files will have found in 'auth' named folder. 
Fixing #1012


deshraj:
Looks good. Merging this.

PR #1015, opened 2017-05-21
Added new dashboard design

aka-jain:
@deshraj @spyshiv please review.


frontend/src/css/core/grad-base.scss
```diff
@@ -25,6 +25,19 @@
...
 }
 
+.grad-btn-transparent {
+    background-color: rgba(0,0,0,0);
+    font-weight: $rob-reg;
+    color: $med-gray;
+    border: 1px solid $med-gray;
+    box-shadow: 0px 4px 8px transparent;
+    &:hover {
+        box-shadow: 0px 0px 8px #9d9d9d;
```
aayusharora: @aka-jain The color codes should be replaced by .scss `variables.scss` .

frontend/src/css/modules/sidebar.scss
```diff
@@ -1,12 +1,13 @@
 .ev-sidebar {
     display: block;
-    height: 100vh;
+    height: calc(100vh - 120px);
```
aayusharora: This seems dubious. Height is calculated with `calc using vh`.
- Subtracting a constant value say `120px` will make it `non-responsive` 
Better to use percentages here. :)

frontend/src/css/modules/sidebar.scss
```diff
@@ -1,12 +1,13 @@
...
     width: 220px;
-    
-    background: $dark-gray;
+    border-right: 1px solid #dcdcdc;
+    background-color: rgba(0, 0, 0, 0);
```
aayusharora: - Replace rgba(0,0,0,0) by variable.

frontend/src/css/modules/sidebar.scss
```diff
@@ -28,32 +29,36 @@
     }
     li {
         a {
```
aayusharora: ## Nesting Required 
 Nesting the properties like
a {
            &:hover {}
 } 
will be better here :)

frontend/src/css/modules/sidebar.scss
```diff
@@ -28,32 +29,36 @@
...
             height: 100%;
             font-size: 14px;
+            transition: all 0.2s ease-in-out;
+            border-right: 1px solid transparent;
         }
         a:hover {
-            color: #fff;
+            color: $highlight;
+        }
+        a.side-active-link {
```
aayusharora: Nesting here too

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -167,6 +167,7 @@
                                                 parameters.callback = {
                                                     onSuccess: function() {
                                                         vm.isParticipated = true;
+
```
aayusharora: Can you please remove this blank line.

frontend/src/views/web/partials/sidebar.html
```diff
@@ -1,6 +1,6 @@
 <!-- define all main links here for navigation -->
-<div class="ev-sidebar ev-fixed z-depth-3 fixed" id="sidebarToggle">
-    <div class="side-intro ev-md-container fixed">
+<div class="ev-sidebar ev-fixed  fixed hide-on-med-and-down">
+    <!--    <div class="side-intro ev-md-container fixed">
```
aayusharora: Remove the extra commented code :)

frontend/src/views/web/partials/sidebar.html
```diff
@@ -47,53 +47,37 @@
...
-                </li>
-            </ul>
+        <div class="links-section">
             <ul>
-                <li class="text-highlight side-title w-500 ">My Details</li>
                 <li>
-                    <a class="waves-effect waves-dark text-light-gray ev-md-container" ui-sref="web.profile" ui-sref-active="side-active-link">
-                        <i class="fa fa-user side-link-icon"></i> My Profile</a>
+                    <a class=" text-light-black" ui-sref="web.dashboard" ui-sref-active="side-active-link">
+                        <div>
```
aayusharora: I think we can remove one extra `div` here 
No need to write `<i></i>` inside a `div`

frontend/src/views/web/profile.html
```diff
@@ -1,24 +1,29 @@
...
-                <span class="text-light-gray fs-12">Last Name<br></span> {{profile.user.last_name}}
-            </li>
-            <li>
-                <span class="text-light-gray fs-12">Affiliated To<br></span> -
-            </li>
-        </ul>
-<!--         <li>
+<!-- <slide-profile></slide-profile> -->
+<div class="ev-md-container center">
+    <img src="dist/images/pro-pic.png" height="80" width="auto">
```
aayusharora: ## Proposed approach for Image size
- Make a mixing in `main.scss` name as `ImageSize` with two parameters `width` and `height`
- Give all the images a class.
- Call the mixin in the class with the arguments stating `width` and `height`

aayusharora: Doing this by HTML will increase DOM overhead.

frontend/src/css/modules/sidebar.scss
aka-jain: No we can't use percentage here, as sidebar should of fix height and there is no need to make it responsive height wise. So sometimes it's good to use fix value.

frontend/src/css/modules/sidebar.scss
aka-jain: Actually this rgba is used rarely in codebase that's why I used it directly.

frontend/src/css/modules/sidebar.scss
aka-jain: Yeap, I know. Will do it in CSS refactoring.

frontend/src/views/web/profile.html
aka-jain: Oh, it has done by mistake. Will make it correct. Thanks.

frontend/src/css/core/grad-base.scss
aka-jain: Forgot to use variable here. Thanks for pointing this out.

frontend/src/views/web/partials/sidebar.html
aka-jain: We might need this later. So let's have this commented code for sometime, will change it later.

frontend/src/views/web/partials/sidebar.html
aka-jain: It's just to make i tag a block element. Otherwise we have to give a class to i to make it a block element.

aka-jain:
Thanks @aayusharora for the review. Will do the required changes as needed.

frontend/src/css/modules/sidebar.scss
aayusharora: Hey @aka-jain The dubious situation is the subtraction of 120px from 100% of viewport height.
Why we need to subtract this. The height of sidebar should always be equal to the viewport height.
Taking in the case of mobiles where this 120px ( CSS ) will scale up by the `<meta content='width-device-width >' value and then get subtracted from the viewport of Mobiles.
IMHO, 120px creates a dubious situation hence must be removed.

frontend/src/css/modules/sidebar.scss
aayusharora: Ok. But, IMHO Getting all color variables from a file will make it easy for the new developers to use them properly.

frontend/src/views/web/partials/sidebar.html
aayusharora: Ok. Cool

frontend/src/views/web/partials/sidebar.html
aayusharora: ok, It is not suitable to make a tag a `block` element with HTML. I guess better to do it with css?

aka-jain: Yes, it will be better to use css instead of semantic tags but I did that intentionally because I don't want to change the default icons styles here. 
But let me think again, will it be good to use div here or not.

frontend/src/css/modules/sidebar.scss
aka-jain: In new design it's not. Sidebar is not starting from topmost point. It has some top space which is equal to 120px here also this happens due to it's fixed position nature. Also in case of mobile we will have different sidebar which has full viewport height, so there is no need to remove 120px here. Getting?

frontend/src/css/modules/sidebar.scss
aka-jain: Agree, will do it.

frontend/src/css/modules/sidebar.scss
aayusharora: Not sure about this, if you need to provide an element 120px from topmost point.
( ` has some top space which is equal to 120px`) IMHO the property `top` must be used instead of decreasing its viewport height.

frontend/src/views/web/partials/sidebar.html
aayusharora: Let's think this towards the reusability.
What if you have 20 `<i>` tags,
What will be the right approach 
- Adding 20 divs Or adding display block to `i class`

deshraj:
@aka-jain: can you please share a screenshot for the same?

frontend/src/css/modules/sidebar.scss
aka-jain: Have you run the PR?

aka-jain:
screenshotatuploadcc_1495832314732

Have a look @deshraj

frontend/src/css/modules/sidebar.scss
aka-jain: I already gave the top popsition to sidebar, but to make a required design I have to give that calc height. Please check the design first I think you are missing something.

frontend/src/views/web/partials/sidebar.html
aka-jain: Yes, I got your point @aayusharora regarding reusability. As I said above, I did that intentionally and I'm also thinking to give some styles to icon container. Just let me finalize the design otherwise I'll remove it.

aka-jain:
Changes done. @deshraj Please review and merge.

frontend/src/css/modules/sidebar.scss
aayusharora: Ok ! np, I can't test this without merging rights, Is it hosted somewhere?

aka-jain: Ohk, no it's not hosted. 
No problem, let's merge this then you can check it. It's correct from my side and according to design we have to do this.

deshraj: We do have a staging server where we can test things.

PR #1017, opened 2017-05-24
Fix: Add tests for __unicode__ functions in participants model.

isht3:
tests for unicode have been added for the following classes:

        1. Participant
        2. ParticipantTeam

#1002

### DIFF
--- a/tests/unit/participants/test_models.py
+++ b/tests/unit/participants/test_models.py
@@ -1,0 +1,42 @@
+from django.test import TestCase
+from django.contrib.auth.models import User
+
+from participants.models import Participant, ParticipantTeam
+
+
+class BaseTestCase(TestCase):
+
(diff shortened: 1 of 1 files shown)


tests/unit/participants/test_models.py
```diff
@@ -0,0 +1,42 @@
+from django.test import TestCase
+from django.contrib.auth.models import User
+
+from participants.models import Participant, ParticipantTeam
+
+
+class BaseTestCase(TestCase):
```
TroJan: Can you please be consistent with your use of single and double quotes everywhere. We however prefer single quotes.

TroJan:
@isht3 : thanks for the help :)

isht3:
@TroJan Anytime!

PR #1018, opened 2017-05-25
Fix: Add tests for __unicode__ functions in web model.

isht3:
tests for unicode have been added for the following classes:

    1. Contact
    2. Team

Should I leave out BaseTestCase or preserve for future additions?

#1002

PR #1020, opened 2017-05-26
Fix #1018: Change remote from upstream to origin in pull request review script

deshraj:


TroJan:
@deshraj : As far as I know this was correct in its place. A maintainer wants to fetch a pull request which another contributor has pushed at his origin. I think you are confused with this origin.

If you write origin in the script, then it means the origin of the maintainer(where the PR does not exists). It exists at the origin of the contributor and the upstream(because he raises a diff from it)

So, this change is not required.

deshraj:
I have been under the impression that this feature is for the maintainers to review the PRs on CloudCV EvalAI's Origin and not the users' origin since the maintainer would want to review the PR submitted to the CloudCV's origin and not the fork's origin. Does that sound good?

TroJan:
I have origin as my fork, upstream as evalai. Hence the comment.

PR #1025, opened 2017-05-29
Fix: Add tests for __unicode__ functions in jobs model.

isht3:
tests for unicode have been added for the following classes:
        1. Submission
#1002


tests/unit/jobs/test_models.py
```diff
@@ -0,0 +1,83 @@
+import os
+import shutil
+
+from datetime import timedelta
+
+from django.test import TestCase
```
TroJan: Can you order the import lexicographically

tests/unit/jobs/test_models.py
```diff
@@ -0,0 +1,83 @@
...
+
+        self.submission = Submission.objects.create(
+            participant_team=self.participant_team,
+            challenge_phase=self.challenge_phase,
+            created_by=self.challenge_host_team.created_by,
+            status='submitted',
+            input_file=self.challenge_phase.test_annotation,
+            method_name='Test Method',
+            method_description='Test Description',
+            project_url='http://testserver/',
```
TroJan: If these fields are not mandatory, we can actually skip this. like project_url and publication_url

TroJan:
@isht3 : thanks :)

PR #1026, opened 2017-06-01
Fix: Add tests for __unicode__ functions in accounts model.

isht3:
Tests for unicode have been added for the following classes:

        1. Profile
        2. UserStatus

#1002


TroJan:
@isht3 : thanks for the fix :)

PR #1027, opened 2017-06-01
Fix: Add tests for __unicode__ functions in Challenge.LeaderboardData model.

isht3:
Last of the PR for issue #1002

### DIFF
--- a/tests/unit/challenges/test_models.py
+++ b/tests/unit/challenges/test_models.py
@@ -9,8 +9,10 @@ from django.test import TestCase
 from django.contrib.auth.models import User
 from django.utils import timezone
 
-from challenges.models import Challenge, ChallengePhase, ChallengePhaseSplit, DatasetSplit, Leaderboard
+from challenges.models import Challenge, ChallengePhase, ChallengePhaseSplit, DatasetSplit, Leaderboard, LeaderboardData
 from hosts.models import ChallengeHostTeam
+from jobs.models import Submission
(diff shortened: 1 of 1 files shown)


tests/unit/challenges/test_models.py
```diff
@@ -115,25 +148,6 @@ class ChallengePhaseTestCase(BaseTestCase):
 
     def setUp(self):
         super(ChallengePhaseTestCase, self).setUp()
-        try:
```
TroJan: @isht3 : I am not able to understand why is this piece of code touched in this me PR. Can you please explain me ?

tests/unit/challenges/test_models.py
```diff
@@ -175,34 +189,6 @@ class ChallengePhaseSplitTestCase(BaseTestCase):
 
     def setUp(self):
         super(ChallengePhaseSplitTestCase, self).setUp()
-        try:
```
TroJan: Same goes for here. Can you explain why is this removed ?

tests/unit/challenges/test_models.py
isht3: @TroJan, Since that Class ChallengePhase Class was to be used to be in 3 TestClass', I thought it would be good to move it into BaseClass because it would remove redundancy and reduce number of lines of code.

TroJan:
@isht3 : thanks for the fix. :)

PR #1028, opened 2017-06-02
Base app : Add a object checking and returning function.

RishabhJain2018:

### DIFF
--- a/apps/base/utils.py
+++ b/apps/base/utils.py
@@ -4,6 +4,7 @@ import uuid
 from django.conf import settings
 from django.utils.deconstruct import deconstructible
 
+from rest_framework.exceptions import NotFound
 from rest_framework.pagination import PageNumberPagination
 
 
@@ -32,3 +33,14 @@ class RandomFileName(object):
         filename = '{}{}'.format(uuid.uuid4(), extension)
         filename = os.path.join(self.path, filename)
         return filename
+
+
+def get_model_object(model_name):
+    def get_model_by_pk(pk):
+        try:
+            model_object = model_name.objects.get(pk=pk)
+            return model_object
+        except model_name.DoesNotExist:
+            raise NotFound('{} {} does not exist'.format(model_name.__name__, pk))
+    get_model_by_pk.__name__ = "get_{}_object".format(model_name)
+    return get_model_by_pk
--- a/apps/base/utils.py
+++ b/apps/base/utils.py
@@ -42,5 +42,5 @@ def get_model_object(model_name):
             return model_object
         except model_name.DoesNotExist:
             raise NotFound('{} {} does not exist'.format(model_name.__name__, pk))
-    get_model_by_pk.__name__ = "get_{}_object".format(model_name.__name__.lower())
+    get_model_by_pk.__name__ = 'get_{}_object'.format(model_name.__name__.lower())
     return get_model_by_pk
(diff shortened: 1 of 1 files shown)


apps/base/utils.py
```diff
@@ -32,3 +33,11 @@ def __call__(self, instance, filename):
...
         filename = os.path.join(self.path, filename)
         return filename
+
+
+def get_model_object(model_name, pk):
+    try:
+        model_object = model_name.objects.get(pk=pk)
+        return model_object
+    except model_name.DoesNotExist:
+        raise NotFound("{} does not exist".format(model_name.__name__))
```
TroJan: Can we have id also included in the log statement. Also can you please use single quotes here?

RishabhJain2018: yeah sure.

tests/unit/base/test_utils.py
```diff
@@ -0,0 +1,62 @@
+from django.contrib.auth.models import User
```
TroJan: this is actually not required. I know we have a convention where in for any code being merged into master, we should have test case for that. But here the case is a bit different. This code will be called by other functions, so it will be covered when test cases are being written for them. So I think we can skip test cases here.

apps/base/utils.py
```diff
@@ -32,3 +33,11 @@ def __call__(self, instance, filename):
         filename = '{}{}'.format(uuid.uuid4(), extension)
         filename = os.path.join(self.path, filename)
         return filename
+
+
+def get_model_object(model_name, pk):
```
TroJan: I think this is not what I meant. I think what I meant was a function which wraps this piece of code for a model and then returns that function

RishabhJain2018: @TroJan ohh..ok

RishabhJain2018:

apps/base/utils.py
```diff
@@ -32,3 +33,14 @@ def __call__(self, instance, filename):
...
+
+
+def get_model_object(model_name):
+    def get_model_by_pk(pk):
+        try:
+            model_object = model_name.objects.get(pk=pk)
+            return model_object
+        except model_name.DoesNotExist:
+            raise NotFound('{} {} does not exist'.format(model_name.__name__, pk))
+    get_model_by_pk.__name__ = "get_{}_object".format(model_name)
```
TroJan: One can pass model name like ```Challenge```, so when creating function, we should use ```model_name.lower()```. Also check if we should use `model_name`or `model_name.__name__`. I think name one should be used.

RishabhJain2018: @TroJan Ok sure.

RishabhJain2018:

apps/base/utils.py
```diff
@@ -32,3 +33,14 @@ def __call__(self, instance, filename):
...
+
+
+def get_model_object(model_name):
+    def get_model_by_pk(pk):
+        try:
+            model_object = model_name.objects.get(pk=pk)
+            return model_object
+        except model_name.DoesNotExist:
+            raise NotFound('{} {} does not exist'.format(model_name.__name__, pk))
+    get_model_by_pk.__name__ = "get_{}_object".format(model_name.__name__.lower())
```
taranjeet: Just a minor glitch over here. Can we please use single quotes instead of double quotes.

RishabhJain2018: yeah sure.

RishabhJain2018:

taranjeet:
@RishabhJain2018 : thanks :)

PR #1035, opened 2017-06-11
[REVIEW PENDING] Zip Challenge: Add new serializers for creating challenge.

### LINKED ISSUE #1036: Add serializers for creating challenge using zip file.

RishabhJain2018:
fixes #1036 @deshraj @taranjeet Please review the PR.

### DIFF
--- a/apps/challenges/serializers.py
+++ b/apps/challenges/serializers.py
@@ -35,10 +35,16 @@ class ChallengeConfigurationSerializer(serializers.ModelSerializer):
     """
     Serialize the Challenge Configuration Model.
     """
+    def __init__(self, *args, **kwargs):
+        super(ChallengeConfigurationSerializer, self).__init__(*args, **kwargs)
+        context = kwargs.get('context')
+        if context:
+            user = context.get('user')
+            kwargs['data']['user'] = user.pk
 
     class Meta:
         model = ChallengeConfiguration
-        fields = ('zip_configuration',)
+        fields = ('zip_configuration', 'user',)
 
 
 class ChallengePhaseSerializer(serializers.ModelSerializer):
@@ -92,7 +98,7 @@ class LeaderboardSerializer(serializers.ModelSerializer):
         fields = '__all__'
 
 
-class ZipFileCreateChallengeSerializer(ChallengeSerializer):
+class ZipConfigurationChallengeSerializer(ChallengeSerializer):
     """
     Serialize the Challenge Model and is used to create challenge using zip file uploaded by user.
     """
@@ -105,7 +111,7 @@ class ZipFileCreateChallengeSerializer(ChallengeSerializer):
 
 class ZipFileCreateChallengePhaseSplitSerializer(serializers.ModelSerializer):
     """
-    Serialize the ChallengePhaseSplits Model and is used to create challenge phase splits
+    Serialize the ChallengePhaseSplit Model and is used to create challenge phase splits
     using zip file uploaded by user.
     """
     class Meta:
--- a/apps/challenges/serializers.py
+++ b/apps/challenges/serializers.py
@@ -31,22 +31,6 @@ class ChallengeSerializer(serializers.ModelSerializer):
                   'published', 'enable_forum', 'anonymous_leaderboard', 'is_active',)
 
 
-class ChallengeConfigurationSerializer(serializers.ModelSerializer):
-    """
-    Serialize the Challenge Configuration Model.
-    """
-    def __init__(self, *args, **kwargs):
-        super(ChallengeConfigurationSerializer, self).__init__(*args, **kwargs)
-        context = kwargs.get('context')
-        if context:
-            user = context.get('user')
-            kwargs['data']['user'] = user.pk
-
-    class Meta:
-        model = ChallengeConfiguration
-        fields = ('zip_configuration', 'user',)
-
-
 class ChallengePhaseSerializer(serializers.ModelSerializer):
 
     is_active = serializers.ReadOnlyField()
@@ -63,6 +47,13 @@ class ChallengePhaseSerializer(serializers.ModelSerializer):
         fields = ('id', 'name', 'description', 'leaderboard_public', 'start_date',
                   'end_date', 'challenge', 'max_submissions_per_day', 'max_submissions',
                   'is_public', 'is_active', 'codename')
+
+
+class DatasetSplitSerializer(serializers.ModelSerializer):
+
+    class Meta:
+        model = DatasetSplit
+        fields = '__all__'
 
 
 class ChallengePhaseSplitSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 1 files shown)


apps/challenges/serializers.py
```diff
@@ -63,11 +67,47 @@ class ChallengePhaseSplitSerializer(serializers.ModelSerializer):
...
         fields = ('id', 'dataset_split', 'challenge_phase', 'challenge_phase_name', 'dataset_split_name', 'visibility')
 
     def get_dataset_split_name(self, obj):
         return obj.dataset_split.name
 
     def get_challenge_phase_name(self, obj):
         return obj.challenge_phase.name
+
+
+class DatasetSplitSerializer(serializers.ModelSerializer):
```
deshraj: I am guessing that you moved the `DatasetSplitSerializer` here so that all the serializers are in alphabetical order, right?

apps/challenges/serializers.py
```diff
@@ -63,11 +67,47 @@ class ChallengePhaseSplitSerializer(serializers.ModelSerializer):
...
+    class Meta:
+        model = Challenge
+        fields = ('id', 'title', 'short_description', 'description', 'terms_and_conditions',
+                  'submission_guidelines', 'start_date', 'end_date', 'creator',
+                  'published', 'enable_forum', 'anonymous_leaderboard', 'is_active',)
+
+
+class ZipFileCreateChallengePhaseSplitSerializer(serializers.ModelSerializer):
+    """
+    Serialize the ChallengePhaseSplits Model and is used to create challenge phase splits
```
deshraj: Replace `ChallengePhaseSplits` with `ChallengePhaseSplit`

apps/challenges/serializers.py
```diff
@@ -63,11 +67,47 @@ class ChallengePhaseSplitSerializer(serializers.ModelSerializer):
...
+class LeaderboardSerializer(serializers.ModelSerializer):
+    """
+    Serialize the Leaderboard Model.
+    """
+    class Meta:
+        model = Leaderboard
+        fields = '__all__'
+
+
+class ZipFileCreateChallengeSerializer(ChallengeSerializer):
```
deshraj: Replace `ZipFileCreateChallengeSerializer` with `ZipConfigurationChallengeSerializer`

apps/challenges/serializers.py
RishabhJain2018: yes.

deshraj:
Looks good. :)

apps/challenges/serializers.py
```diff
@@ -63,11 +73,47 @@ class ChallengePhaseSplitSerializer(serializers.ModelSerializer):
...
+    Serialize the Leaderboard Model.
+    """
+    class Meta:
+        model = Leaderboard
+        fields = '__all__'
+
+
+class ZipConfigurationChallengeSerializer(ChallengeSerializer):
+    """
+    Serialize the Challenge Model and is used to create challenge using zip file uploaded by user.
```
taranjeet: Can we rewrite the doc string as "Serializer used for creating challenge through zip file". Same goes for ZipFileCreateChallengePhaseSplitSerializer.

apps/challenges/serializers.py
```diff
@@ -30,6 +31,22 @@ class Meta:
                   'published', 'enable_forum', 'anonymous_leaderboard', 'is_active',)
 
 
+class ChallengeConfigurationSerializer(serializers.ModelSerializer):
+    """
+    Serialize the Challenge Configuration Model.
```
taranjeet: Writing doc string is a good practice. I really appreciate that. Just that it will be great if we can keep the model name as `ChallengeConfiguration` (no spaces).

apps/challenges/serializers.py
```diff
@@ -48,13 +65,6 @@ class Meta:
                   'is_public', 'is_active', 'codename')
 
 
-class DatasetSplitSerializer(serializers.ModelSerializer):
```
taranjeet: Also there is no need of reordering this serializer here. One of the main reason can be it will actually confuse someone of why this model was involved in this commit. Also in the master, its model `DatasetSplit` is not ordered alphabetically. So in future(may be) we will open an issue where in you can reorder this.

One thought for keeping the serializer as it is it actually tells the history of how the model came into existence. Dont know how much of this is true, but yeah one can simply infer that yeah `Challenge` model was introduced before `ChallengePhase` model.

RishabhJain2018: I agree that it can confuse other people in future. I will reorder it as  previously and yes definitely we can open an issue in future to reorder them.

apps/challenges/serializers.py
```diff
@@ -63,11 +64,55 @@ class ChallengePhaseSplitSerializer(serializers.ModelSerializer):
 
     class Meta:
         model = ChallengePhaseSplit
-        fields = '__all__'
```
taranjeet: I am actually not sure why is this line removed?

apps/challenges/serializers.py
```diff
@@ -63,11 +64,55 @@ class ChallengePhaseSplitSerializer(serializers.ModelSerializer):
...
         fields = ('id', 'dataset_split', 'challenge_phase', 'challenge_phase_name', 'dataset_split_name', 'visibility')
 
     def get_dataset_split_name(self, obj):
         return obj.dataset_split.name
 
     def get_challenge_phase_name(self, obj):
         return obj.challenge_phase.name
+
+
+class ChallengeConfigurationSerializer(serializers.ModelSerializer):
```
taranjeet: Can we have a some what short name like `ChallengeConfigSerializer` for it?

apps/challenges/serializers.py
```diff
@@ -63,11 +64,55 @@ class ChallengePhaseSplitSerializer(serializers.ModelSerializer):
...
+        if context:
+            user = context.get('user')
+            kwargs['data']['user'] = user.pk
+
+    class Meta:
+        model = ChallengeConfiguration
+        fields = ('zip_configuration', 'user',)
+
+
+class LeaderboardSerializer(serializers.ModelSerializer):
```
taranjeet: I think this PR is for adding serializers related to Challenge Config. Hence this should not be here. Can you please check ?

apps/challenges/serializers.py
```diff
@@ -63,11 +64,55 @@ class ChallengePhaseSplitSerializer(serializers.ModelSerializer):
...
+    """
+    Serialize the Leaderboard Model.
+    """
+    class Meta:
+        model = Leaderboard
+        fields = '__all__'
+
+
+class ZipConfigurationChallengeSerializer(ChallengeSerializer):
```
taranjeet: Can we rename it to `ZipChallengeSerializer` ?

apps/challenges/serializers.py
```diff
@@ -63,11 +64,55 @@ class ChallengePhaseSplitSerializer(serializers.ModelSerializer):
...
+    Serializer used for creating challenge through zip file.
+    """
+    class Meta:
+        model = Challenge
+        fields = ('id', 'title', 'short_description', 'description', 'terms_and_conditions',
+                  'submission_guidelines', 'start_date', 'end_date', 'creator',
+                  'published', 'enable_forum', 'anonymous_leaderboard', 'is_active',)
+
+
+class ZipFileCreateChallengePhaseSplitSerializer(serializers.ModelSerializer):
```
taranjeet: Can we rename it to `ZipChallengePhaseSplitSerializer` ?

apps/challenges/serializers.py
RishabhJain2018: Actually, this line is updated in the line just after it. So, there is no use of this line. That's why I removed it. Please check [here](https://github.com/Cloud-CV/EvalAI/blob/master/apps/challenges/serializers.py#L66).

apps/challenges/serializers.py
RishabhJain2018: Actually , this PR is adding all the serializers that are needed during the creation of challenge using zip file and as we are creating the Leaderboard model instance there , so I added the serializer for it. Should i create a separate PR for this?

taranjeet: Ohh, then let it be

Issue #957, opened 2017-04-23
Tests: Inconsistency in unittests for URL Patterns


deshraj:
### Issue

The tests for URL Patterns are inconsistent. For reference, see this https://github.com/Cloud-CV/EvalAI/blob/master/tests/unit/challenges/test_urls.py#L68

There is no `challenge_phase` being created in the test and challenge_pk is passed in the URL instead of `challenge_phase_pk`. 

### How to fix

1. Create a challenge_phase for the challenge created in the tests.
2. Pass the `challenge_phase_pk` instead of `challenge_pk` in the URLs wherever required.

RishabhJain2018:
@deshraj I also noticed that the urls are not tested in a manner that they connect to correct view. 
For Reference please see http://stackoverflow.com/questions/18987051/how-do-i-unit-test-django-urls 
I think we should also add this. What do you say ?

deshraj:
Yeah I agree. We have not covered the tests where we check if the particular URL pattern matches with the particular view. Please let me know if you want to take lead on this.

deshraj:
But I think it would be great if you can finish the currently assigned issues first before jumping on this one.

RishabhJain2018:
@deshraj Yeah, I would like to work on this but after completing all the assigned issues. :+1:

deshraj:
Sounds good.

isht3:
@RishabhJain2018 Are you planning on working on this? I'll be happy to take over if you have some other things to work on.

RishabhJain2018:
@isht3 Yeah sure, you can take this up.

isht3:
@RishabhJain2018 are we doing it the method you mentioned in the SOF answer?

RishabhJain2018:
@isht3  I think so , but I would like to ask @deshraj @taranjeet for their views over this.

taranjeet:
@isht3 : as @deshraj said, first we would like to complete this issue. We will open another issue where in we will cover the said thing in SO answer.

Can you take this up quickly and get it released in master?

isht3:
@taranjeet Sorry, I might've missed this notif. Ill get right on it. :+1:

PR #1040, opened 2017-06-13
Fix: Inconsistency in unittests for URL Patterns.

isht3:
#957


taranjeet:
@isht3 : thanks for the fix.

PR #1042, opened 2017-06-13
[REVIEW PENDING] Fix bug in RandomFileName class.

### LINKED ISSUE #1009: Change path of storage of Evaluation Scripts
## Current behaviour

Currently, the evaluation scripts are stored in the `submission_files` directory which should not happen ideally.

## Solution
Modify the `RandomFile()` method such that the evaluation scripts should be uploaded in the `evaluation_scripts` directory instead of the `submission_files` directory.

### LINKED ISSUE #1041: RandomFileName class returns incorrect path when id is not provided.
**Bug:** The `RandomFileName` class in [base.utils](https://github.com/Cloud-CV/EvalAI/blob/master/apps/base/utils.py#L22) returns incorrect path when `id` is not provided in the path.

**Fix:** The condition on the [line 28](https://github.com/Cloud-CV/EvalAI/blob/master/apps/base/utils.py#L28) needs to be changed to `or` from `and` as `instance.pk` will always be available for upon saving a model.

RishabhJain2018:
fixes #1041 and fixes #1009. @deshraj @taranjeet Please review the PR.


deshraj:
@RishabhJain2018 this really doesn't solve the problem. I committed that particular change some time back since this was not working properly for the challenge submissions and the submission number was misaligned with the entries being saved in the database after a successful execution of a submission. 

To replicate this, please follow the following steps:
1. Make some submissions to a challenge
2. Make sure that the submissions are executed successfully.
3. Go to django admin, and open the `submissions` link.
4. If you have the change that you just made, this will result in misalignment in the submission_number in the path of storing the submissions and it's stdout and sterr files.

RishabhJain2018:
@deshraj @taranjeet  Please review the PR. I think now the problem will be solved.

RishabhJain2018:

apps/base/utils.py
```diff
@@ -1,8 +1,8 @@
 import os
+import re
```
taranjeet: I think this import statement is not required now. Can we please remove this?

apps/base/utils.py
```diff
@@ -18,20 +18,18 @@ def paginated_queryset(queryset, request):
...
 
     def __call__(self, instance, filename):
         extension = os.path.splitext(filename)[1]
         if 'id' in self.path and instance.pk:
-            self.path = self.path.format(id=instance.pk)
-        elif 'id' not in self.path and instance.pk:
-            path = "submission_files/submission_{id}"
-            self.path = path.format(id=instance.pk)
+            path = self.path.format(id=instance.pk)
+        else:
```
taranjeet: What if we define path as a variable above the if condition and then when the if condition is true, we can simply change it. This way it will be more better. Let me know what are your thoughts over this.

RishabhJain2018: @taranjeet Yeah I agree.

RishabhJain2018:

apps/base/utils.py
```diff
@@ -18,20 +17,17 @@ def paginated_queryset(queryset, request):
     return (paginator, result_page)
 
 
-@deconstructible
```
taranjeet: I am actually not getting of the point of removing this `deconstrutible`. Can you please explain?

RishabhJain2018: @taranjeet I actually misunderstood the use of `deconstructible`. Basically, it is used to add a `deconstruct()` method to a class . The `deconstruct()` method is used to help perform model migrations that aren't handled by the system. Like Suppose if we have a `UUID Field` for a model instance, then django can't serialize that field and we will need a `deconstruct()` method. 
Please refer https://stackoverflow.com/a/35928526 and https://docs.djangoproject.com/en/1.10/topics/migrations/#adding-a-deconstruct-method these links.
So, I think we can remove `@deconstructible` from here. What do you say?

taranjeet: What will be the repercussions of this? Are you sure this wont impact the existing db or the app in any way?

RishabhJain2018: Removing `@deconstructible` will not cause any impact.
But if we merge this PR into the production, then it will change the paths for the upcoming submissions that will be stored in the database, as currently all the submissions are being stored in only one folder.
I think we should deploy this after the current `VQA challenge` is over. What are your views about it?

taranjeet:
@RishabhJain2018 : this PR is on hold from my side. I actually need to see the impact of this throughout the whole project.

RishabhJain2018:
@taranjeet  yeah sure. I think we can merge this after the current `VQA Challenge` will be over.

taranjeet:
@deshraj : your input here please.

apps/base/utils.py
taranjeet: @deshraj : can i have your suggestion here?

deshraj:
@taranjeet I think this is quite important and we should get this ready really soon. @RishabhJain2018  can you please add test cases for this for different fields like `evaluation_script`, `submissions` etc.

RishabhJain2018:
@deshraj Yeah sure.

RishabhJain2018:

PR #1043, opened 2017-06-13
Js unit test

aka-jain:
Done with JS unit test installations. Runners and frameworks used:

- Mocha

- Chai

- Karma

@spyshiv, @deshraj  I've done some changes in your branch, please have a look and let's close this down.


frontend/src/js/services/services.js
```diff
@@ -4,7 +4,7 @@
 (function() {
 
     'use strict';
-
+
```
deshraj: Can we get rid of this?

frontend/test/controllers-test/authCtrl.test.js
```diff
@@ -0,0 +1,6 @@
+'use strict';
```
deshraj: Can we rename the directory from `frontend/test` to `frontend/tests`

aka-jain: Sure @deshraj

aka-jain:
@deshraj please review.

karma.conf.js
```diff
@@ -0,0 +1,70 @@
+// Karma configuration
```
deshraj: @aka-jain should we keep the karma config file here or in `frontend/` directory?

aka-jain: No, actually we have our gulpfile, package.json all are in root directory that's why I putted karma config under root directory.

gulpfile.js
```diff
@@ -356,6 +357,25 @@ gulp.task('watch', function() {
...
+  new Server({
+    configFile: __dirname + '/karma.conf.js',
+    singleRun: true
+  }, done).start();
+});
+
+/**
+ * Watch for file changes and re-run tests on each change
+ */
+gulp.task('test:watch', function (done) {
```
aka-jain: Actually we have two types of test switches, one runs only once and another starts on changes like watch so just to make it distinguish as watch it will be good to take it as test:watch. And I don't think it's confusing it's watch for test simple. Isn't it?

karma.conf.js
```diff
@@ -0,0 +1,70 @@
...
+// Generated on Tue Mar 14 2017 06:19:19 GMT+0530 (IST)
+
+module.exports = function(config) {
+    config.set({
+
+        // base path that will be used to resolve all patterns (eg. files, exclude)
+        basePath: '',
```
aka-jain: Can you help here, what should be the base path here for files/directory?

aayusharora: We can set this to parent folder from which all other files will take the `path`.
Here, it can be taken as `../..` also specified in karma.config docs.
http://karma-runner.github.io/1.0/config/configuration-file.html

gulpfile.js
aayusharora: No worries. 👍

karma.conf.js
aayusharora: Hi @aka-jain @deshraj  This seems wrong. It should not be kept in the root, the gulpfile as well as the Karma.config must be inside the frontend folder.
The root is specific for the files that are common to `frontend` and `backend`.

aka-jain: Earlier we have our gulpfile and other frontend dependencies in frontend folder but after we changed it to the root directory because we want to run all the commands from the root and in our case as we are using Frontend as a separate directory we don't want to make contributors to go inside to run the Frontend tasks. All should run from the root directory. So here it's not a problem and is reasonable as we want all the configuration files in root directory. I hope you got our point why we are using them in root dir.

aka-jain:
@spyshiv please review and merge this PR.

karma.conf.js
aka-jain: @aayusharora I'm confused here, will check it later. Thanks for the suggestion.

aka-jain:
@deshraj  @spyshiv  any update you want with this PR?

spyshiv:
@aka-jain : LGTM. Let's merge it.

PR #1045, opened 2017-06-15
[REVIEW PENDING] Add an API endpoint to show all the submissions to Challenge Host for a particular Challenge.

### LINKED ISSUE #850: Create an API Endpoint to show all the submissions to Challenge Host for a particular Challenge
### Objective

The aim is to create an API endpoint which fetches all the submissions of a Challenge and and returns as a JSON Response. 

### Constraints:
- URL: `/api/challenges/<challenge_pk>/submissions/`
- Only the Challenge Host of that particular challenge can access this information. 
- Request Method: `GET`
- Response on success: `200`
- Response on not allowed: `406`
- Pagination: `YES`

### Stretch Goal
After we have the API ready, we would like to add support for filtering the submissions according to the `date of submission`, `challenge phase`, `status of submissions`.

RishabhJain2018:
fixes #850  @deshraj @taranjeet Please review the PR.

### DIFF
--- a/apps/challenges/views.py
+++ b/apps/challenges/views.py
@@ -361,9 +361,7 @@ def get_all_submissions_of_challenge(request, challenge_pk):
         return Response(response_data, status=status.HTTP_404_NOT_FOUND)
 
     try:
-        user = User.objects.get(username=request.user)
-        challenge_host = ChallengeHost.objects.get(user=user)
-        challenge_host_team = ChallengeHostTeam.objects.get(challengehost=challenge_host)
+        challenge_host_team = ChallengeHostTeam.objects.get(created_by=request.user)
     except ChallengeHostTeam.DoesNotExist:
         response_data = {'error': 'You are not a challenge host'}
         return Response(response_data, status=status.HTTP_400_BAD_REQUEST)
@@ -371,13 +369,13 @@ def get_all_submissions_of_challenge(request, challenge_pk):
     try:
         challenge_created_by_host = Challenge.objects.get(pk=challenge_pk, creator=challenge_host_team)
     except Challenge.DoesNotExist:
-        response_data = {'error': 'Challenge {} is not created by {}'.format(challenge_pk, challenge_host)}
+        response_data = {'error': 'Challenge {} is not created by {}'.format(challenge_pk, challenge_host_team)}
         return Response(response_data, status=status.HTTP_404_NOT_FOUND)
 
     submission = Submission.objects.filter(challenge_phase__challenge=challenge).order_by('-submitted_at')
     paginator, result_page = paginated_queryset(submission, request)
     try:
-        serializer = SubmissionSerializer(result_page, many=True, context={'request': request})
+        serializer = SubmissionSerializer(result_page, many=True)
         response_data = serializer.data
         return paginator.get_paginated_response(response_data)
     except:
--- a/apps/challenges/views.py
+++ b/apps/challenges/views.py
@@ -12,7 +12,7 @@ from rest_framework.throttling import UserRateThrottle, AnonRateThrottle
 from accounts.permissions import HasVerifiedEmail
 from base.utils import paginated_queryset
 from hosts.models import ChallengeHost, ChallengeHostTeam
-from hosts.utils import get_challenge_host_teams_for_user
+from hosts.utils import get_challenge_host_teams_for_user, is_user_a_host_of_challenge
 from jobs.models import Submission
 from jobs.serializers import SubmissionSerializer
 from participants.models import Participant, ParticipantTeam
@@ -351,7 +351,7 @@ def challenge_phase_split_list(request, challenge_pk):
 @authentication_classes((ExpiringTokenAuthentication,))
 def get_all_submissions_of_challenge(request, challenge_pk, challenge_phase_pk):
     """
-    Returns all the submissions for a particular Challenge.
+    Returns all the submissions for a particular challenge
     """
     # check if challenge exists or not.
     try:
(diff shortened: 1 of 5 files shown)


apps/challenges/views.py
```diff
@@ -340,3 +343,56 @@ def challenge_phase_split_list(request, challenge_pk):
...
+@authentication_classes((ExpiringTokenAuthentication,))
+def get_all_submissions_of_challenge(request, challenge_pk, challenge_phase_pk):
+    """
+    Returns all the submissions for a particular challenge
+    """
+    # check if challenge exists or not.
+    challenge_model = get_model_object(Challenge)
+    challenge = challenge_model(challenge_pk)
+
+    # check if challenge phase exists or not.
```
taranjeet: Instead of `check if challenge phase exists or not` what if we write the comment as `passed challenge as argument to make sure that this challenge phase belongs to the concerned challenge only`. This comment can be re-phrased in more better words. My idea here is that the comment should be less about what the code is, it should be more about why we need this or how are we using it further. Writing comments explaining what is not bad, but I feel that the code should be self explanatory.

apps/challenges/views.py
```diff
@@ -340,3 +343,56 @@ def challenge_phase_split_list(request, challenge_pk):
...
+    try:
+        challenge_phase = ChallengePhase.objects.get(pk=challenge_phase_pk, challenge=challenge)
+    except ChallengePhase.DoesNotExist:
+        response_data = {'error': 'Challenge Phase {} does not exist'.format(challenge_phase_pk)}
+        return Response(response_data, status=status.HTTP_404_NOT_FOUND)
+
+    # check if challenge_host_team exists or not.
+    if is_user_a_host_of_challenge(user=request.user, challenge_pk=challenge_pk):
```
taranjeet: The code is very well written here. But this can be further shortened as all we are doing is performing some mandatory checks and then paginating over a queryset and returning response. See if you can take serialization and response pagination part outside of the if condition. Let me know if this is not possible.

apps/hosts/utils.py
```diff
@@ -1,6 +1,13 @@
...
+from challenges.models import Challenge
 
 
 def get_challenge_host_teams_for_user(user):
     """Returns challenge host team ids for a particular user"""
     return ChallengeHost.objects.filter(user=user).values_list('team_name', flat=True)
+
+
+def is_user_a_host_of_challenge(user, challenge_pk):
+    """Returns boolean if the Participant Team participated in particular Challenge"""
```
taranjeet: I think it will be more appropriate if we use `Participant` instead of `Participant Team`. ?

apps/challenges/views.py
RishabhJain2018: @taranjeet Since we are using different serializers(`ChallengeSubmissionManagementSerializer` and `SubmissionSerializer`) based on the type of the user we are getting, I don't think there is a way to take that part of code out of the `if condition`. What do you say?

apps/challenges/views.py
```diff
@@ -340,3 +343,56 @@ def challenge_phase_split_list(request, challenge_pk):
...
+@throttle_classes([UserRateThrottle])
+@api_view(['GET'])
+@permission_classes((permissions.IsAuthenticated, HasVerifiedEmail))
+@authentication_classes((ExpiringTokenAuthentication,))
+def get_all_submissions_of_challenge(request, challenge_pk, challenge_phase_pk):
+    """
+    Returns all the submissions for a particular challenge
+    """
+    # To check for the corresponding challenge from challenge_pk.
+    challenge_model = get_model_object(Challenge)
```
taranjeet: Can we please use challenge getter function from challenge.utils instead of using it here.

apps/challenges/views.py
```diff
@@ -340,3 +343,56 @@ def challenge_phase_split_list(request, challenge_pk):
...
+            return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+    # To check for the user as a participant of the challenge from the request and challenge_pk.
+    elif has_user_participated_in_challenge(user=request.user, challenge_id=challenge_pk):
+
+        # get participant team object for the user for a particular challenge.
+        participant_team_pk = get_participant_team_id_of_user_for_a_challenge(
+                request.user, challenge_pk)
+
+        submissions = Submission.objects.filter(participant_team=participant_team_pk,
```
taranjeet: For a challenge host, we are showing submissions corresponding to a challenge but for a participant only for a particular challenge phase. May I know what is the reason behind this.

Also it will be better if we write this in comments. The comment should actually tell why instead of what. What can easily be inferred from the code. Comments which tell why a particular piece of code is there are more helpful.

apps/challenges/views.py
```diff
@@ -340,3 +343,56 @@ def challenge_phase_split_list(request, challenge_pk):
...
+
+    # To check for the corresponding challenge phase from the challenge_phase_pk and challenge.
+    try:
+        challenge_phase = ChallengePhase.objects.get(pk=challenge_phase_pk, challenge=challenge)
+    except ChallengePhase.DoesNotExist:
+        response_data = {'error': 'Challenge Phase {} does not exist'.format(challenge_phase_pk)}
+        return Response(response_data, status=status.HTTP_404_NOT_FOUND)
+
+    # To check for the user as a host of the challenge from the request and challenge_pk.
+    if is_user_a_host_of_challenge(user=request.user, challenge_pk=challenge_pk):
```
taranjeet: A slightly better way of doing this can be catch result of `is_user_a_host_of_challenge` and `has_user_participated_in_challenge` in appropriate variable before. Then first go for the else condition where in you are returning 400. So this will look something like this

```
if not is_challenge_host and not is_participant:
 # return 400 one here
else:
  if is_challenge_host:
    # form submission_params here 
    submission_params['challenge_phase__challenge'] = challenge.

```

Also `ChallengeSubmissionManagementSerializer` and `SubmissionSerializer` can be easily catched in a variable and then used accordingly.

This is just a suggestion. Let me know if you find any difficulty in this or if this is not suitable.

apps/hosts/utils.py
```diff
@@ -1,6 +1,13 @@
 from .models import ChallengeHost
+from challenges.models import Challenge
```
taranjeet: This is just to maintain a better history. Can we please move this in a separate PR?

apps/hosts/utils.py
```diff
@@ -1,6 +1,13 @@
...
 
 
 def get_challenge_host_teams_for_user(user):
     """Returns challenge host team ids for a particular user"""
     return ChallengeHost.objects.filter(user=user).values_list('team_name', flat=True)
+
+
+def is_user_a_host_of_challenge(user, challenge_pk):
```
taranjeet: The function and doc string does not match. Can you please check and update them as per the function defintion?

As per my understanding, function name is correct.

apps/jobs/serializers.py
```diff
@@ -66,14 +66,14 @@ class ChallengeSubmissionManagementSerializer(serializers.ModelSerializer):
     participant_team = serializers.SerializerMethodField()
     challenge_phase = serializers.SerializerMethodField()
     created_by = serializers.SerializerMethodField()
-    participant_team_members_email_id = serializers.SerializerMethodField()
```
taranjeet: I am actually not sure why is this change a part of this PR?

apps/challenges/views.py
RishabhJain2018: Yes , Currently the API doesn't contain the support for the query parameters for a challenge host like `challenge_phase or date` etc. But we will have that support in future. So, for now the submissions are filtered on the basis of `challenge` only. I had a talk regarding this with @deshraj.
Moreover, for the participants the results are being filtered on the basis of `challenge phase `that the participant selects from the dropdown menu on frontend.
Please tell if something is still not clear.

RishabhJain2018: I will surely mention it in comments. :+1:

apps/jobs/serializers.py
RishabhJain2018: I apologise for this mistake. I have created a separate PR for this.

apps/challenges/views.py
RishabhJain2018: Please guide why is this better?
Moreover, I have a question here that in the above code snippet the if condition will always get checked, we are checking a condition every time that is highly unlikely to occur, because the user will be either host or participant, I think we should check a condition first that are most likely to occur. What are your views about it?

taranjeet: @deshraj : your review here please.

deshraj: @taranjeet I think this approach is fine by me and we can go ahead with this. :)

PR #1047, opened 2017-06-16
Frontend: Fix base url in route-config

spyshiv:
As for now templateurl is having extra `/` which is redundant
i.e. `templateUrl` for state `home` is `dist/views//web/landing.html` which should be `dist/views/web/landing.html` .
@aka-jain : Please review and merge this PR.

### DIFF
--- a/frontend/src/js/route-config/route-config.js
+++ b/frontend/src/js/route-config/route-config.js
@@ -7,7 +7,7 @@
         .module('evalai')
         .config(configure);
 
-    var baseUrl = "dist/views/";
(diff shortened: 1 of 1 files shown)


aka-jain:
Yeap, I got that problem before but how this works I mean that's not possible to change view with wrong route.

spyshiv:
I agree that should not work in the case of `//` but somehow it's working. :grin:

aka-jain:
Yes, let's merge it then 😅

spyshiv:
cool :+1:

PR #1055, opened 2017-06-22
Admin: Update ModelAdmin class for various models.

taranjeet:
@RishabhJain2018 : please have a look at it

### DIFF
--- a/apps/jobs/admin.py
+++ b/apps/jobs/admin.py
@@ -13,6 +13,6 @@ class SubmissionAdmin(TimeStampedAdmin, ImportExportModelAdmin):
                     'submission_number', 'submitted_at', 'execution_time', 'input_file', 'stdout_file', 'stderr_file',
                     'submission_result_file', 'submission_metadata_file', )
     list_filter = ('participant_team', 'challenge_phase',
-                   'created_by', 'status', 'is_public')
+                   'status', 'is_public')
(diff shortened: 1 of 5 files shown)


apps/jobs/admin.py
```diff
@@ -13,6 +13,6 @@ class SubmissionAdmin(TimeStampedAdmin, ImportExportModelAdmin):
                     'submission_number', 'submitted_at', 'execution_time', 'input_file', 'stdout_file', 'stderr_file',
                     'submission_result_file', 'submission_metadata_file', )
     list_filter = ('participant_team', 'challenge_phase',
-                   'created_by', 'status', 'is_public')
```
RishabhJain2018: @taranjeet Please explain why we need to remove `created_by` from here?

taranjeet: `created_by` does Foreign key on user which further means that a user can create more than one Submission. With time, the number of submission made by a user will increase, I agree, so will the number of users. A time will come when the list_filter(if used) will populate a user list so long that it will become heavy for the DOM to load it.

Right now if the user count is more than 50, I am sure you will have to scroll page many a times to see the complete list. So for such cases we dont prefer list filter. We will look for some alternate for it in the future.

I hope this clears your doubt.

RishabhJain2018: ok sure, I agree with it.

PR #1058, opened 2017-06-23
Analytics: Api to get submission count for a challenge phase

taranjeet:
@RishabhJain2018 : very basic api to get submission made corresponding to a challenge phase.
Tests are missing, will be adding them soon

### DIFF
--- a/apps/analytics/serializers.py
+++ b/apps/analytics/serializers.py
@@ -1,0 +1,10 @@
+from rest_framework import serializers
+
+
+class SubmissionStats(object):
+    def __init__(self, submission_total):
+        self.submission_total = submission_total
+
+
+class SubmissionStatsSerializer(serializers.Serializer):
+    submission_total = serializers.IntegerField()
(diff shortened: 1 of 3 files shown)


RishabhJain2018:
Can I add the test case?

apps/analytics/serializers.py
```diff
@@ -0,0 +1,10 @@
+from rest_framework import serializers
+
+
+class SubmissionStats(object):
```
deshraj: I would prefer to keep this in the `jobs` app where the model for this serializer resides. :) Please let me know what you think about this.

taranjeet: I am actually not sure if this will go in master or not. So will update only if it needs to be merged. I hope that is fine.

deshraj:
@taranjeet can we please close this if we are not going to merge this?

taranjeet:
Closing since this is not required now.

Issue #840, opened 2017-03-16
Add support in submission worker to update file field changes for Challenge and Challenge Phase


taranjeet:
__Problem__

If the Challenge Host updates any of `evaluation_script` or `test_annotation` file, then there is no way to propagate that change to the worker.

__Please be more clear__ ?

Presently the submission worker when starts fetches all the active challenges and loads `evaluation_script` for each challenge. It also fetches active challenge phase for each challenge and loads `test_annotation` for each phase. We need to extend that code to add support for handling update requests for these files.

__Why__ ?

Imagine a case where in a Challenge Hosts whose challenge is active and hence running, wants to update `evaluation_script` for a challenge. He logs in and updates the file but still sees that submission are being processed using the older version of `evaluation_script`. He wonders why is this happening ??

So we want to help him and add support to propagate the changes done by him to the worker, because it is the worker who is doing the heavy task of processing submission.

__Impact__ ?

It actually guarantees that whenever any submission is evaluated, it is evaluated against the write set of parameters.

__How__ ?

Overall, this will involve refactoring `scripts/workers/submission_worker.py`. The concerned function which extracts challenge and phase data presently is `extract_challenge_data`. This function may be modified or a new function may be added. This will actually require some **thoughts** from the contributor.

One very important thing which needs to be pointed out is how the worker will detect that the file is changed. Here RabbitMQ comes into action. If you see `add_challenge_queue` in `scripts/workers/submission_worker.py`, it binds on key `challenge.*.*`. The first `*` here will help us decide what the intent of the message is. 

Whenever any new challenge will be added, you can assume that `challenge.add.*` key will be there and whenever any update is done, you can assume that `challenge.edit.*` will be published.

So you can now add checks according to the key and then code accordingly.

For now being you can assume that keys published will be `challenge.add.*` and `challenge.edit.*`.

__TakeAways__

About how the submission worker works in detail, RabbitMQ and its related terminologies. This issue will provide you more depth about RabbitMQ and how queue function.

Also I will prefer that before coding, its better to discuss the approach over here only, so that we can get an idea and update if something is not right.

TroJan:
@mayank-agarwal-96 : sure :)

TroJan:
@mayank-agarwal-96 : Let me know if you are struck somewhere. Will be happy to help :)

mayank-agarwal-96:
[...]

PR #1059, opened 2017-06-24
[WIP] Adds support to submission worker for updating challenge fields #840 

yadavankit:
@taranjeet @deshraj please review and add suggestions :) #840


yadavankit:
@taranjeet can you please see if am approaching this in correct way? Also, please suggest changes :)

taranjeet:
@yadavankit : been busy with office these days. Will get back to you positively by tomorrow. Thanks for your time and help :)

scripts/workers/submission_worker.py
```diff
@@ -483,6 +500,9 @@ def main():
     add_challenge_queue_name = '{hostname}_{process_id}'.format(hostname=socket.gethostname(),
                                                                 process_id=str(os.getpid()))
 
+    edit_challenge_queue_name = '{hostname}_{process_id}'.format(hostname=socket.gethostname(),
```
taranjeet: We have planned to use the same queue, but different routing keys.

apps/challenges/models.py
```diff
@@ -80,6 +84,13 @@ def is_active(self):
         return False
 
 
+@receiver(post_save, sender=Challenge)
+def publish_message_on_change(instance, created=False, **kwargs):
```
taranjeet: I think that this PR is missing a very important thing. This signal handler will be called when every any field of model instance is updated and corresponding to it, a message will be published each time. If we clearly see that the worker only needs challenge id (which will never change) to maintain map, and then corresponding files, we come to a realization that the worker should only be notified of changes in file field. 

This way we prevent ourselves and the worker from unnecessarily loading files again and again.

scripts/workers/submission_worker.py
```diff
@@ -461,6 +467,17 @@ def add_challenge_callback(ch, method, properties, body):
         traceback.print_exc()
 
 
+def edit_challenge_callback(ch, method, properties, body):
+    try:
+        logger.info("[x] Received edit challenge message %s" % body)
+        body = yaml.safe_load(body)
+        process_edit_challenge_message(body)
```
taranjeet: As far as the issue is concerned, I think something is missing here also. The file field which have changes needs to be downloaded again and then imported. I think here the code to download those files is missing. 

I may be wrong here, let me know if something is not clear here.

taranjeet:
@yadavankit : Sorry for the delay in reviewing. I have left some major concept level feedback. Let me know if I missed something or if anything is not clear to you.

Thanks for your help.

scripts/workers/submission_worker.py
yadavankit: Sure, will be routing this with different routing keys and same queue.

apps/challenges/models.py
yadavankit: Yeah, I actually wanted to get rid of this too. `post_save` signal provides `update_fields` and that could have done the task, but for some reasons, `update_fields` is `None` when `post_save` signal is invoked. Am looking into that and will follow up asap.

yadavankit: @taranjeet sorry for the delay, can you please help me figure out why `update_fields` is coming `None`. For now, I am trying [this](https://stackoverflow.com/questions/36719566/identify-the-changed-fields-in-django-post-save-signal) solution. Please see if its the correct approach.

RishabhJain2018:
@yadavankit Please update the status of this PR? Is it ready to be merged ?

yadavankit:
@RishabhJain2018 no its not actually, I have been very busy with my Startup's work for the last two months. Am sorry for the delay caused in this, but I couldn't help it.

deshraj:
Thanks for letting us know @yadavankit. 

Closing this since not complete.

isht3:
@yadavankit Can I take over from your progress in this PR? If yes, could you tell me what all should be should be added to this? :smile:

yadavankit:
@isht3 Apologies, but actually am not an active contributor here. Thanks to loads of work at my startup :P You can have a look at the code and comments by @taranjeet for starters

isht3:
@yadavankit That's cool! All the best and thanks!

PR #1065, opened 2017-06-26
[WIP]: Added basic flow for analytics dashboard

aka-jain:
@taranjeet @RishabhJain2018 @deshraj please check the workflow once for this analytics dashboard.

### DIFF
--- a/frontend/src/css/modules/analytics.scss
+++ b/frontend/src/css/modules/analytics.scss
@@ -1,0 +1,38 @@
+li.analytics-list {
+    margin-top: 10px;
+}
+
+.analytics-challenge-title {
+    margin-bottom: 20px;
+}
+
+.analytics-team-input {
+    input[type="radio"].selectTeam + label {
+        padding-left: 0px;
+        width: 100%;
+    }
+}
+
+.show-analytics-btn {
+    margin-top: 10px;
+}
+
+h4.analytics-details-head {
+    margin-top: 2px;
+}
+
+.analytics-image {
+    img {
+        width: 100%;
+    }
+}
+
+.analytics-card-content {
+    h3 {
+        margin-bottom: 2px;
+    }
+}
+
+.analytics-phases-card {
+    margin-top: 10px;
+}
--- a/frontend/src/css/modules/base.scss
+++ b/frontend/src/css/modules/base.scss
@@ -67,6 +67,23 @@ input.dark-autofill:-webkit-autofill:active {
     -webkit-text-fill-color: #4d4d4d !important;
 }
 
+input[type="radio"].selectTeam + label {
+    font-size: 15px;
+    color: $med-black;
+    vertical-align: middle;
+}
+
+[type="radio"]:checked + label:after,
+[type="radio"].with-gap:checked + label:after {
+    background-color: #252833;
+}
+
+[type="radio"]:checked + label:after,
+[type="radio"].with-gap:checked + label:before,
+[type="radio"].with-gap:checked + label:after {
+    border: 2px solid #252833;
+}
+
 body,
 html {
     font-family: 'mons-reg';
@@ -74,6 +91,10 @@ html {
 
 a {
     color: $med-gray;
+}
+
+p.no-top-para {
+    margin-top: 2px;
 }
 
 
@@ -102,8 +123,16 @@ a {
     font-size: 14px;
 }
 
+.fs-16 {
+    font-size: 14px;
+}
+
 .fs-18 {
     font-size: 18px;
+}
+
+.fs-20 {
+    font-size: 20px;
 }
 
 .pointer {
@@ -314,7 +343,7 @@ nav {
     height: auto;
     border-radius: 4px;
     overflow: hidden;
-    box-shadow: 0px 0px 12px #d6dbf1;
+    box-shadow: 0px 0px 12px $shadow-black;
 }
 
 .ev-card-content {
@@ -362,7 +391,7 @@ nav {
 }
 
 .ev-card-hover:hover {
-    box-shadow: 0px 0px 4px #d6dbf1;
+    box-shadow: 0px 0px 4px $shadow-black;
     .btn-card-detail {
         background: $highlight;
         color: #fff;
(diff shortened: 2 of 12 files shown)


aka-jain:
Attached Screenshots 
screenshotatuploadcc_1498499889433
screenshotatuploadcc_1498499889434

deshraj:
@aka-jain: Looks great. 😄 

Can we get rid of the commented lines please?

frontend/src/js/controllers/analyticsCtrl.js
```diff
@@ -0,0 +1,296 @@
...
+                    $state.go("auth.login");
+                    $rootScope.isAuth = false;
+
+                }
+            }
+        };
+        utilities.sendRequest(parameters);
+
+
+        parameters.url = 'challenges/challenge?mode=host';
```
aayusharora: Repeated code. We can make a function that takes `URL, method, token as parameter and array` as parameters.
`$scope.ajaxCall = function(url, token, parameter, arrayName) {`
    `// on successful callback choose which array should the data be pushed.`
 
`}`

frontend/src/js/controllers/analyticsCtrl.js
```diff
@@ -0,0 +1,296 @@
...
+                parameters.callback = {
+                    onSuccess: function(response) {
+                        var status = response.status;
+                        var details = response.data;
+
+
+                        if (status === 200) {
+                            vm.currentPhase = details.results;
+
+                            for (var phaseCount = 0; phaseCount < vm.currentPhase.length; phaseCount++) {
```
aayusharora: Better to use `for in` loop here

frontend/src/css/modules/analytics.scss
```diff
@@ -0,0 +1,38 @@
+li.analytics-list {
```
aka-jain: @aayusharora here .analytics-list is not a child of list element then how can you write this inside the nested loop?

frontend/src/css/modules/base.scss
```diff
@@ -67,6 +67,23 @@ input.dark-autofill:-webkit-autofill:active {
     -webkit-text-fill-color: #4d4d4d !important;
 }
 
+input[type="radio"].selectTeam + label {
```
aka-jain: That's because I wanted to override the materialize css styles by same property. They are not using class instead changing property directly to elements.

frontend/src/js/controllers/analyticsCtrl.js
```diff
@@ -0,0 +1,296 @@
...
+
+        var parameters = {};
+        parameters.url = 'hosts/challenge_host_team/';
+        parameters.method = 'GET';
+        parameters.token = userKey;
+        parameters.callback = {
+            onSuccess: function(response) {
+                var status = response.status;
+                var details = response.data;
+                if (status == 200) {
```
aka-jain: No @aayusharora we don't know wether this status comes as a string or an integer so === might get fail in some cases.

frontend/src/js/controllers/analyticsCtrl.js
```diff
@@ -0,0 +1,296 @@
...
+                    vm.hostTeam = details.results;
+
+                }
+            },
+            onError: function(response) {
+                var status = response.status;
+                var error = response.data;
+                if (status == 403) {
```
aka-jain: Same as above.

frontend/src/js/controllers/analyticsCtrl.js
```diff
@@ -0,0 +1,296 @@
...
+                    vm.error = error;
+
+                    // navigate to permissions denied page
+                    $state.go('web.permission-denied');
+                } else if (status == 401) {
```
aka-jain: Same as above.

frontend/src/js/controllers/analyticsCtrl.js
aka-jain: I think we already using in function as a service here. Don't know how we can do this here?

frontend/src/js/controllers/analyticsCtrl.js
aka-jain: https://stackoverflow.com/questions/500504/why-is-using-for-in-with-array-iteration-a-bad-idea
Check this please.

frontend/src/css/modules/analytics.scss
spyshiv: @aka-jain: We can write something like this. 
**Case 1: For the same element we can write** 
```
li {
    &.analytics-list {}
}
```
**Case 2: for parent child relationship we can write like this**
```
li {
    & .analytics-list {}
}
```
or 
```
li {
    analytics-list {}
}
```

aayusharora: Thanks for this. Giving a class to ul and doing the nesting can be a better approach ?

frontend/src/css/modules/base.scss
aayusharora: `[type="radio"]:checked + label:after` But, this line can override any checked radio in future, How will you protect this?

frontend/src/js/controllers/analyticsCtrl.js
aayusharora: It should be an integer. Why it can be a string ....   status === 200;
please guide.

frontend/src/js/controllers/analyticsCtrl.js
aayusharora: Writing same code three times to make requests for three endpoints does not seem a good solution.
## Proposed Approach
- Make a service name AJAX Requests
- Call this service whenever you want the data with parameters.

frontend/src/js/controllers/analyticsCtrl.js
aayusharora: Thank you. You can take it like this then :)

aka-jain:
@deshraj it's WIP, will clear the code once we done with the final version :)

deshraj:
@aka-jain Any update on this?

aka-jain:
Waiting for the Analytics APIs to be completed, I think they are in WIP.

deshraj:
Ok, sounds good. Thanks for the letting me know.

deshraj:
@aka-jain can you please tell me what all APIs are you waiting for? I might get them done.

RishabhJain2018:
@deshraj I think we should close this one as it is now deployed with the API's.

deshraj:
Sure.

PR #1068, opened 2017-06-27
[REVIEW PENDING]Add test case for analytics Api.

RishabhJain2018:
This PR adds the test case for PR #1058 .

### DIFF
--- a/tests/unit/analytics/test_urls.py
+++ b/tests/unit/analytics/test_urls.py
@@ -1,0 +1,113 @@
+import os
+
+from datetime import timedelta
+
+from django.core.urlresolvers import reverse_lazy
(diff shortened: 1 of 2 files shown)


taranjeet:
@RishabhJain2018 : Will update on this once I have more clarity on what all we need for v1 in analytics. Thanks for the help :)

taranjeet:
I think this wont be required as of now, since we are not merging its corresponding api as of now. Thanks for your help @RishabhJain2018 .

RishabhJain2018:
@taranjeet No problem.

taranjeet:
Thanks for the help @RishabhJain2018 . This is actually not required.

PR #1071, opened 2017-07-01
Base: Add a Standard Pagination class for extending max page size

### LINKED ISSUE #1070: Change the pagination limit in leaderboard
Currently, in the VQA challenge, only 10 entries are visible on the leaderboard and there is no pagination on the client side. To show all the entries of the leaderboard, there are two ways to do that:
1. Add a pagination on the client side
2. Show all the entries on a single page (involves adding a new Pagination Class)

The second approach is a better way since then others will be able to compare their results with everyone else in the team.

deshraj:
Fixes #1070

### DIFF
--- a/apps/base/utils.py
+++ b/apps/base/utils.py
@@ -8,7 +8,7 @@ from rest_framework.exceptions import NotFound
 from rest_framework.pagination import PageNumberPagination
 
 
-class StandardResultsSetPagination(PageNumberPagination):
+class StandardResultSetPagination(PageNumberPagination):
     page_size = 100
     page_size_query_param = 'page_size'
     max_page_size = 1000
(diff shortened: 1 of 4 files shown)


deshraj:
@taranjeet @RishabhJain2018 @yadavankit Can you please review this quickly? We want to get this fixed really soon.

apps/base/utils.py
```diff
@@ -8,11 +8,17 @@
 from rest_framework.pagination import PageNumberPagination
 
 
-def paginated_queryset(queryset, request):
+class StandardResultsSetPagination(PageNumberPagination):
```
taranjeet: Can we make it StandardResultSetPagination please?

PR #1075, opened 2017-07-02
Analytics: Add api to fetch participant team count for challenge

taranjeet:


taranjeet:
@deshraj @RishabhJain2018 : please review.

this api fetches count of teams that participated corresponding to a particular challenge.

apps/analytics/serializers.py
```diff
@@ -0,0 +1,10 @@
+from rest_framework import serializers
+
+
+class ParticipantTeamCount(object):
```
deshraj: Can we please keep this class in `participant` app since this is a part of the participant app? We might use this at other places as well in future.

apps/analytics/serializers.py
```diff
@@ -0,0 +1,10 @@
...
+    def __init__(self, participant_team_count):
+        self.participant_team_count = participant_team_count
+
+
+class ParticipantTeamCountSerializer(serializers.Serializer):
```
deshraj: Same as above. :)

apps/analytics/serializers.py
taranjeet: Sure will do that

taranjeet:
@deshraj : this is ready for review again.

apps/analytics/urls.py
```diff
@@ -1,4 +1,10 @@
 from django.conf.urls import url  # noqa
```
RishabhJain2018: Just a small suggestion here, I think now we can remove `# noqa` from here as there will be no need to ignore this line from `flake8` now.

taranjeet: Nice catch @RishabhJain2018 :)

taranjeet:
@deshraj : this can now be merged.

PR #1037, opened 2017-06-12
[REVIEW PENDING] Add Challenge creation using zip file.

RishabhJain2018:
fixes #1038 @deshraj @taranjeet Please review the PR.

PR #1076, opened 2017-07-02
Modify challenge serializers to accept file input.

RishabhJain2018:

### DIFF
--- a/apps/challenges/serializers.py
+++ b/apps/challenges/serializers.py
@@ -115,7 +115,6 @@ class ZipChallengeSerializer(ChallengeSerializer):
         kwargs['data']['image'] = image
         kwargs['data']['evaluation_script'] = evaluation_script
 
-
     class Meta:
         model = Challenge
         fields = ('id', 'title', 'short_description', 'description', 'terms_and_conditions',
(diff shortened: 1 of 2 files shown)


apps/challenges/serializers.py
```diff
@@ -102,11 +106,22 @@ class ZipChallengeSerializer(ChallengeSerializer):
...
     Serializer used for creating challenge through zip file.
     """
+    def __init__(self, *args, **kwargs):
+        super(ZipChallengeSerializer, self).__init__(*args, **kwargs)
+
+        context = kwargs.get('context')
+        if context:
+            image = context.get('image')
+            evaluation_script = context.get('evaluation_script')
+
```
taranjeet: If context is None, then variables `image` and `evaluation_script` will not be defined. Can you simply assign their value in the if condition.

Also I am still not clear why are we adding this ? Was this missed or some new functionality ?

RishabhJain2018: No, this isn't new feature. it is to offload all the saving task to the serializers. Please refer [here](https://github.com/Cloud-CV/EvalAI/pull/1037#discussion-diff-123895061R489)

PR #1077, opened 2017-07-03
Add a file reading function in challenge utils.

RishabhJain2018:

### DIFF
--- a/apps/challenges/utils.py
+++ b/apps/challenges/utils.py
@@ -1,3 +1,5 @@
+import os
+
 from base.utils import get_model_object
 
 from .models import Challenge, ChallengePhase
@@ -6,3 +8,9 @@ from .models import Challenge, ChallengePhase
 get_challenge_model = get_model_object(Challenge)
 
 get_challenge_phase_model = get_model_object(ChallengePhase)
+
+
+def get_file_content(file_path, mode):
+    if os.path.isfile(file_path):
+        with open(file_path, mode) as file_content:
+            return file_content.read()
--- a/tests/unit/challenges/test_utils.py
+++ b/tests/unit/challenges/test_utils.py
@@ -5,6 +5,7 @@ from django.conf import settings
 
 from challenges.utils import get_file_content
 
+
 class BaseTestCase(unittest.TestCase):
 
     def setUp(self):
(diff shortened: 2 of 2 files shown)


apps/challenges/utils.py
```diff
@@ -6,3 +8,9 @@
 get_challenge_model = get_model_object(Challenge)
 
 get_challenge_phase_model = get_model_object(ChallengePhase)
+
+
+def get_file_content(file_path, mode):
```
taranjeet: May I know why are we adding this function? I am still not able to infer its appropriate use.

RishabhJain2018: Actually earlier we were reading all the data from the yaml file like `challenge_description` , `terms_and_conditions` etc. but as this content will be long, the yaml file configuration will not look good. So, now we are reading this content also from the different files, that's why I created this utility function to remove the redundancy in code. Please tell if I am not clear somewhere.

taranjeet: Okay I got it.

apps/challenges/utils.py
```diff
@@ -6,3 +8,9 @@
...
+    if os.path.isfile(file_path):
+        with open(file_path, mode) as file_content:
```
taranjeet: The function intends to be general, thats great actually But here we are only reading the file and returning its content. If this is concerned, then I think mode should have a default value of read(`r`). Infact we can drop this mode as parameter if we are sure that there wont be any binary (mode=rb) file.

@deshraj : your thoughts here.

tests/unit/challenges/test_utils.py
```diff
@@ -0,0 +1,17 @@
...
+
+from django.conf import settings
+
+from challenges.utils import get_file_content
+
+
+class BaseTestCase(unittest.TestCase):
+
+    def setUp(self):
+        self.test_file_path = os.path.join(settings.BASE_DIR, 'examples', 'example1', 'test_annotation.txt')
```
taranjeet: Nice use of file. :)

apps/challenges/utils.py
deshraj: I think this looks good for now and we can use `r` as default parameter later on.

PR #1078, opened 2017-07-03
Analytics: Add api to get no of participants for a particular challenge.

taranjeet:
@deshraj @RishabhJain2018 : api to get participant count for a particular challenge.


taranjeet:
@deshraj : can we merge this in master?

PR #1083, opened 2017-07-09
Analytics: Add API to get submission count for a particular challenge.

taranjeet:
@deshraj @RishabhJain2018

### DIFF
--- a/apps/analytics/views.py
+++ b/apps/analytics/views.py
@@ -1,3 +1,7 @@
+from datetime import timedelta
+
+from django.utils import timezone
+
 from rest_framework import permissions, status
 from rest_framework.decorators import (api_view,
                                        authentication_classes,
@@ -11,6 +15,10 @@ from accounts.permissions import HasVerifiedEmail
 
 from challenges.permissions import IsChallengeCreator
 from challenges.utils import get_challenge_model
+from jobs.models import Submission
+from jobs.serializers import (SubmissionCount,
+                              SubmissionCountSerializer,
+                              )
 from participants.models import Participant
 from participants.serializers import (ParticipantCount,
                                       ParticipantCountSerializer,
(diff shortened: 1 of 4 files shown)


apps/analytics/views.py
```diff
@@ -42,3 +50,41 @@ def get_participant_count(request, challenge_pk):
...
+@permission_classes((permissions.IsAuthenticated, HasVerifiedEmail, IsChallengeCreator))
+@authentication_classes((ExpiringTokenAuthentication,))
+def get_submission_count(request, challenge_pk, duration):
+    '''
+        Returns submission count for a challenge according to the duration
+        Valid values for duration are all, daily, weekly and monthly.
+    '''
+    # make sure that a valid url is requested.
+    if duration.lower() not in ('all', 'daily', 'weekly', 'monthly'):
+        response_data = {'error': 'Wrong url pattern!'}
```
deshraj: Please change the message to `Wrong URL pattern!`

apps/analytics/views.py
```diff
@@ -42,3 +50,41 @@ def get_participant_count(request, challenge_pk):
...
+
+    challenge_phase_ids = challenge.challengephase_set.all().values_list('id', flat=True)
+
+    q_params = {'challenge_phase__id__in': challenge_phase_ids}
+    since_date = None
+    if duration.lower() == 'daily':
+        since_date = timezone.now().date()
+
+    elif duration.lower() == 'weekly':
+        since_date = (timezone.now() - timedelta(days=1)).date()
```
deshraj: Shouldn't the timedelta should be `7` instead of `1`? Please let me know if I am wrong.

taranjeet: yeah sure, just missed it.

apps/analytics/views.py
taranjeet: Yeah, I will update that.

taranjeet:
@deshraj : I have updated the changes.

PR #1091, opened 2017-07-14
Docs: Add docs for challenge creation

deshraj:


docs/source/challenge_creation.md
```diff
@@ -0,0 +1,195 @@
...
+    max_submissions_per_day: 100
+    max_submissions: 1000
+
+dataset_splits:
+  - id: 1
+    name: Name of the dataset split
+    codename: codename of dataset split
+  - id: 2
+    name: Name of the dataset split
+    codename: codename of dataset split
```
RishabhJain2018: I think `codename` can't be same in two dataset splits. Can we modify it to `another codename of dataset split`?

deshraj: No, we don't want to give the capability to edit the dataset split codename later.

RishabhJain2018: @deshraj I didn't get the point exactly. Please explain.

deshraj:
> Moreover, I think we can also provide the link to EvalAI-Examples repository for the detailed example of the zip file and its contents. @deshraj What do you say?

@RishabhJain2018 there is some amount of work that we need to do on that examples repository before linking it in the docs. Let's get that done first and then we can add it to the docs.

docs/source/challenge_creation.md
deshraj: Changing the dataset split codename in the database is fine but then the challenge host needs to do the same change in his evaluation script. So, it's better if we only restrict them before they mess up the challenge creation process on their end.

PR #1094, opened 2017-07-16
Base: Add admin base class for import export

taranjeet:
@deshraj @RishabhJain2018 

So that we can use a single class for our admin classes, like

```
@admin.register(DatasetSplit)
class DatasetSplitAdmin(ImportExportTimeStampedAdmin):
    list_display = ("name", "codename")
    list_filter = ("name", "codename")
    search_fields = ("name", "codename")
```


apps/base/admin.py
```diff
@@ -1,5 +1,11 @@
...
 
+from import_export.admin import ImportExportMixin
+
 
 class TimeStampedAdmin(admin.ModelAdmin):
     exclude = ('created_at', 'modified_at',)
+
+
+class ImportExportTimeStampedAdmin(ImportExportMixin, admin.ModelAdmin):
+    exclude = ('created_at', 'modified_at',)
```
deshraj: I think the challenge host or EvalAI Admin would need these fields to analyze the data based on timestamp, so let's not remove these fields.

taranjeet: We are already using `TimeStampedAdmin` for all of our `ModelAdmin` classes. 

This `excludes` means that fields `created_at` and `updated_at` will not be displayed in edit view of any row. It in any ways has no relation with csv import export. For that, we define `exclude` and `fields` in `ModelResource` class.

I hope you got my point.

deshraj: Sorry, my bad. Got your point.

PR #1096, opened 2017-07-17
Challenges: Use ImportExportTimeStampedAdmin for model admin class

taranjeet:
@deshraj @RishabhJain2018 

Also @RishabhJain2018 : if this PR is fine and merged, can you do the same for other apps as well ?


RishabhJain2018:
@taranjeet yes sure. I will add it for other apps.

PR #1100, opened 2017-07-22
Adds deconstructible class for creating migrations.

RishabhJain2018:
Moreover , please refer [here](https://docs.djangoproject.com/en/1.10/topics/migrations/#serializing-values)


deshraj:
@RishabhJain2018: Can you please give some context that why are we doing this?

RishabhJain2018:
@deshra The issue here is that the migrations system needs to serialize the function definition, and isn't able to do so with the dynamically-created object returned by `RandomFileName` Class. To fix this the solution is given in docs to add a `@deconstructible` decorator. Please refer here https://docs.djangoproject.com/en/1.10/topics/migrations/#serializing-values

deshraj:
Looks good. Merging this one.

PR #1102, opened 2017-07-22
Add models for challenge phase description file.

### LINKED ISSUE #1101: Add models to store challenge phase description file.

RishabhJain2018:
fixes #1101 @deshraj @taranjeet Please review the PR.


deshraj:
@RishabhJain2018: I was thinking of dropping this idea of adding the new model. Instead, can we have the option of the uploading the image using the URL? That would be a much cleaner way of adding the images in the description field.

RishabhJain2018:
@deshraj Yes, that can also be one approach but won't this will be overhead for user to upload the image somewhere and then paste the url here. What do you say ?

deshraj:
Closing this one since we are not planning to go ahead with this feature as of now. We might reopen this in the future and then merge it.

PR #1105, opened 2017-07-23
Typo fix in jobs urls.

RishabhJain2018:


deshraj:
@RishabhJain2018 Can you please add test for the same so as to make sure that it is not affecting the previously defined URL patterns?

RishabhJain2018:

PR #1126, opened 2017-08-06
Add automatic reloading of rabbitmq worker from django admin.

### LINKED ISSUE #1128: Add automatic reloading of rabbitmq worker from django admin.

RishabhJain2018:
- [x] **Dependency:** #1161 should be merged before this PR.
fixes #1128 @deshraj Please review the PR.


apps/challenges/templates/rabbitmq_not_reloaded.html
```diff
@@ -0,0 +1,9 @@
+<!DOCTYPE html>
```
taranjeet: We can use same template here. Just pass some context variable to maintain your checks.

apps/challenges/urls.py
```diff
@@ -33,5 +33,8 @@
         name='get_all_submissions_of_challenge'),
     url(r'(?P<challenge_pk>[0-9]+)/download_all_submissions_file/(?P<file_type>[A-Za-z]+)/$',
         views.download_all_submissions_file,
-        name='download_all_submissions_file')
+        name='download_all_submissions_file'),
+    url(r'challenges/reload_rabbitmq_submission_worker/$',
```
taranjeet: Submission worker is a python script. Rabbit mq is a messaging broker. I dont think there is a need to include `rabbitmq` in urls. In fact make the end point as 

```
challenges/submission_worker/reload
```

This way in future(not sure here) if we want to add an option for restart, we could simply expose `/restart` endpoint or a better way can be to take it as keyword argument in the urls like 

```
challenges/submission_worker/:action
```

apps/challenges/views.py
```diff
@@ -791,3 +793,11 @@ def download_all_submissions_file(request, challenge_pk, file_type):
     else:
         response_data = {'error': 'The file type requested is not valid!'}
         return Response(response_data, status=status.HTTP_400_BAD_REQUEST)
+
+
+def reload_rabbitmq_submission_worker(request):
+    if request.user.is_superuser:
+        subprocess.call("scripts/workers/restart_rabbitmq_worker.sh", shell=True)
```
taranjeet: I think if we are specifying interpreter at the beginning, there is no need to add `.sh` extension to it.
Can we please drop it?

scripts/workers/restart_rabbitmq_worker.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+
+#    List the process ids for all the running consumer or receiver processes.
```
taranjeet: Lets make the comment in present imperative tense. like `Lists`, `Kills`, `Runs` and `Lists`.
Can we update the comments

```
Lists the process ids for all the running worker processes.
Kills the running worker processes.
Runs new processes for worker and lists their ids.
```

scripts/workers/restart_rabbitmq_worker.sh
```diff
@@ -0,0 +1,19 @@
...
+#    Kill the running consumer or receiver processes.
+#    Run new instances of the consumer or receiver processes.
+#    List the process ids of the running consumer or receiver processes.
+
+for pid in `ps -ef | grep 'scripts/workers/submission_worker.py' | grep -v 'grep' | awk '{print $2}'` ; do echo $pid ; done
+
+for pid in `ps -ef | grep 'scripts/workers/submission_worker.py' | grep -v 'grep' | awk '{print $2}'` ; do kill $pid ; done
+
+# Running four consumer or receiver processes in background. We can increase its number for further requirements.
```
taranjeet: This is a hack. We need to add support in worker, where in it could run itself as a master  process and then fork N(here 4) child out of it.
But as of now we can go ahead with this.

scripts/workers/submission_worker.py
```diff
@@ -212,7 +212,7 @@ def load_active_challenges():
     '''
          * Fetches active challenges and corresponding active phases for it.
     '''
-    q_params = {'published': True}
+    q_params = {'published': True, 'approved_by_admin': True}
```
taranjeet: This should go in a separate PR.

settings/common.py
```diff
@@ -87,7 +87,7 @@
 TEMPLATES = [
     {
         'BACKEND': 'django.template.backends.django.DjangoTemplates',
-        'DIRS': [],
+        'DIRS': [APPS_DIR],
```
taranjeet: Isn't there any other way to add a `<li>` element to Django admin. I used `django-admin-tools` which was pretty configurable and quick links can be very easily added without updating the html. @deshraj : your thoughts here.

apps/challenges/views.py
RishabhJain2018: @taranjeet @deshraj I didn't get this point . Please explain a bit further.

deshraj:
@RishabhJain2018 can you please resolve the conflicts?

RishabhJain2018:
@deshraj Done!

scripts/workers/restart_rabbitmq_worker.sh
```diff
@@ -0,0 +1,18 @@
...
+python 'scripts/workers/submission_worker.py' &
+python 'scripts/workers/submission_worker.py' &
+python 'scripts/workers/submission_worker.py' &
+python 'scripts/workers/submission_worker.py' &
```
deshraj: Instead of hardcoding this four times, can set a variable that decides that how many worker processes we need to run?

settings/common.py
deshraj: @taranjeet do you think that it is a good idea to have Django admin tools for EvalAI? IMO: it would be really easy for the admins to manage different challenges via Admin interface. Let me know what you think about this.

scripts/workers/restart_rabbitmq_worker.sh
RishabhJain2018: Yes sure. I will update this accordingly.

settings/common.py
taranjeet: I think there should be a admin dashboard for challenge host where in they have some quick actions and settings. Django admin is good but it is only suitable for evalai admins. For challenge host they should be provided with some ui friendly dashboard.

deshraj:
I am closing this since we are now planning to shift to AWS SQS instead of RabbitMQ.

PR #1127, opened 2017-08-06
Fix #1121 

### LINKED ISSUE #1121: UI issues on signup/login page
I found some UI/UX related issues in the signup page.

- [x] Upon updating profile, the notification displayed has a minor UI issue.
screen shot 2017-08-05 at 1 35 22 pm

Ideally the exclamation mark should come in the same line as the update message.

- [x] Same issue exists when you sign up

screen shot 2017-08-05 at 1 50 30 pm

- [x] In the Sign up page, the password field only blocks copying by keyboard shortcut.
It allows you to copy the password if you do it via mouse (right click and then copy).
This is not a good UX approach and the copying should be either totally allowed or totally blocked.

Browser : Safari
OS : OSX

the-dagger:
Fix : #1121 
- [x] Disable pasting in password fields


aayusharora:
@dagger The problem still persists with the breakage of `address` word.

deshraj:
We would probably just change the notifications to the materializecss toasts once @aka-jain confirms it.

the-dagger:
@deshraj Alright, then I'll just solve the issue with password field in this PR.

aayusharora:
@deshraj The issue will remain the same either you use any of them. 
The only solution here is to write a code that can check how many words the container can take in and append `-` at the end of the line ( i.e when the container reaches its limit )
```
Like Registered Successfully. Please verify your email addr-
ess.
```

the-dagger:
@aayusharora a better option is to limit the number of lines shown in the toast and modify the width of toast accordingly.

the-dagger:
@deshraj @aayusharora made the changes, please review.

spyshiv:
For notification UI @the-dagger @aayusharora  No need to count the words etc. etc. I have figured it out. 
I am taking up UI issue in the notification. I will mention you in a PR regarding it.

aayusharora:
Thanks @spyshiv It will be interesting to see, how you handle this on different screen sizes then.

aayusharora:
@the-dagger @deshraj  Taking the `Notification` issue from this PR, making it a separate issue.

frontend/src/js/controllers/authCtrl.js
```diff
@@ -85,7 +85,7 @@
                         if (response.status == 201) {
                             vm.isFormError = false;
                             // vm.regMsg = "Registered successfully, Login to continue!";
-                            $rootScope.notify("success", "Registered successfully. Please verify your email address!");
+                            $rootScope.notify("success", "Registered successfully.Please verify your email address!");
```
deshraj: We want to have a space after the full stop. :)

the-dagger: Done!

aayusharora:
@the-dagger Will you please remove the changes that are not required 
   ```
 toaster.pop({
                type: type,
                body: message,
                timeout: timeout
                timeout: timeout,
                **bodyOutputType: 'trustedHtml'**
              });
```

the-dagger:
@aayusharora But don't we need `bodyOutputType: 'trustedHtml'`?
Toast wasn't accepting newline because of this.

aayusharora:
@dagger This issue is addressed in another PR ( It was a CSS fix ), hence we don't require `\n` in this case.

the-dagger:
Alright, thanks for the heads up.
Fixed it 👍

deshraj:
Thanks @the-dagger. Will merge this one! :)

Issue #1132, opened 2017-08-07
Fix Notification Message Styling


aayusharora:
- [ ] Fix notification on sign up
<img width="414" alt="screen shot 2017-08-06 at 6 24 33 pm" src="https://user-images.githubusercontent.com/6565270/29003583-d7236dbc-7ad6-11e7-8c7c-567ea83b2a12.png">

- [ ] Fixed notification on profile update

PR #1133, opened 2017-08-07
Frontend: Fix word break in notification ui

spyshiv:
Fix for #1132 
Additonal changes:
* Move notification override scss into notifictaion.scss from base.scss
* Remove gulp-purify-css since no need to use this.
* Code commented for change in colors for notification UI. @aka-jain will provide colors according to theme.

@aayusharora @the-dagger : The problem was with package `gulp-purifycss`. It removes css of notification override since the toaster element can't be found inside the paths given in the config. I could also provide the path for `toaster.js` in order to do so. But I am sure there is no need to use purify css in this way. Read: [How purifycss works](https://github.com/purifycss/purifycss) if you want to use it in future. also see: [gulp-purifycss](https://github.com/purifycss/gulp-purifycss)

PS: first of all I have changed some css to override notification ui but it was not reflecting into `dist/main.css`. Then i go through gulp task `css` and figured out the problem.

Must read [word wrap](https://www.w3.org/TR/css-text-3/#overflow-wrap-property) and [word break](https://www.w3.org/TR/css-text-3/#word-break)

Screenshots
screenshot from 2017-08-07 22 26 00
screenshot from 2017-08-07 22 26 57
screenshot from 2017-08-07 22 28 48
screenshot from 2017-08-07 22 29 09


frontend/src/css/modules/notification.scss
```diff
@@ -1,17 +1,52 @@
...
+    &.toast-error {
+        //     background-color: #bd362f;
+    }
+
+    &.toast-info {
+        //     background-color: #2f96b4;
+    }
+
+    &.toast-wait {
+        //     background-color: #2f96b4;
```
the-dagger: probably remove these fields instead of commenting them out?

spyshiv: These are commented out intentionly. It will save time while overriding the default colors of notification ui. @aka-jain : Please provide colors

spyshiv:
@the-dagger @aayusharora : Please verify that now things are fine with notification UI so that i can merge this PR.

frontend/src/css/modules/notification.scss
the-dagger: Ah I see.
Why don't we use Material toasts though?
They look and perform better than the Angular Toaster.
https://material.angularjs.org/latest/demo/toast

spyshiv: Thanks for the suggestion. There is no hidden purpose in not using Material Toasts. Actually we were avoiding using angular material since it had many issues back then. But Material Toasts looks awesome :+1:. So In the mean time we can integrate this. I will discuss using of it with @aka-jain then notify you.

frontend/src/css/modules/notification.scss
```diff
@@ -1,17 +1,52 @@
...
-		opacity: 1;
-		-moz-border-radius: 3px 0px 0px 3px;
-		-webkit-border-radius: 3px 0px 0px 3px;
-		border-radius: 3px 0px 0px 3px;
-		box-shadow: 0 0 6px #9d9d9d;
-	}
+    top: 64px;
+    right: 0;
+    box-shadow: none;
+    &>:hover {
```
aayusharora: Do we need a `>` here ?

aayusharora:
@spyshiv I tried it with `word-wrap: break-word` earlier, but it was overwritten as you said.
We used `purify-css` to remove unused css, and there will be a lot of unused CSS in coming development which will be left without using it.
How can we going to tackle that problem then?
Is it a right solution for removing this?
Can we use `!important` tag here so that it gets the `global` importance.

spyshiv:
`gulp-purifycss` will remove css having `important!` tag too. Because it looks into specified html and js file and if there is no such css selector(html tags/elements) found in it it remove that css part from your code. 

So basically we can insert path of js libs. in order to tackle the problem. We should do it as different task. It also takes extra time. So running it on small changes doesn't seems good. Sometimes it halts gulp dev:runserver too for more than 10 seconds.

I think we should run this before git push/commit. There are some packages for doing tasks before git push/commit. Same we can do for `unit tests` and `lint`. Since It is too annoying to test, lint in too small changes with `gulp watch` . What do you think?

aayusharora:
I too analyzed the complexity of doing this at the current stage. Probably, this files can only be added as a separate task then. But, the point is, we only need this, when the app will go into `production` phase.

deshraj:
@spyshiv I will test it and will merge soon.

spyshiv:
@deshraj : sure. Also please ping me in case of any error.

deshraj:
@spyshiv I just checked out to this PR and it doesn't seem to work properly. Please see the screenshot attached whenever you get time. 
screen shot 2017-08-08 at 6 09 07 am

spyshiv:
@deshraj : That's strange. It's working well on my machine.
screenshot from 2017-08-08 16 30 48
Which browser are you using?? I have checked it for google-chrome and firefox and it's working well on both.

frontend/src/css/modules/notification.scss
spyshiv: Yes. It is used in toaster lib's default css. So we have to use it to override.

deshraj:
@spyshiv Sorry my bad. That was a cache issue. Working perfectly now. Thanks for fixing this. :)

spyshiv:
cool :smile:

Issue #1155, opened 2017-08-10
Remove pagination from challenge phase split list API


deshraj:
## Observed behavior

The Challenge Phase Split list API endpoint currently returns the challenge phase splits in a paginated manner. But while showing the leaderboard, we need the list of all the challenge phase splits requesting once. 

## Objective

Remove the pagination from this API (https://github.com/Cloud-CV/EvalAI/blob/master/apps/challenges/views.py#L357).

taranjeet:
@deshraj : Can we instead of removing pagination, take a query parameter like noPage or some better name, which if presents, causes the api to return complete result set instead of pagination?

PR #1158, opened 2017-08-10
Remove pagination from challenge phase split list API

S2606:
Fix https://github.com/Cloud-CV/EvalAI/issues/1155: pagination has been removed from the API

### DIFF
--- a/apps/challenges/urls.py
+++ b/apps/challenges/urls.py
@@ -23,8 +23,6 @@ urlpatterns = [
         views.get_challenge_by_pk, name='get_challenge_by_pk'),
     url(r'challenge$', views.get_challenges_based_on_teams,
         name='get_challenges_based_on_teams'),
-    url(r'(?P<challenge_pk>[0-9]+)/challenge_phase_split$', views.challenge_phase_split_list,
-        name='challenge_phase_split_list'),
     url(r'challenge/challenge_host_team/(?P<challenge_host_team_pk>[0-9]+)/zip_upload/$',
         views.create_challenge_using_zip_file,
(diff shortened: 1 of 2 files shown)


S2606:
@deshraj Please review the PR

apps/challenges/urls.py
```diff
@@ -23,8 +23,6 @@
         views.get_challenge_by_pk, name='get_challenge_by_pk'),
     url(r'challenge$', views.get_challenges_based_on_teams,
         name='get_challenges_based_on_teams'),
-    url(r'(?P<challenge_pk>[0-9]+)/challenge_phase_split$', views.challenge_phase_split_list,
```
RishabhJain2018: @S2606 We don't have to remove the endpoint. We need to remove only the pagination from the API.
So, please revert these changes.

apps/challenges/views.py
```diff
@@ -352,25 +352,6 @@ def challenge_phase_detail(request, challenge_pk, pk):
         return Response(status=status.HTTP_204_NO_CONTENT)
 
 
-@throttle_classes([AnonRateThrottle])
```
RishabhJain2018: The PR remove the complete view function which should not be the case.
In order to dive into `pagination`, please read [here](http://www.django-rest-framework.org/api-guide/pagination/).

deshraj:
@S2606 please let us know if you are working on this otherwise we would ideally close this.

S2606:
please close the issue

RishabhJain2018:
@deshraj I think we should close this.

deshraj:
Thanks for letting us know @S2606. Closing it.

PR #1168, opened 2017-08-11
Fix the download submission file url in challenge controller.

RishabhJain2018:
@deshraj Please review this and merge this asap as the feature won't work without it.


frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -961,7 +961,7 @@
         vm.downloadChallengeSubmissions = function() {
             if(vm.phaseId) {
                 var parameters = {};
-                parameters.url = "challenges/"+ vm.challengeId + "/phase/" + vm.phaseId + "/download_all_submissions_file/" + vm.fileSelected + "/";
+                parameters.url = "challenges/"+ vm.challengeId + "/phase/" + vm.phaseId + "/download_all_submissions/" + vm.fileSelected + "/";
```
taranjeet: I am missing here the reason for this change. @RishabhJain2018 can you please point me out ?

RishabhJain2018: Yes, Actually during review @deshraj suggested me to change the `endpoint` but I forgot to update in the controller. That's why the feature isn't working.
Please refer [here](https://github.com/Cloud-CV/EvalAI/blob/master/apps/challenges/urls.py#L34)

taranjeet:
@RishabhJain2018 : Thanks :)

PR #1171, opened 2017-08-12
Add models to star a challenge.

### LINKED ISSUE #1170: Add models to star a challenge
**Deliverable**:

- [x] Add a model in `apps/challenges/model.py` for starring a challenge.

**Attributes:** 
1. `user` - Foreign key to `User` model
2. `challenge` - Foreign key to `Challenge` model.

- [x] Add the migration file.

- [x] Register the model in django-admin.

RishabhJain2018:
Fix #1170 . @deshraj Please review the PR.


apps/challenges/models.py
```diff
@@ -220,3 +220,15 @@ class ChallengeConfiguration(TimeStampedModel):
     class Meta:
         app_label = 'challenges'
         db_table = 'challenge_zip_configuration'
+
+
+class StarChallenge(TimeStampedModel):
```
taranjeet: Should we track an `unstar` action after a user has starred ?

If yes then tb schema needs to be updated.

apps/challenges/migrations/0029_adds_model_for_starring_a_challenge.py
```diff
@@ -0,0 +1,31 @@
+# -*- coding: utf-8 -*-
```
deshraj: Agree.

apps/challenges/models.py
deshraj: Umm...I don't think we really need this feature. In my opinion, we should be fine with this. @taranjeet please let me know if I am missing of some analytics that will be super useful for us with this feature.

taranjeet: I think one analytics can be to track an unstar action after a user has starred a challenge. This may be too much thinking, but this is something which the challenge host should know

RishabhJain2018: Just a curiosity question here... :sweat_smile: that there can be many users who will star and then unstar, what can a challenge host do by knowing about them ?

deshraj: @taranjeet lets just ignore the feature of storing how many times a challenge was starred/unstarred. This doesn't seem too promising to me.

PR #1192, opened 2017-08-18
chore : Added .DS_Store to .gitignore

the-dagger:
Update gitignore to ignore the DS_Store files added by OSX.


.gitignore
```diff
@@ -89,4 +89,4 @@ challenge_data/*
 media/
 
 # .DS_Store for OSX
-*/.DS_Store
+**.DS_Store
```
taranjeet: Simply `.DS_Store` would work for any directory. :)

the-dagger: Yep, that's true.
Thanks, updated.

PR #1193, opened 2017-08-19
Add Boolean field in star models

RishabhJain2018:


apps/challenges/models.py
```diff
@@ -228,6 +228,7 @@ class StarChallenge(TimeStampedModel):
     """
     user = models.ForeignKey(User)
     challenge = models.ForeignKey(Challenge)
+    is_starred = models.BooleanField(default=False)
```
taranjeet: Entry in this model will be made only when a user stars a particular challenge. Why an extra boolean field here ?

RishabhJain2018: Without this field when a user will unstar a challenge then we have to delete the entry from this model but now with this boolean field when a user will unstar it then we will toggle this field which will maintain our database schema indexes. Moreover, we can also get the analytics like how many people starred the challenge and how many are currently active.

taranjeet: I thought we were not tracking unstar action.

deshraj: We have done this so that we can get information in future if we need some analytics for this. :)

PR #1199, opened 2017-08-22
Jobs: Update serializer for when made public timestamp.

### LINKED ISSUE #1180: Update when made public timestamp.
**Current behaviour:** Currently we are not updating `when_made_public` timestamp when a user makes his submission public.
**Expected behaviour:** When a user makes his submission public then update this timestamp.
The timestamp must be in` UTC`.

**Dependency:** Dependent on the PR #1179

RishabhJain2018:
fixes #1180 @deshraj Please review the PR.


RishabhJain2018:
@deshraj @taranjeet The issue in this was when the `save` method of `Submission` model is called during **the updating of submission meta-data or status**, the `completed_at` field also gets updated due to this line.
I have fixed it but please check thoroughly before merging it as it will affect the `execution_time` for the submission.

RishabhJain2018:
@deshraj @taranjeet Please review as it is very important to be merged.

PR #1200, opened 2017-08-22
Update controller for when made public field.

### LINKED ISSUE #1180: Update when made public timestamp.

RishabhJain2018:
fixes #1180. @deshraj Please review the PR.

### DIFF
--- a/frontend/src/js/controllers/challengeCtrl.js
+++ b/frontend/src/js/controllers/challengeCtrl.js
@@ -888,12 +888,25 @@
         };
 
         vm.changeSubmissionVisibility = function(submission_id) {
+            var dateTime = new Date();
+            dateTime = dateTime.toISOString();
+            var splitDateTime = dateTime.split("T");
+            var date = splitDateTime[0];
+            var time = splitDateTime[1].split(".")[0];
+
             var parameters = {};
             parameters.url = "jobs/challenge/" + vm.challengeId + "/challenge_phase/" + vm.phaseId + "/submission/" + submission_id;
             parameters.method = 'PATCH';
-            parameters.data = {
-                "is_public": vm.submissionVisibility[submission_id]
-            };
+            if (vm.submissionVisibility[submission_id]) {
+                parameters.data = {
+                    "is_public": vm.submissionVisibility[submission_id],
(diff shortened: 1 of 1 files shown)


frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -888,12 +888,25 @@
         };
 
         vm.changeSubmissionVisibility = function(submission_id) {
+            var dateTime = new Date();
+            dateTime = dateTime.toISOString();
+            var splitDateTime = dateTime.split("T");
+            var date = splitDateTime[0];
+            var time = splitDateTime[1].split(".")[0];
```
deshraj: The datetime should be computed at the backend and not at the frontend. Both are decouped.

deshraj: The datetime should be computed at the backend and not at the frontend. Both frontend and backend are decoupled.

RishabhJain2018: okay sure.

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -888,12 +888,25 @@
...
+
             var parameters = {};
             parameters.url = "jobs/challenge/" + vm.challengeId + "/challenge_phase/" + vm.phaseId + "/submission/" + submission_id;
             parameters.method = 'PATCH';
-            parameters.data = {
-                "is_public": vm.submissionVisibility[submission_id]
-            };
+            if (vm.submissionVisibility[submission_id]) {
+                parameters.data = {
+                    "is_public": vm.submissionVisibility[submission_id],
```
aayusharora: Please store ` vm.submissionVisibility[submission_id]` in  a seperate variable, and use it everywhere @RishabhJain2018 in this function

RishabhJain2018:
Since we will be calculating the time at the backend, so I am closing this one.

PR #1215, opened 2017-08-27
Minor fix in README

xamfy:
Change install to installed in point 8


deshraj:
Thanks for fixing @xamfy. Merging it.

PR #1222, opened 2017-09-02
Added `^` symbol before url paths in url.py files

### LINKED ISSUE #1216: Add ^ & $ in regex of urls for proper matching
' `^` ' sign should be introduced in the regular expressions of the url's, mentioned in the url files so that our search is always matched up correctly.

For example:  
If we have not added the initial caret character (i.e., 'hello/$'), Django would match any URL that ends with hello/, such as /my/site/hello/.
Also it will be more syntactically correct.

HargovindArora:
Fix #1216 . 
Url files modeified by adding `^` symbol before url patterns for correct path matching.


deshraj:
@HargovindArora can you please check that why the travis build is failing?

HargovindArora:
@deshraj I am closing this PR and making a new one.

PR #1235, opened 2017-09-09
Fix #1228: Text being displayed Star challenge feature

### LINKED ISSUE #1228: Text not being displayed for the 'star challenge feature'
Please see the below image for the illustration:

screen shot 2017-09-06 at 6 09 47 pm

Ram81:
Fixed #1228 @RishabhJain2018 review.

### DIFF
--- a/frontend/src/js/controllers/challengeCtrl.js
+++ b/frontend/src/js/controllers/challengeCtrl.js
@@ -1065,11 +1065,14 @@
                 if (details['is_starred'] === false){
                     vm.data = 'Star';
                 }
-                else{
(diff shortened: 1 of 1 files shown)


aka-jain:
@Ram81 Just initialise star count with 0 at the start of controller and it will solve the problem.

Ram81:
@aka-jain count variable is being fetched from response that's why initially when there is no record of star challenge it fails.

RishabhJain2018:
@Ram81 Thanks for the fix! I am closing this now as this is already fixed. :)

PR #1248, opened 2017-09-14
Fix #1139: Disable the Challenge Phase buttons which are inactive while making submissions.

### LINKED ISSUE #1139: Disable the Challenge Phase buttons which are inactive while making submissions
## Observed Behaviour
Currently, if a challenge phase is inactive then also a participant can select the phase to make submissions. 

See screenshot for example:
screen shot 2017-08-08 at 7 21 16 pm

## Expected Behaviour

The challenge phases that are not active now should be ideally disabled and the user shouldn't be allowed to select those challenge phases.

Ram81:
@RishabhJain2018 Fixed.

### DIFF
--- a/frontend/src/js/controllers/challengeCtrl.js
+++ b/frontend/src/js/controllers/challengeCtrl.js
@@ -81,8 +81,6 @@
                         onSuccess: function(response) {
                             var details = response.data;
                             vm.currentDate = details.time;
-                            console.log(userKey);
-                            console.log(details.time);
                             for (var i in details.challenge_participant_team_list) {
                                 if (details.challenge_participant_team_list[i].challenge !== null && details.challenge_participant_team_list[i].challenge.id == vm.challengeId) {
                                     vm.isParticipated = true;
--- a/frontend/src/js/controllers/challengeCtrl.js
+++ b/frontend/src/js/controllers/challengeCtrl.js
@@ -32,7 +32,7 @@
         vm.isChallengeHost = false;
         vm.stopLeaderboard = function() {};
         vm.stopFetchingSubmissions = function() {};
-        vm.currentDate = new Date().toISOString();
+        vm.currentDate = null;
 
 
         // loader for existing teams
@@ -80,6 +80,9 @@
                     parameters.callback = {
                         onSuccess: function(response) {
                             var details = response.data;
+                            vm.currentDate = details.time;
+                            console.log(userKey);
+                            console.log(details.time);
                             for (var i in details.challenge_participant_team_list) {
                                 if (details.challenge_participant_team_list[i].challenge !== null && details.challenge_participant_team_list[i].challenge.id == vm.challengeId) {
                                     vm.isParticipated = true;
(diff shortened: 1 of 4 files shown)


frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -32,6 +32,7 @@
         vm.isChallengeHost = false;
         vm.stopLeaderboard = function() {};
         vm.stopFetchingSubmissions = function() {};
+        vm.currentDate = new Date().toISOString();
```
RishabhJain2018: @Ram81 Please use the time from the server as both front-end and back-end are decoupled.

Ram81: Sure @RishabhJain2018 will do it.

Ram81: @RishabhJain2018 is there an api already available to get server's current time ?

Ram81:
@RishabhJain2018 will it be fine if I create an api endpoint to get server time?

frontend/src/js/controllers/challengeCtrl.js
RishabhJain2018: No, there isn't any API for that.

RishabhJain2018:
Well, I suggest that we can modify the existing API to return the server's time along with other details. Please let me know if you have thoughts regarding this.

Ram81:
@RishabhJain2018 I can create one endpoint for returning time or if you want we can include it in some existing one, what do you suggest?

RishabhJain2018:
@Ram81 You can modify https://github.com/Cloud-CV/EvalAI/blob/master/apps/participants/views.py#L182 endpoint to get the server's time. Please let me know if you have more doubts.

Ram81:
@RishabhJain2018 What about the tests for this endpoint?

RishabhJain2018:
@Ram81 Can you please fix the travis builds ?

Ram81:
@RishabhJain2018 I don't get why the assert Statements are getting failed eventhough I have added the required fields in expected output.

RishabhJain2018:
@Ram81 Please check the format of the time in assert statements. I think you may be missing something there.

Ram81:
@RishabhJain2018 I have used timezone.now() in both i.e. test & api endpoint. Is it necessary for us to parse date as a string in expected output?

Ram81:
@RishabhJain2018 I made formats for both test & views.py I can't get why the build errors aren't getting resolved!

Ram81:
@RishabhJain2018 I fixed the test errors what about the coveralls?

RishabhJain2018:
@Ram81 yes, that is due to unavailability of tests for development urls. We will fix that in another issue.
Moreover, please break this into two PR's : 
1. For backend part only
2. for frontend part only
and referring the same issue. Please feel free to ask if something is not clear.

Ram81:
Sure @RishabhJain2018  :+1:

Ram81:
Closing this as I made separate PR's for both frontend & Backend.

PR #1273, opened 2017-09-23
Jobs: Remove unncessary db index in Submission model.

taranjeet:
Reasons for removing db index:

* url fields: Length is not fixed, Hence it does not make sense to create
index. A hash of url of fixed length if present is better candidate for
indexing

* execution time limit: No `where execution_time_limit = X` queries are
in use as of now. Also int can assume various possible values, thereby
slowing the index.

@deshraj @RishabhJain2018


apps/jobs/models.py
```diff
@@ -81,11 +81,11 @@ class Submission(TimeStampedModel):
         upload_to=RandomFileName("submission_files/submission_{id}"), null=True, blank=True)
     submission_metadata_file = models.FileField(
         upload_to=RandomFileName("submission_files/submission_{id}"), null=True, blank=True)
-    execution_time_limit = models.PositiveIntegerField(default=300, db_index=True)
-    method_name = models.CharField(max_length=1000, null=True, db_index=True)
```
RishabhJain2018: @deshraj @taranjeet I think we can create a hash of method name as in future it will be easy for  challenge host to filter the results on the basis of `method name`.

apps/jobs/models.py
```diff
@@ -81,11 +81,11 @@ class Submission(TimeStampedModel):
...
-    method_description = models.TextField(blank=True, null=True, db_index=True)
-    publication_url = models.CharField(max_length=1000, null=True, db_index=True)
-    project_url = models.CharField(max_length=1000, null=True, db_index=True)
+    execution_time_limit = models.PositiveIntegerField(default=300)
+    method_name = models.CharField(max_length=1000, null=True)
```
deshraj: @taranjeet: can you tell me that why we don't want to index the char fields. I agree that it doesn't make sense for the `TextField()`.

taranjeet: Because queries with where clause on `method_name` would not be there. This is just another info which we are storing (not for querying). 

`method_name` will very likely assume unique values, where I think it does not make sense to create index. Even if index is created, it does not guarantee that they will be used for querying purposes.

deshraj: No no, we will provide a feature where the user will be able to search for the submissions containing the particular substring of the method name. People could try different methods for training their models and if there are a lot of submissions and they want to look for the particular method, then we need the searching capability with `method_name` field. 

Please let me know if I am unclear somewhere.

taranjeet: Ok I got your point.  I have updated the code.

apps/jobs/models.py
taranjeet: @RishabhJain2018 : As of now I think that this is not required. We will see if any need pops up in the future.

RishabhJain2018: @taranjeet I was talking of the same thing as @deshraj was talking in the below comment :sweat_smile:

PR #1274, opened 2017-09-23
FIX #1272 : Changing Bower Dependencies ( fixes 1272 )

### LINKED ISSUE #1272: Frontend (angular-ui-router): State change events are deprecated
**Issue:** 
In `angular-ui-router<=1.x`,  State change events are deprecated, DISABLED and replaced by Transition Hooks. So redirection and state change will not work if lnstalled version of `angular-ui-router` is `>=1.x`. 
[More info](https://github.com/angular-ui/ui-router/wiki#state-change-events)

**How to Fix:**
* change bower dependencies `"angular-ui-router": "^0.3.1"` to  `"angular-ui-router": "0.3.1"` (without `^`) so that installation will be restricted to version `0.3.1`
or 
* if we need to go with latest version of angular-ui-router then we have to replace `$rootScope.$on('$stateChangeStart', function(){ ... })` with `$transitions.onStart({...})` and also have to update bower dependency for the same.

@aka-jain @RishabhJain2018 @deshraj

Ayukha:


RishabhJain2018:
@Ayukha Any updates on this?

RishabhJain2018:
@deshraj Closing this one as @Ayukha has been inactive on this thread.

PR #1275, opened 2017-09-23
Fix #1139: Disable the Challenge Phase buttons which are inactive while making submissions (Frontend-Fix).

### LINKED ISSUE #1139: Disable the Challenge Phase buttons which are inactive while making submissions

Ram81:
Fixes #1139


Ram81:
Done :+1:

RishabhJain2018:
@deshraj Please merge it.

Ram81:
@deshraj frontend fixes for #1139, this is with #1276.

PR #1269, opened 2017-09-21
Fix: Restrict the update profile API Endpoint from updating username.

isht3:
Fixes #864 

* Override the default django-rest-auth serializer by adding
  REST_AUTH_SERIALIZERS to settings and adding the location to
  the new serializer.
* Make a custom serializer in accounts app, and add the username field
  as a read_only_fields.

The test for the fix is being written. :)

PR #1281, opened 2017-09-26
Test for update profile API

xamfy:
See: #1269 
Test for checking whether username can be updated through API or not


isht3:
@xamfy Can you clean up the commit history? It should have only one commit, since it is a test for that particular feature. Ill test this on my system too!

RishabhJain2018:
Also, @xamfy Can you please fix `flake8` errors.

PR #1292, opened 2017-09-29
Fix runtime errors while running tests

### LINKED ISSUE #1277: RuntimeWarning: DateTimeField received a naive datetime
This error occurs while running the tests.
See https://stackoverflow.com/questions/18622007/runtimewarning-datetimefield-received-a-naive-datetime
image
Partial solution: change
```
when_made_public=datetime.datetime.now()
```
in tests/unit/jobs/test_views.py to 
```
when_made_public=timezone.now()
```

xamfy:
fixes #1277

### DIFF
--- a/tests/unit/jobs/test_views.py
+++ b/tests/unit/jobs/test_views.py
@@ -1,4 +1,3 @@
-import datetime
 import os
 import shutil
 
@@ -497,7 +496,7 @@ class ChangeSubmissionDataAndVisibilityTest(BaseAPITestClass):
             project_url="http://testserver/",
             publication_url="http://testserver/",
             is_public=True,
-            when_made_public=datetime.datetime.now()
+            when_made_public=timezone.now()
         )
 
         self.url = reverse_lazy('jobs:change_submission_data_and_visibility',


xamfy:
@RishabhJain2018 there are still 3 warnings. I can't find the solution for that.

RishabhJain2018:
@xamfy Can you please paste the screenshot for the warnings.

xamfy:
image

RishabhJain2018:
@xamfy The warnings are due to https://github.com/django/django/blob/1.7/django/db/models/fields/__init__.py#L1219.
Moreover, it is due to the fact that `auto_now_add` in django uses `datetime.datetime.now()`. For more details please refer here https://github.com/django/django/blob/d92b08536d873c0966e8192e64d8e8bd9de79ebe/django/db/models/fields/__init__.py#L1146

So, I think for now we can ignore these warnings. @deshraj What do you say?

deshraj:
No, we should not get these warnings and we should use `timezone.now()` instead of `datetime.datetime.now()`. 

@xamfy this looks good. I am merging this. Thanks for the fix.

PR #1296, opened 2017-09-30
Fix #1295: The file upload limit has been increased to 10 megabytes

### LINKED ISSUE #1295: Increase the data upload limit.
## Current Behaviour
By default django supports only 2.5 MB of the data to be uploaded on the web app. Refer [here](https://docs.djangoproject.com/en/1.10/ref/settings/#data-upload-max-memory-size) for more info.

## Effects
Due to the low upload limit the file in challenge creation using zip isn't being uploaded on the app as the size exceeds.

## Expected Behaviour
The upload limit must be increased to 10MB.

memclutter:
Fix #1295.

### DIFF
--- a/settings/common.py
+++ b/settings/common.py
@@ -276,5 +276,5 @@ RABBITMQ_PARAMETERS = {
 
 # The maximum size in bytes for request body
 # https://docs.djangoproject.com/en/1.10/ref/settings/#data-upload-max-memory-size
-FILE_UPLOAD_MAX_MEMORY_SIZE = 10485760 # 10 MB
-DATA_UPLOAD_MAX_MEMORY_SIZE = 10485760 # 10 MB
+FILE_UPLOAD_MAX_MEMORY_SIZE = 10485760  # 10 MB
+DATA_UPLOAD_MAX_MEMORY_SIZE = 10485760  # 10 MB
--- a/settings/common.py
+++ b/settings/common.py
@@ -276,5 +276,5 @@ RABBITMQ_PARAMETERS = {
 
 # The maximum size in bytes for request body
 # https://docs.djangoproject.com/en/1.10/ref/settings/#data-upload-max-memory-size
-FILE_UPLOAD_MAX_MEMORY_SIZE = 10485760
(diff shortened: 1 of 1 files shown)


settings/common.py
```diff
@@ -273,3 +273,8 @@
     },
     'SUBMISSION_QUEUE': 'submission_task_queue',
 }
+
+# The maximum size in bytes for request body
+# https://docs.djangoproject.com/en/1.10/ref/settings/#data-upload-max-memory-size
+FILE_UPLOAD_MAX_MEMORY_SIZE = 10485760
+DATA_UPLOAD_MAX_MEMORY_SIZE = 10485760
```
RishabhJain2018: Can you please write in comment the size in `MB` for future reference like `10485760 # 10MB`.

memclutter: ok, one moment

settings/common.py
```diff
@@ -273,3 +273,8 @@
     },
     'SUBMISSION_QUEUE': 'submission_task_queue',
 }
+
+# The maximum size in bytes for request body
+# https://docs.djangoproject.com/en/1.10/ref/settings/#data-upload-max-memory-size
+FILE_UPLOAD_MAX_MEMORY_SIZE = 10485760  # 10 MB
+DATA_UPLOAD_MAX_MEMORY_SIZE = 10485760  # 10 MB
```
RishabhJain2018: @memclutter Just a small change, please change the limit to `500 MB`.

RishabhJain2018:
@memclutter Any update on this? We need to merge it asap as it hinders the `challenge creation feature`.

settings/common.py
taranjeet: @RishabhJain2018 : 500 mb is too much? Also you will need to increase the limit at Nginx as well

@deshraj

RishabhJain2018: @taranjeet I know, but challenge file can be of `500 MB` as @deshraj suggested. We can increase the limit at nginx too.

deshraj:
Closing this one due to inactivity.

PR #1310, opened 2017-10-04
#1295: increase max file upload size to 500MB

### LINKED ISSUE #1295: Increase the data upload limit.

cwiggs:
Fixes #1295


RishabhJain2018:
@deshraj Please review and merge .

PR #1317, opened 2017-10-08
Fix #630: Resend email feature.

### LINKED ISSUE #630: Auth: Need email verification mail resend feature.
Why??
Suppose user loses verification email without verifying the link. Then neither user will verify his account nor create a new account with same email address. 

# Requirements
## Backend 
- [ ] API endpoint to resend verification email (User token based)
@deshraj : Please take care of it.

## Frontend 
I think there should be button/Link ("Resend Verification Email" ) inside rounded (dark red) portion of following screenshot. 
screenshot from 2017-02-05 11 39 21

@aka-jain : Need your thoughts about it.

isht3:
Fixes #630. 

Before hovering with mouse.

After hovering with mouse.

After it's clicked.

### DIFF
--- a/apps/accounts/views.py
+++ b/apps/accounts/views.py
@@ -6,8 +6,11 @@ from rest_framework.response import Response
 from rest_framework import permissions, status
 from rest_framework.decorators import (api_view,
                                        authentication_classes,
-                                       permission_classes,)
+                                       permission_classes,
+                                       throttle_classes,)
 from rest_framework_expiring_authtoken.authentication import (ExpiringTokenAuthentication,)
+
+from .throttles import ResendEmailThrottle
 
 
 @api_view(['POST'])
@@ -21,6 +24,7 @@ def disable_user(request):
     return Response(status=status.HTTP_200_OK)
 
 
+@throttle_classes([ResendEmailThrottle])
 @api_view(['POST'])
 @permission_classes((permissions.IsAuthenticated,))
 @authentication_classes((ExpiringTokenAuthentication,))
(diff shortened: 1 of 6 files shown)


apps/accounts/views.py
```diff
@@ -12,9 +14,17 @@
...
     user.is_active = False
     user.save()
     logout(request)
     return Response(status=status.HTTP_200_OK)
+
+
+@api_view(['POST'])
+@permission_classes((permissions.IsAuthenticated,))
+@authentication_classes((ExpiringTokenAuthentication,))
+def resend_email(request):
```
deshraj: Can we add a decorator to throttle this request? it would be nice to have a throttling of 3 requests per minute so that no one can abuse the system.

isht3: Okay, sounds good! :+1:

isht3: @deshraj For that I'll have to create a custom Throttle Class, yes?

isht3: @deshraj For that a separate Throttle class would have to be written, right?

deshraj: Yes.

isht3: @deshraj Okay, cool! :smile:

isht3:
@deshraj I've implemented the throttles!

RishabhJain2018:
@isht3 can you please break this PR into 2 different PR's with frontend and backend. I will review it by tomorrow :)

isht3:
@RishabhJain2018 Sure! On it.

PR #1320, opened 2017-10-11
Fix #1286 Docker Compose works well now

### LINKED ISSUE #1286: Composing Dockerfiles not properly working
Hey, as requested on the gitter, I open an issue regarding the Dockerfile.
The expected behavior is to run the docker compose in order to have a full EvalAI ready.

For the observed behavior,various issues happens, mostly the nodejs image where sass can't be installed.
I know that switching the docker to an ubuntu16:04 image solve the issue.
However, the compose still does  not work, and especially the submisson working exit immediatly.

To reproduce the  issue, just try the docker-compose module as it is.

varunagrawal:
Made the following changes in order to take care of #1286 :

- Set the SASS gem to a version compatible with Ruby 1.9.1 (default for Ubuntu 14.04)
- Added the libfontconfig dependency for PhantomJS
- The container also adds the `karma.conf.js` file.
- The volume for the `nodejs` container in the compose file is removed as it was causing overwrites.


deshraj:
The changes look good to me. I will test it on my local system sometime tonight.

deshraj:
This works flawlessly. Thanks again @varunagrawal for fixing this. :)

PR #1276, opened 2017-09-23
Fix #1139: Disable the Challenge Phase buttons which are inactive while making submissions (Backend Fix).

Ram81:
Fixes #1139

PR #1326, opened 2017-10-16
Fix: Added check for removing Stderr File for Sucessful Submission(#…

### LINKED ISSUE #1175: Remove Stderr file for the successful submissions
## Observed behavior

Currently when there is a successful submission, then also a `stderr_file` is created for that submission which leads to the creation of a blank file that is of no use and adds a small overhead in processing the successful submissions.

## Expected behavior

When there is a successful submission, then `stderr_file` field should be set None instead of creating a blank file.

Ram81:
Fixes #1175.
@deshraj Please Review. 
temp


scripts/workers/submission_worker.py
```diff
@@ -398,9 +398,10 @@ def run_submission(challenge_id, challenge_phase, submission_id, submission, use
     with open(stdout_file, 'r') as stdout:
         stdout_content = stdout.read()
         submission.stdout_file.save('stdout.txt', ContentFile(stdout_content))
-    with open(stderr_file, 'r') as stderr:
-        stderr_content = stderr.read()
-        submission.stderr_file.save('stderr.txt', ContentFile(stderr_content))
+    if (submission_status is not Submission.FINISHED):
```
deshraj: I think the check should be 

`if (submission_status is not Submission.FAILED):`

Please correct me if I am wrong.

Ram81: @deshraj if the submission is failed then you want an Stderr file to be generated right?

Ram81: @deshraj you want an Stderr file to be generated in case of submission Failure, right?

deshraj: Yeah.

Ram81: And @deshraj 
if (submission_status is not Submission.FAILED):
this check will save an stderr file for every successful submission, right!

Ram81:
@deshraj I updated the check condition, it is less ambiguous now. Please review

deshraj:
Looks good. Merging it.

deshraj:
Thanks @Ram81 for fixing this. :)

Ram81:
@deshraj can you review #1276, It's pending from few days :smile:

PR #1327, opened 2017-10-16
Fixes Dataset split with that codename already exists.

### LINKED ISSUE #1294: 'Dataset split with that codename already exists' error
## Current Scenario

When a user creates a challenge using a zip configuration that has some dataset split codename, then the challenge is created successfully. If he tries to create a similar challenge with similar dataset split name then he gets an error that `Dataset split with that codename already exists` which should not happen since different challenges can have same dataset split codename. 

## Deliverables

- [ ] Remove `unique=True` from DatasetSplit's codename field
- [ ] Set unique_together `pk` and `dataset_split` in Challenge phase model
- [ ] Don't forget to name migrations

isht3:
* Remove unique=True from DatasetSplit's codename field
* Set unique_together pk and dataset_split in Challenge phase model
* Add migrations for the changes.

Fixes #1294.


deshraj:
@isht3 I dont think the change  
- Set unique_together pk and dataset_split in Challenge phase model

is incorporated in this PR.

apps/challenges/models.py
```diff
@@ -124,7 +124,7 @@ def __init__(self, *args, **kwargs):
     class Meta:
         app_label = 'challenges'
         db_table = 'challenge_phase'
-        unique_together = (('codename', 'challenge'),)
+        unique_together = (('codename', 'challenge'), ('id', 'codename'),)
```
deshraj: We don't need this change.

isht3:
@deshraj That's where I was confused, I tried `('pk', 'codename')` where `pk` is the primary key of the ChallengePhase model (I think), then I got this `challenges.ChallengePhase: (models.E012) 'unique_together' refers to the non-existent field 'pk'.` Then I thought since `pk` and `id` are similar, the changes would be the same.

apps/challenges/models.py
deshraj: So, we need the additional `unique_together` here with attributes `pk` and `dataset_split`

deshraj:
To clear your doubt: in the models, `pk` and `id` are the same. It should work if you are using `pk` instead of `id`. You might be making some mistake somewhere.

isht3:
@deshraj Okay, Ill look into what's wrong.

isht3:
@deshraj It's a ManyToMany field. What we are trying to achieve is to have the unique DatasetSplit name only within a single challenge, yes?

isht3:
@deshraj I don't think it's a error on my side, because the travis tests are also failing on the `pk` field. I think this answer discusses a potential solution https://stackoverflow.com/questions/4870935/django-enforcing-manytomanyfield-unique-items Should I look into it?

deshraj:
@isht3 can you try replacing `pk` -> `id` here (https://github.com/Cloud-CV/EvalAI/pull/1327/files#diff-06139830cb2f143920edd3015bb7b1baR127) and see if that works.

isht3:
@deshraj `ProgrammingError: column "dataset_split" named in key does not exist` was raised, should I look into other solutions?

PR #1344, opened 2017-10-29
Fix #1338: uses main-header instead of sim-header.

### LINKED ISSUE #1338: Navigation menu is not coming on mobile after login.
Navigation menu icon is not coming up on mobile for logged in user.

AmosHason:
#1338

### DIFF
--- a/frontend/src/views/web/partials/main-header.html
+++ b/frontend/src/views/web/partials/main-header.html
@@ -27,11 +27,12 @@
                     </div>
                 </li>
                 <li><a class="waves-effect waves-dark" ui-sref="home" ui-sref-active="active"><i class="fa fa-home"></i>Home</a></li>
-                <li ng-if="!isAuth"><a class="waves-effect waves-dark main-header-link" ui-sref="auth.signup" ui-sref-active="active"><i class="fa fa-sign-out"></i>Sign Up</a></li>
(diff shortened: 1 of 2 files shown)


deshraj:
@AmosHason can you please paste screenshot here?

AmosHason:
@deshraj I will paste later when I get back home.

AmosHason:
Desktop:

image

Mobile:

image

image

New sign up icon (it was logout icon before):

image

By the way, I think that the line in bold in the menu should reflect the current page (currently "Home" is always the line in bold). I'm not sure how exactly it's intended to be.

AmosHason:
Also there is an issue with the desktop view of main-header that should be addressed afterwards:

Log in -> Push "Hi" button -> Push "Dashboard" button -> Push "Hi" button (**view does not change**) -> Push "Hi" button again (view changes)

aka-jain:
@AmosHason don't change the header instead implement that menu icon in new header.

AmosHason:
@aka-jain Are you sure? It's inconsistent this way. It doesn't make sense to not put the logout button where the login button is located.

aka-jain:
@AmosHason Hi, sorry for the late response. 
Our design is settled like that we have to use two types of header. One is for landing page and another one is for dashboard. You the logged in user will have different experience as compared to non logged in user. Hope you understand.
So, please do the required changes. Thanks.

aka-jain:
Closing this PR as there is no response and there is no need of changes that has been done in this PR 
cc @RishabhJain2018 @deshraj

PR #1349, opened 2017-10-30
Define unicode in Python 3 

cclauss:
Fixes the undefined names in https://travis-ci.org/Cloud-CV/EvalAI/jobs/300525447#L643-L651  Unicode was dropped for Python 3 because all strs are Unicode.


tests/unit/accounts/test_urls.py
```diff
@@ -5,6 +5,11 @@
 
 from rest_framework.test import APITestCase, APIClient
 
+try:
+    unicode        # Python 2
+except NameError:
+    unicode = str  # Python 3
+
```
deshraj: @cclauss instead of declaring it everywhere, it would make sense to declare it once in `tests/unit/__init__.py`. Does that sound good?

cclauss: That does not work.  😏

PR #1354, opened 2017-10-31
Fix #630: Add frontend for resend E-mail feature.

isht3:
Fixes #630. 

Before hovering with mouse.

After hovering with mouse.

After it's clicked.

PR #1355, opened 2017-10-31
Fix #630: Add backend for resend E-mail feature.

### LINKED ISSUE #630: Auth: Need email verification mail resend feature.

isht3:
Fixes: #630 by adding backend API endpoints and throttles.


settings/common.py
```diff
@@ -158,11 +158,13 @@
...
         'rest_framework.throttling.AnonRateThrottle',
-        'rest_framework.throttling.UserRateThrottle'
+        'rest_framework.throttling.UserRateThrottle',
+        'accounts.throttles.ResendEmailThrottle',
     ),
     'DEFAULT_THROTTLE_RATES': {
         'anon': '100/minute',
-        'user': '100/minute'
+        'user': '100/minute',
+        'resend_email': '2/minute',
```
deshraj: Can we make it 3 per hour?

deshraj:
@isht3 this looks good but it would be nice if you can add unit test for this? See https://stackoverflow.com/questions/39309046/what-is-the-proper-way-of-testing-throttling-in-drf. It's really easy.

settings/common.py
isht3: @deshraj Okay, cool!

isht3:
@deshraj On it. Thanks for pointing it out!

isht3:
@deshraj For some reason the throttles aren't throttling requests after 3, that's why the Travis tests are failing, I'll look into what's wrong.

deshraj:
I think there are some imports that are not being used. That's why the travis checks are failing. See https://travis-ci.org/Cloud-CV/EvalAI/jobs/300185433#L1356

isht3:
@deshraj I think there's a problem, I checked in Postman by sending throttling requests to all the scopes and I don't think any of the throttles are working. Using Postman, I set `UserRateThrottle` to `10/minute` and tried to get my requests throttled and it didn't work allowing all the requests, but it was working for `resend_email` and `user` when I set it to `0/minute`, it throttled the first request itself. Does it have something to do with the development environment or is it an issue?

deshraj:
Throttling seems to work really well in the production environment. This seems strange if it is not working on your system. @RishabhJain2018 can you please check the throttling request on your machine and see if you can reproduce the same problem which @isht3 is facing. I am a bit occupied with some other work and will be able to test this only after 15th.

RishabhJain2018:
@deshraj Sure. I will update asap.

deshraj:
@RishabhJain2018 please test this whenever you get time.

aka-jain:
@RishabhJain2018 please review this once.

RishabhJain2018:
@deshraj Yes, I am able to produce the same error as faced by @isht3 . I am trying to fix it. I will update on it asap.

isht3:
@RishabhJain2018 Should this be opened as a issue?

RishabhJain2018:
@isht3 Can you please help us in figuring out why the throttle is not working. Till then I am keeping your both the pull requests(#1355 & #1354 ) on hold.

isht3:
@RishabhJain2018 Okay, I will try to figure out why throttles aren't working.

isht3:
@RishabhJain2018 How do I figure out if throttling is working in production?

apps/accounts/throttles.py
```diff
@@ -0,0 +1,20 @@
...
+class ResendEmailThrottle(SimpleRateThrottle):
+    """
+    Used to limit the requests to /accounts/user/resend-email
+    to 2/minute.
+    """
+    scope = 'resend_email'
+
+    def get_cache_key(self, request, view):
+        if request.user.is_authenticated:
+            ident = request.user.pk
```
deshraj: Can you please change it to `user_id`?

deshraj:
@isht3: We tried to test it on our local machine and it looks like throttling is not working. We are looking into it. Please let us know if you can find a solution for the same. The sooner we are able to fix the throttling issue, the sooner we will be able to merge your PR. :)

isht3:
@deshraj I'm also looking into this. I've also asked SO if it speeds things up. https://stackoverflow.com/questions/48473621/django-rest-framework-throttling-not-working-in-development-environment

isht3:
@RishabhJain2018 @deshraj I think this throttling error is due to the usage of `DummyCache` in the settings. I'll test more to confirm.

isht3:
@deshraj @RishabhJain2018 I think it is because of the `DummyCache` the throttling isn't working. I found one more issue where the user had a similar problem https://stackoverflow.com/questions/33830283/django-rest-framework-caching-bug I don't think this solution will work for us. Any other suggestions?

deshraj:
Good catch! You can remove the tests for the throttling and then we can merge this pull request. Thanks!

RishabhJain2018:
@deshraj  I have figured the problem for throttling and why it isn't working.It is due to `DummyCache` in development environment. For the production server, we can have to install `sudo apt-get install memcached` and `pip install python-memcached`. I think currently we can also use the same backend in our development as we aren't caching anything heavy. @deshraj What do you say?
 Also, I suggest @isht3 to change the cache backend and install the above packages and then re-write the tests for throttling and now they should work. please let me know if you have more doubts.

isht3:
@RishabhJain2018 I've updated the requirements and the settings. I'll write the tests in a few hours.

isht3:
@RishabhJain2018 @deshraj I think this is done, could you go through it?

apps/accounts/views.py
```diff
@@ -1,20 +1,34 @@
...
     user.save()
     logout(request)
     return Response(status=status.HTTP_200_OK)
+
+
+@throttle_classes([ResendEmailThrottle])
+@api_view(['POST'])
+@permission_classes((permissions.IsAuthenticated,))
+@authentication_classes((ExpiringTokenAuthentication,))
+def resend_email(request):
```
deshraj: Can you please add docstring to this API? Also, please rename this to `resend_email_confirmation`

isht3:
@deshraj Done!

PR #1366, opened 2017-11-03
Analytics: Optimize queries to get Submission count

deshraj:
Improves the response time from 10 seconds to 540 ms.


deshraj:
Edit: Optimized more and now the response time is reduced to 120 ms and 7 queries.

PR #1385, opened 2017-11-30
Changed to Rounded Buttons

gautamjajoo:
Changed the Rectangular buttons to Rounded buttons by adding "grad-btn" in the button tag.
Thanks!


gautamjajoo:
Ok. Will keep these things in mind! Thanks..

PR #1409, opened 2017-12-12
Fix minor grammatical errors

mrMarce:


aka-jain:
@RishabhJain2018 I think we should merge this. Thoughts?

RishabhJain2018:
@aka-jain Yes sure ! It looks good to me as well.

PR #1410, opened 2017-12-14
Compressed Images

weakit:
Loading the website should be a bit faster now.


aka-jain:
Looks good to me. Merging it.
cc @deshraj @RishabhJain2018

PR #1413, opened 2017-12-16
show meaninful error messsage when email verification not done

16116073:
Fix #1413 : Now a meaningful error message is shown to inform the user about the email verification link in order to activate his account and able to proceed ahead.
pr

### DIFF
--- a/requirements/common.txt
+++ b/requirements/common.txt
@@ -12,7 +12,7 @@ drfdocs==0.0.11
 pika==0.10.0
 pickleshare==0.7.4
 Pillow==3.4.2
-psycopg2==2.6.2
+psycopg2==2.7.3.1
 PyYaml==3.12
 proc==0.10.1
 rstr==2.2.5


requirements/common.txt
```diff
@@ -12,7 +12,7 @@ drfdocs==0.0.11
 pika==0.10.0
 pickleshare==0.7.4
 Pillow==3.4.2
-psycopg2==2.6.2
```
RishabhJain2018: @16116073 The screenshot you pasted isn't in synchronization with this code change. Also, please revert this change as it is not the part of this pull request.

16116073: i faced this issue while installing, 
https://github.com/psycopg/psycopg2/issues/594
After upgrading my postgres version and updating my requirements/common.txt  file made it work for me.

RishabhJain2018: Okay. So, I think this should be the part of the another pull request and this pull request should contain the changes relevant to this issue. Please make this change in another pull request as need to maintain the modularity in the code.

PR #1414, opened 2017-12-16
Fix #1346: Backend for sending email to invite users to participant team using URL.

### LINKED ISSUE #1346: Add feature to invite users to participant team using URL
## Deliverables

Add a feature to invite new team member to a participant team using a Unique URL. Also, don't forget to write unit tests. :) 

## Steps

1. **user1** enters the email of his teammate (let's say **user2**)
2. Send an email to **user2** (if he/she is registered on EvalAI) with a unique URL
3. **user2** clicks on the URL to accept the invitation to the team.
4. *user1* gets an email saying that "**user2** has accepted your invite to team <blah>" (we can iterate over the email message later) 

### Some important points to note

- [ ] Checks performed before inviting to the team should be similar to current invitation logic

isht3:
* Define URL End-Points.

    * Define unique string encoder and decoder.

    * Set up a simple server to send E-mail.
   
    * Edit failing tests and add new tests for features.

The settings for sending email is not configured yet.
#1346


RishabhJain2018:
@isht3 Please fix the travis build.

isht3:
@RishabhJain2018 The backend tests are failing because of the feature change, I didn't get time to change and add the tests. I just got a new rig, I'll configure it and do the tests by this weekend. 😄

RishabhJain2018:
@isht3 No problem. Take your time ! :-) Please let me know when you are done with it.

isht3:
@RishabhJain2018  @deshraj A small problem I faced while writing the tests was since after sending an invite, it currently shows `'The user was sent an invite.`  to the user and prints the email in the terminal (I can't access the email while testing). So to write tests for this feature I would have to add the email and the team name to the response message, something like  `'user@domain.com was sent an invite to join host team.`.  Only then I could get the team and email during testing and check if it actually gives the desired output. Should I do it?

isht3:
@RishabhJain2018 I believe this is completed. :smile:

deshraj:
@isht3: I have left some comments on the Pull request for inviting the host to host teams. So, can you please do the similar changes here?

Also, it seems like `encode_data()` and `decode_data()` are same in both the pull requests. So can you please put them in `apps/base/utils.py` and then import from there? Does that sound good? Please see if there is some other duplicate code that we can avoid, then that would be great. I might be missing something.

isht3:
@deshraj Yes, that was better! I did the changes, except the template part. Could you merge #1528 first?

RishabhJain2018:
Closing this due to inactivity.

PR #1420, opened 2017-12-19
Update login.html - change forgot password -> forgot password or username?

trexdinosaur:
Username is already included in the email, so it lets the user know that the username can be checked from there.


aka-jain:
@RishabhJain2018 Is this ready to merge?

aka-jain:
Merging this. cc @RishabhJain2018

PR #1426, opened 2017-12-27
Fix #1424 : Password strength indication

### LINKED ISSUE #1424: Password strength indication
When a user fills SIGN UP form, 
- Observed behaviour
after the clicking on 'sign up', user is informed about the password being common or not secure.

- Expected Behaviour
there should be an indicator telling the user about his password strength as he types each character.

Moreover, we do not have any specifications for password except that it should consist of 8 characters,
for making it better we could impose certain conditions.

Would You Like to Take this Issue?
YES !

Thanks!

Obtin:
@aka-jain Please review the PR.

### DIFF
--- a/frontend/src/js/controllers/authCtrl.js
+++ b/frontend/src/js/controllers/authCtrl.js
@@ -176,6 +176,63 @@
             }
         };
 
+
+      //Function To check password Strength
+        vm.checkStrngth = function(password) {
+            //TextBox left blank.  
+            var password_strength = document.getElementById("#password_strength");
+            if (pswd == 0) {
+                $("#password_strength").html("");
+                return;
+            }
+            //Regular Expressions.  
+            var regex = new Array();
+            regex.push("[A-Z]"); //Uppercase Alphabet.  
+            regex.push("[a-z]"); //Lowercase Alphabet.  
+            regex.push("[0-9]"); //Digit.  
+            regex.push("[$$!%*#?&]"); //Special Character.  
+            var passed = 0;
+            //Validate for each Regular Expression.  
+            for (var i = 0; i < regex.length; i++) {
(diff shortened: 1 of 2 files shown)


trexdinosaur:
@Obtin I think for passwords it might not be the best idea to test if it has a specific set of characters. 

Even though it might be a bit of performance hit (a small one), libraries like https://github.com/dropbox/zxcvbn offer a bit more advanced algorithms to determine whether a password is truly secure by checking against commonly used passwords and repetitive characters.

Obtin:
@trexdinosaur I agree it is indeed a good idea. Let's have a final say from mentors. .:-)
@aka-jain

frontend/src/js/controllers/authCtrl.js
```diff
@@ -176,6 +176,63 @@
...
+
+      //Function To check password Strength
+        vm.checkStrngth = function(password) {
+            //TextBox left blank.  
+            var password_strength = document.getElementById("#password_strength");
+            if (pswd == 0) {
+                $("#password_strength").html("");
+                return;
+            }
+            //Regular Expressions.
```
aka-jain: Create a service for this expression check. @Obtin

aka-jain:
@Obtin @trexdinosaur Agree with you guys but as we are not using password strength much in our app then it would be good too create our own service (a new task :D).
@Obtin try to create a service that will check the password strength. Also add the screenshot of your final result.

aka-jain:
Any update @Obtin

Obtin:
@aka-jain trying to fix it.

Obtin:
I had some problem with my git as a result of which I had to setup my project again. I have created a new pull request in this regard updated with the requested changes. Thank You.

PR #1445, opened 2018-01-06
Issue#1424: Password strength indication using service

Obtin:
#1424 
@deshraj @RishabhJain2018 @aka-jain updated the pr with required changes. Thank you.


Obtin:
The screenshots of the requested changes . :-)
pswd1
pswd2

aka-jain:
Remove that regex syntax from Password indication.  @Obtin

frontend/src/views/web/auth/signup.html
```diff
@@ -30,9 +30,12 @@
         </div>
         <!-- password -->
         <div class="input-field align-left">
-            <input type="password" id="password" onpaste="return false;" class="dark-autofill" name="password" ng-model="auth.regUser.password" ng-minlength="8" required>
+            <input type="password" id="password" onpaste="return false;" class="dark-autofill" name="password" ng-model="auth.regUser.password" ng-minlength="8" ng-change= "auth.checkStrength(password)" required>
             <span class="form-icon form-icon-dark"><i class="fa fa-key"></i></span>
-            <label for="password">Password (min 8 characters) *</label>
+            <label for="password">
+                <strong>Password(minlength:8,required:*[a-z],*[A-Z],*[0-9],*[!@#\$%\^&\*])</strong>
```
aka-jain: Replace this with Password(minlength:8,required:*[a-z],*[A-Z],*[0-9],*[!@#\$%\^&\*]) with Password - minimum length:8 required.

Obtin: @aka-jain it is to inform the user to include these in his password. Do you still want me to remove it?

frontend/src/views/web/auth/signup.html
```diff
@@ -30,9 +30,12 @@
...
+                <span id ="pswdStrength"></span>
```
aka-jain: For id/class don't use camelcase. Replace it with `pswd-strength-message`

frontend/src/js/controllers/authCtrl.js
```diff
@@ -176,6 +176,14 @@
             }
         };
 
+        vm.checkStrength = function(password){
+            //TextBox left blank. 
+            var password_strength = utilities.check(password) ;
+            var pswdStrength = document.getElementById("pswdStrength");
```
aka-jain: Change name of id here to `pswd-strength-message`

Obtin:
@aka-jain please review the PR , the changes have been done. Thank you

frontend/src/js/controllers/authCtrl.js
```diff
@@ -176,6 +176,14 @@
             }
         };
 
+        vm.checkStrength = function(password){
+            //TextBox left blank. 
+            var password_strength = utilities.check(password) ;
+            var pswd-strength-message = document.getElementById("pswd-strength-message");
```
aka-jain: @Obtin Not this way, use camel case in JS (vvariable declaration) adnd dash format in css classes and id's. I hope you get it. Do these changes please. Also, use angular element selector to get element instead of using `document.getElementById`.

spyshiv:
@Obtin any update on this PR ?

frontend/src/js/services/services.js
```diff
@@ -104,6 +104,53 @@
             angular.element("#sim-loader").fadeOut();
             angular.element(".web-container").removeClass('low-screen');
         };
+
+        this.check = function(password){
+
+           //Regular Expressions.  
+            var regex = new Array();
+            regex.push("[A-Z]"); //Uppercase Alphabet.
```
spyshiv: There is no need to write 4 statements to do similar operation. You can push multiple items in a array in single statement i.e. `array.push(item1, item2, item3 ......)`

frontend/src/js/services/services.js
```diff
@@ -104,6 +104,53 @@
...
+
+           //Regular Expressions.  
+            var regex = new Array();
+            regex.push("[A-Z]"); //Uppercase Alphabet.  
+            regex.push("[a-z]"); //Lowercase Alphabet.  
+            regex.push("[0-9]"); //Digit.  
+            regex.push("[$$!%*#?&]"); //Special Character.  
+            var passed = 0;
+            //Validate for each Regular Expression.  
+            for (var i = 0; i < regex.length; i++) {
```
spyshiv: Please replace `regex.length` with `4` (Integer) since length of array is fixed and equal to 4.

frontend/src/js/services/services.js
```diff
@@ -104,6 +104,53 @@
...
+                if (new RegExp(regex[i]).test(password)) {
+                    passed++;
+                }
+            }
+            //Validate for length of Password.  
+            if (passed > 2 && password.length > 5) {
```
spyshiv: In signup form minimum length of password is `8` so please change `5` with `8`.

frontend/src/js/services/services.js
aka-jain: @spyshiv I think it's good to have dynamic length here as maybe in future if we push another item in regex then we again have to change the length. What do you think?

frontend/src/js/services/services.js
```diff
@@ -104,6 +104,53 @@
...
+                passed++;
+            }
+            //Display status.  
+            var color = "";
+            var strength = "";
+            var status = false;
+            if (passed == 1) {
+                strength = "Weak";
+                color = "red";
+                var status = false;
```
spyshiv: Remove the statement `var status = false;` that because `status` is already defined with same value at Line 130.

frontend/src/js/services/services.js
```diff
@@ -104,6 +104,53 @@
...
+            var strength = "";
+            var status = false;
+            if (passed == 1) {
+                strength = "Weak";
+                color = "red";
+                var status = false; 
+            } else if (passed == 2) {
+                strength = "Average";
+                color = "darkorange";
+                var status = false;
```
spyshiv: Remove the statement var status = false; that because status is already defined with same value at Line 130.

frontend/src/js/services/services.js
```diff
@@ -104,6 +104,53 @@
...
+            } else if (passed == 3) {
+                strength = "Good";
+                color = "green";
+                var status = true;
```
spyshiv: Replace `var status = true;` with `status = true;`. No need to use `var` again and again.

frontend/src/js/services/services.js
```diff
@@ -104,6 +104,53 @@
...
+            } else if (passed == 4) {
+                strength = "Strong";
+                color = "darkgreen";
+                var status = true;
```
spyshiv: Replace `var status = true;` with `status = true;`. No need to use `var` again and again.

frontend/src/js/services/services.js
```diff
@@ -104,6 +104,53 @@
...
+            } else if (passed == 5) {
+                strength = "Very Strong";
+                color = "darkgreen";
+                var status = true;
```
spyshiv: Replace `var status = true;` with `status = true;`. No need to use `var` again and again.

frontend/src/js/services/services.js
spyshiv: @aka-jain: I agree with your point but what else can be used as password character apart from `[A-Z]`, `[a-z]`, `[0-9]`, `[$$!%*#?&]`.

aka-jain: Take in this way..  what if we remove any item from rejex 🤔

spyshiv: Ok then `regex.length` is better choice.

Obtin:
@aka-jain @spyshiv dealing with exams right now, will try give you do all the suggested changes soon.Thankyou.

frontend/src/js/services/services.js
Obtin: then I guess we must remove ng-minlength="8"

Obtin:
@aka-jain @spyshiv I have done all the minor changes, 

> document.getElementById
to the angular selector as I am struggling with it :-p, Will be done soon.

Obtin:
@aka-jain @spyshiv, All the requested changes have been done. Please review it. Thank you.

Obtin:
@aka-jain @spyshiv @RishabhJain2018 I guess all the changes are now made , please review it and let me know if it needs any other thing to fix it. Thank you.

frontend/src/js/controllers/authCtrl.js
```diff
@@ -176,6 +176,18 @@
             }
         };
 
+        // variable required for checkStrength
+        vm.color={};
+
+        // function to check password strength
+        vm.checkStrength = function(password){
+            //TextBox left blank. 
+            var password_strength = utilities.check(password) ;
```
deshraj: Please use camel casing here instead of following the underscore code styling.

frontend/src/js/controllers/authCtrl.js
```diff
@@ -176,6 +176,18 @@
...
+            var pswdStrength =angular.element(document.querySelector('#pswd-strength-message'));
```
deshraj: Can you please replace `pswdStrength ` with `passwordStrength`?

Obtin:
@deshraj done

PR #1449, opened 2018-01-08
Fix #1430: Combined frontend backend servers

### LINKED ISSUE #1430: Combine frontend and backend servers
@RishabhJain2018, @aka-jain  
Currently, the frontend and the backend are hosted on different servers.
But the frontend can be hosted on the backend server.

viditjain08:
- STATIC_URL and STATIC_DIRS in dev.py changed to avoid single and double server conflict

- gulp dev:syncserver added- It watches the frontend and serves the required changes to the backend's static folder (dist).

- Home url added- In the backend, the home template of the frontend has been added, hence, adding all the frontend urls, since they are used from /dist/

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -52,11 +52,11 @@ gulp.task('vendorjs', function() {
         gulp.src(paths)
             .pipe(concat(chunkName + '.js'))
             //.on('error', swallowError)
-            .pipe(gulp.dest("frontend/dist/vendors"))
+            .pipe(gulp.dest("dist/vendors"))
             .pipe(gulp.watch(paths).on('change', function(event) {
                 if (event.type == 'deleted') {
                     var filePathFromSrc = path.relative(path.resolve(paths), event.path);
-                    var destFilePath = path.resolve('frontend/dist/vendors', filePathFromSrc);
+                    var destFilePath = path.resolve('dist/vendors', filePathFromSrc);
                     del.sync(destFilePath);
                 }
             }))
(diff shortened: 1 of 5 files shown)


aka-jain:
Please add description. @viditjain08

gulpfile.js
```diff
@@ -410,3 +409,8 @@ gulp.task('prod', function(callback) {
 gulp.task('dev:runserver', function(callback) {
     runSequence('dev', 'connect', 'watch', 'test:watch', callback);
 });
+// Syncserver for development in backend
+gulp.task('dev:syncserver', function(callback) {
```
aka-jain: @viditjain08 What's the difference between dev:runserver and dev:syncserver.. both will do the same thing as you have changed the source directory. (except server running)

viditjain08: Yes, but I thought it would be better if there are no servers running (frontend + phantomjs) when thr person wants to use only a single server, plus in heroku deployment, we wont be able to run multiple servers.

aka-jain: So my question is that how will I start dev server then if I developing locally. It will also sync Backend server. Isn't it.
You should write different tasks for the syncserver. Correct me if I'm missing something please @viditjain08

viditjain08: Different tasks can be written, but it would be pointless to maintain 2 directories (one for backend and other for frontend) since previously, there was only 1 which was watched, so indirectly, it was also synced with backend except for the fact that they were on different servers.
But if you want I can write different tasks for the syncserver.

aka-jain: Read this, maybe it will help you to understand why we have accepted Frontend driven approach. https://developer.telerik.com/featured/front-end-driven-applications-new-approach-applications/ @viditjain08

viditjain08: I have added a commit which maintains the frontend directory as before when gulp dev:runserver, and a different directory with syncserver. 
But I dont know why travis ci is not able to use the cache while checking.
What may be the problem?
@aka-jain

aka-jain: Looks Good, thanks.

aka-jain:
@deshraj @spyshiv I think this approach is good to create a single server for frontend and backend. But also we are going to use this approach only for Heroku deployment, otherwise it will create some confusions between contributors.
@viditjain08

aka-jain:
Also @deshraj @spyshiv @RishabhJain2018 please check this PR once locally in your machine too.

deshraj:
@viditjain08: Thanks for all the work on this feature. But since now we have started the porting the frontend on Angular 5, so we don't think we would need this feature anymore. We look forward to your cotnributions for Heroku Deployment for the individual projects.

PR #1485, opened 2018-01-21
Fix #1458: Error not displayed issue

### LINKED ISSUE #1458: No error is shown if a team is not selected.
A red bar is visible on the top right corner when trying participate in a challenge without any team selected.

screenshot from 2018-01-10 21-31-38

I would like to take this issue.

codervivek:
This will fix issue #1458 .

### DIFF
--- a/frontend/src/js/controllers/challengeCtrl.js
+++ b/frontend/src/js/controllers/challengeCtrl.js
@@ -156,10 +156,10 @@
                                                         vm.stopLoader();
                                                     },
                                                     onError: function(response) {
-                                                        if (response.data['detail']) {
+                                                        if (response.status == '404') {
                                                             var error = "Please select a team first!";
-                                                        } else if (response.data['error']) {
-                                                            error = response.data['error'];
+                                                        } else {
+                                                            error = "Server error";
                                                         }
                                                         $rootScope.notify("error", error);
                                                         vm.stopLoader();


frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -156,10 +156,10 @@
                                                         vm.stopLoader();
                                                     },
                                                     onError: function(response) {
-                                                        if (response.data['detail']) {
+                                                        if (response.status == '404') {
                                                             var error = "Please select a team first!";
-                                                        } else if (response.data['error']) {
-                                                            error = response.data['error'];
+                                                        } else {
+                                                            error = "Server error";
```
deshraj: @codervivek: I am curious to know that what is the value of response.status in the else condition. Can you please paste the value here?

codervivek: @deshraj Actually response.data returns a HTML string of Django page. Since response.data['detail'] didn't had any value, I checked for response.status. Since onError will only execute when there some error so I have only checked for 404 or else there I have classified them as "server error". Please suggest if you want to add some other case.

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -156,10 +156,10 @@
                                                         vm.stopLoader();
                                                     },
                                                     onError: function(response) {
-                                                        if (response.data['detail']) {
+                                                        if (response.status == '404') {
```
RishabhJain2018: @codervivek There is no need to check for string as `response.status` is a number. You can simply use `if (response.status === 404)`

PR #1538, opened 2018-02-22
Add support for changing Participant Team Name

### LINKED ISSUE #1536: Add support for changing Participant Team Name
Currently, we don't provide support for editing the participant team name on UI. We would like to add a support for this.

isht3:
Fixes #1536


RishabhJain2018:
@isht3 Thanks for the quick fix! I will review the PR asap.

frontend/src/js/controllers/teamsCtrl.js
```diff
@@ -359,6 +359,55 @@
...
+                var parameters = {};
+                parameters.url = 'participants/participant_team/' + participantTeamId;
+                parameters.method = 'PATCH';
+                parameters.data = {
+                    "team_name": result,
+                };
+                parameters.token = userKey;
+                parameters.callback = {
+                    onSuccess: function() {
+                        $rootScope.notify("success", "The Participant team name was successfully updated.");
```
RishabhJain2018: `The Participant team name was successfully updated.` -> `The Participant team name is successfully updated!`

frontend/src/js/controllers/teamsCtrl.js
```diff
@@ -359,6 +359,55 @@
...
+                        var parameters = {};
+                        parameters.url = 'participants/participant_team';
+                        parameters.method = 'GET';
+                        parameters.token = userKey;
+                        parameters.callback = {
+                            onSuccess: function(response) {
+                                vm.existTeam.results = response.data.results;
+                            },
+                            onError: function() {
+                                $rootScope.notify("error", "Name couldn't be changed.");
```
RishabhJain2018: `Name couldn't be changed.` -> `Team name cannot be changed!`

frontend/src/js/controllers/teamsCtrl.js
```diff
@@ -359,6 +359,55 @@
             }, function() {
             });
         };
+
+
+        vm.editTeamName = function(ev, participantTeamId) {
+            ev.stopPropagation();
+            // Appending dialog to document.body to cover sidenav in docs app
```
RishabhJain2018: Can you please modify the comment to make it meaningful ? Also please add comments where ever you think they are needed.

frontend/src/js/controllers/teamsCtrl.js
```diff
@@ -359,6 +359,55 @@
...
+            var confirm = $mdDialog.prompt()
+                .title('Change Team Name')
+                .textContent('Enter the new name for the Team.')
```
RishabhJain2018: `Enter the new name for the Team.` -> `Enter new team name`

isht3:
@RishabhJain2018 I made the changes!

deshraj:
The code looks good but there are two problems:

**Case 1**

Cannot see the alert message when there is an error. Please see the screenshot below:

screen shot 2018-02-28 at 6 58 52 pm 3

**Case 2**

Follow the following steps and then you will see what is the problem:
1. Create a team name "TeamA"
2. Create a team name "TeamB"
3. Delete "TeamA"
4. Try to rename "TeamB" to "TeamA". You won't be able to do this and see an error saying "participant team with this team name already exists". I agree that this is different issue but would be good to get it solved in this since we have encountered it now.

deshraj:
@isht3 do you have updates on this?

deshraj:
@isht3 do you have updates on this?

isht3:
@deshraj Sorry, I forgot to look into it. I'll do it in 24 hours.

deshraj:
@isht3 I think you can fix the **Case 1** and then we can get this merged.

isht3:
@deshraj, I checked it and the error happens because when the team is being deleted usually the participant is removed from the team and the team is intact. This can be rectified by removing the team when the last participant of the team deletes the team.

deshraj:
Yes agree. Would you mind creating a pull request for that?

deshraj:
Also, please fix the **Case 1** as I have mentioned the above comment and then we can get this merged.

isht3:
@deshraj That is directly related to this issue because the response JSON is delivered as `team_name: ["participant team with this team name already exists."]` not as an error, so it will be solved when the backend is fixed.

deshraj:
@isht3 would you mind working on the backend fix for the same? Also, thanks for keeping up with the pace.

isht3:
@deshraj No problem! Which backend fix? 😄

deshraj:
> @deshraj That is directly related to this issue because the response JSON is delivered as team_name: ["participant team with this team name already exists."] not as an error, so it will be solved when the backend is fixed.

I meant this one.

isht3:
@deshraj A fix for giving out the response as an `["participant team with this team name already exists."]` error?

deshraj:
Yeah either this or we can process the `key` and `value` of the json being returned to the frontend if there is an error and show the `value` as the notification. Does that make sense?

isht3:
@deshraj Yes, that kind of makes sense, I will do the second method of processing the key value in the front-end and will let you know if anything else comes up. Should that be added to this PR?

deshraj:
Yeah adding to this PR sounds good.

isht3:
@deshraj It's fixed!

frontend/src/js/controllers/teamsCtrl.js
```diff
@@ -359,6 +359,62 @@
...
+                    },
+                    onError: function(response) {
+                        var error;
+                        if ('team_name' in response.data) {
+                            error = response.data['team_name'];
+                        }
+                        else {
+                            error = response.data['error'];
+                        }
+                        $rootScope.notify("error", error);
```
deshraj: Since `error` variable is a list of errors thrown from the backend, so can you please make sure that we are not showing `[` and `]` while showing the errors. The ideal solution would be to iterate on the error list and then show the error.

isht3:
@deshraj Done!

PR #1541, opened 2018-02-25
Fix #1341: Increase width of edit details dialog box

### LINKED ISSUE #1341: Increase the width of edit details dialog box
**Observed Behaviour :**
dialog_box
The dialog box looks small in the screenshot.

**Expected Behaviour:** Increase the width of the dialog box suitably and paste the screenshot in the PR.

Obtin:
IN concern with #1341
Before:
check
After:
check4

### DIFF
--- a/frontend/src/css/modules/edit-challenge.scss
+++ b/frontend/src/css/modules/edit-challenge.scss
@@ -2,8 +2,12 @@
 	padding: 20px;
 }
 
-.width-500 {
-	width: 500px;
+.width-550 {
+	width: 550px;
+}
+
+.font-size-16 {
+        font-size: 18px;
 }
 
 trix-toolbar .trix-button-row {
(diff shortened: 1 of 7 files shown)


Obtin:
@RishabhJain2018, Does it look fine? Please suggest me the changes if needed and if not I could follow the same approach for other edit boxes as well.

frontend/src/css/modules/edit-challenge.scss
```diff
@@ -2,8 +2,12 @@
 	padding: 20px;
 }
 
-.width-500 {
-	width: 500px;
+.width-550 {
+	width: 550px;
```
deshraj: I don't think making it fixed will help on larger screens. You might want to set it by `%` so that it can be proportionate according to the screen resolution.

Obtin:
Please suggest me which one would be a better way.
- Solution 1: we can remove  ` display:block` from trix-toolbar
edit1
- Solution 2:
setting `width:90%`
edit2
I think the 2nd approach would be better.

deshraj:
I would let @aka-jain make a call here. @aka-jain can you please suggest here what is the better way?

Obtin:
@aka-jain I guess you missed it. Please have a look at it.

RishabhJain2018:
@Obtin Yes, you can go with the second approach.

Obtin:
@RishabhJain2018 I will fix this soon. :-)

RishabhJain2018:
Thanks @Obtin for the efforts. I have updated your PR in #1666. Hence, I am closing this one :-).

PR #1551, opened 2018-03-09
FIX: Delete participant team when the last participant leaves the team.

isht3:
Follow up #1538

### DIFF
--- a/apps/participants/views.py
+++ b/apps/participants/views.py
@@ -234,4 +234,7 @@ def remove_self_from_participant_team(request, participant_team_pk):
         return Response(response_data, status=status.HTTP_403_FORBIDDEN)
     else:
         participant.delete()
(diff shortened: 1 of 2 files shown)


apps/participants/views.py
```diff
@@ -234,4 +234,7 @@ def remove_self_from_participant_team(request, participant_team_pk):
         return Response(response_data, status=status.HTTP_403_FORBIDDEN)
     else:
         participant.delete()
+        participants = Participant.objects.filter(team=participant_team)
+        if len(participants) == 0:
```
deshraj: Can we change it to `if participants.count() == 0:`

isht3:
@deshraj Okay. :smile:

isht3:
@deshraj I've made the desired changes!

PR #1557, opened 2018-03-11
Fix #1547: add support for sorting leaderboard column wise

### LINKED ISSUE #1547: Add support for sorting leaderboard column wise
## Current Scenario

One cannot sort the fields based on different columns in the leaderboard for a challenge.

## Deliverables

- [ ] Add support for sorting leaderboard based on different columns
- [ ] Make sure to add support for sorting based on the submission date. For example, see VQA 2017 Challenge leaderboard: https://evalai.cloudcv.org/web/challenges/challenge-page/1/leaderboard/
- [ ] Add proper comments in the codebase
- [ ] Write tests if possible to make sure that you are covering all the corner cases such as when the data is empty etc.

vladan-jovicic:
Fix #1547:
Leaderboard table can be sorted column-wise in ascending or descending order.

Examples:
sort by score2 in decreasing order

sort by submission date in increasing order (from earliest to latest)


frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -33,7 +33,9 @@
         vm.stopLeaderboard = function() {};
         vm.stopFetchingSubmissions = function() {};
         vm.currentDate = null;
-
+        vm.sortColumn = 'score';
```
deshraj: @vladan-jovicic can we make this dynamic? The label name comes from the backend so there can be multiple labels. For example, VQA Challenge has 4 labels `yes/no`, `number`, `other`, `overall`. 

To get the labels for a particular challenge, you will have to fetch the label name from Leaderboard API for the particular challenge.

deshraj: I think that your approach is a lot neater and we can go with this. One minor request, can we change `columnSort = 'score'` to `columnSort = 'number'` and `sortColumn === 'team'` to `sortColumn === 'string'` for better readability.

vladan-jovicic:
@deshraj can you review it now?

deshraj:
@vladan-jovicic can you please resolve the conflicts?

deshraj:
@vladan-jovicic any update on this?

vladan-jovicic:
@deshraj I tried to solve the conflicts in my local repo, but it did not go well since many changes have been made in the leaderboard table. I need to modify my sorting function since the table now contains a a column with date and delta time (time passed since last submission). I will finish it this week.

deshraj:
Sounds good. Thanks for the update.

frontend/src/views/web/challenge/leaderboard.html
```diff
@@ -28,28 +28,54 @@
...
-                                        <td class="align-right" ng-repeat="score in key.result track by $index" >{{score | number : 2}}</td>
-                                        <td class="align-right">
-                                        {{key.submission__submitted_at | number: 0}}&nbsp;{{key.timeSpan}} ago
-                                        </td>
-                                    </tr>
+                                <tr ng-repeat="key in challenge.leaderboard|orderBy:challenge.sortFunction:challenge.reverseSort">
+                                    <td class="align-left">{{$index+1}}</td>
+                                    <td class="font-weight-b align-left">{{key.submission__participant_team__team_name}}</td>
+                                    <td class="align-right" ng-repeat="score in key.result track by $index" >{{score | number : 2}}</td>
+                                    <td class="align-right">{{ key.submission__submitted_at | date:'medium'}}&nbsp;{{key.timeSpan}} ago</td>
```
deshraj: there seems to be a small error here. Can you please replace the line 77 and 78 with this code:

```
<td class="align-right">{{key.submission__submitted_at | number: 0}}&nbsp;{{key.timeSpan}} ago</td>
```

frontend/src/views/web/challenge/leaderboard.html
```diff
@@ -28,28 +28,54 @@
...
                             <table ng-if="challenge.isResult && challenge.leaderboard.length" class="highlight">
                                 <thead>
-                                   <tr>
-                                        <td class="align-left" data-field="rank">Rank</td>
-                                        <td class="align-left" data-field="team">Participant Team</td>
-                                        <td class="align-right" ng-repeat="key in challenge.leaderboard[0].leaderboard__schema.labels">{{key}}</td>
-                                        <td class="align-right" data-field="submission_time">Last Submission at</td>
-                                    </tr>
+                                <tr>
+                                    <td class="align-left" data-field="rank">
```
deshraj: It looks like you are not shifting the value of Rank column when sorted. Can you please make sure that this column value is also shifted whenever someone tries to sort the leaderboard?

vladan-jovicic: final

Does it look good?

Remark: values in rank column will be always 1 ... n or n ... 1. In other words, sorting does not preserve the initial ranking number of a participant team. Should it be like this or you want to keep the initial values when the leaderboard is sorted by some other column?

vladan-jovicic: another question: do the arrows look good or I should reverse them?

deshraj: > Remark: values in the rank column will be always 1 ... n or n ... 1. In other words, sorting does not preserve the initial ranking number of a participant team. Should it be like this or you want to keep the initial values when the leaderboard is sorted by some other column?

The rank of the participants should be preserved to keep the initial values when the leaderboard is sorted by some other column.

Also, it would be even cooler if we can add the feature to have a unique URL for each sorting type.
For example: 
Sort in ascending order by `score1` has URL: `http://localhost:8888/web/challenges/challenge-page/1/leaderboard/1?sortby=score1`

Sort in descending order by `score1` has URL: `http://localhost:8888/web/challenges/challenge-page/1/leaderboard/1?sort=-score1`

Let me know if you will be able to add this feature in this pull request. Otherwise, we can open a new issue for this.

deshraj: > another question: do the arrows look good or I should reverse them?

I do think that the arrows can look much better. Can you check this table UI: https://dribbble.com/shots/4218286-Media-Configuration-Table

We would like to have something similar where the column arrow gets highlighted when it is sorted by that. Do you see what I mean?

vladan-jovicic: Is there a way to have these URLs without modifying the backend code? If there is one, can you give me an example? If no, then I should be able to do this by modifying challenge urls.

Regarding the arrows: consider it done.

deshraj: > Is there a way to have these URLs without modifying the backend code? If there is one, can you give me an example? If no, then I should be able to do this by modifying challenge urls.

I think it's fine for now to not have the URL feature. That would delay this feature. We would go with the current implementation only once you have changed the arrows.

deshraj:
@vladan-jovicic any updates on this?

vladan-jovicic:
@deshraj Now it looks like this:

deshraj:
Thanks @vladan-jovicic. This all looks perfect. I think there is just one small thing you are missing here. Can you not sort the leaderboard when it is shown for the first time. The default order should be properly ranked which is not the case right now. Please see the screenshot attached: 
screen shot 2018-04-17 at 11 50 13 am

PR #1565, opened 2018-03-16
Fix #1561: Improve dataset_split selection logic in submission worker

### LINKED ISSUE #1561: Improve dataset_split selection logic in submission worker
After integrating #1560, there can be multiple dataset_splits with the same name and codename and the code at  

https://github.com/Cloud-CV/EvalAI/blob/master/scripts/workers/submission_worker.py#L335

will fail as it will return `MultipleObjectsReturned` exception in the case where there are two or more dataset splits which have the same values of codename and split name.

deshraj:
Fixes #1561

### DIFF
--- a/scripts/workers/submission_worker.py
+++ b/scripts/workers/submission_worker.py
@@ -48,7 +48,6 @@ DJANGO_SERVER = os.environ.get('DJANGO_SERVER', "localhost")
 from challenges.models import (Challenge,
                                ChallengePhase,
                                ChallengePhaseSplit,
-                               DatasetSplit,
                                LeaderboardData) # noqa
 
 from jobs.models import Submission          # noqa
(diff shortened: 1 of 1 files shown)


scripts/workers/submission_worker.py
```diff
@@ -329,24 +328,26 @@ def run_submission(challenge_id, challenge_phase, submission_id, submission, use
...
                 try:
-                    split_code_name = split_result.items()[0][0]  # get split_code_name that is the key of the result
-                    dataset_split = DatasetSplit.objects.get(codename=split_code_name)
+                    challenge_phase_split = ChallengePhaseSplit.objects.get(challenge_phase=challenge_phase,
+                                                                            dataset_split__codename=split_code_name)
                 except:
-                    stderr.write("ORGINIAL EXCEPTION: The codename specified by your Challenge Host doesn't match"
-                                 " with that in the evaluation Script.\n")
+                    stderr.write("ORGINIAL EXCEPTION: No such relation between between Challenge Phase and DatasetSplit"
+                                 " specified by Challenge Host \n")
```
RishabhJain2018: Please change the exception message to : `No such relation is found between Challenge Phase and DatasetSplit specified by Challenge Host` .

deshraj: Thanks. Will change this.

PR #1581, opened 2018-03-20
Submission Worker: Add support for keyword arguments for submission evaluation

deshraj:


deshraj:
Currently, we don't give access to the submission metadata to the challenge evaluation script due to which if a challenge host wants to send a slack notification with all the relevant details then he cannot do that. This PR solves that problem.

PR #1610, opened 2018-03-29
Settings: Update django project settings

deshraj:
This PR solves the following problems:

- Easy installation instructions (removes the dependency of adding `--settings=settings.dev` for each management command)
- Help to setup celery in an easier way for both local and production environment
- Removes the dependency on three project settings. We can use settings/prod.py now for the staging environment. This will make sure that our staging environment has the exact configuration of the production environment.


RishabhJain2018:
@deshraj There is some error. Please take a look at it.
error

deshraj:
Seems like this is an issue with the Python version. Can you try with python version >=2.7.10?

isht3:
@deshraj I'm using 2.7.12, I'm still getting this error.

RishabhJain2018:
@isht3 Did you try debugging it ?

isht3:
@RishabhJain2018 No, I didn't. I will look into it and will let you know.

RishabhJain2018:
@isht3 Please check what python version you are using in virtualenv. It should be >=2.7.10 because it worked for me. Let us know if you still face this error.

Moreover, Thanks @deshraj !

RishabhJain2018:
@isht3 We have fixed the bug in running development server. Please take a look at https://github.com/Cloud-CV/EvalAI/pull/1613. Also, make sure that you have the latest pull of master, once you start hacking.

isht3:
@RishabhJain2018 YES! It started working. Thanks a lot!

Issue #1606, opened 2018-03-28
Allow and block particular email domains for a challenge


deshraj:
Currently, anyone can participate in a challenge. It would be great to have a functionality to allow/block only particular email domains to participate in a challenge as specified by the challenge host. 

## Deliverables

- [x] Add a field to Challenge model to store the `allowed_email_domains` (default=None, blank=True)
- [x] Add a field to Challenge model to store the `blocked_emails_domains` (default=None, blank=True)
- [ ] Update the challenge creation API to add support to incorporate `allowed_email_domains` and `blocked_email_domains`
- [ ] Update the Challenge creation using zip file documentation to add support for `allowed_email_domains`   and `blocked_email_domains`
- [ ] Update the challenge participation API to only allow `allowed_email_domains` and block `blocked_email_domains`

isht3:
@deshraj @RishabhJain2018 Can I work on this?

deshraj:
Yeah sure. Assigning this to you.

isht3:
@deshraj for the `allowed_email_domains` and the other fields I should be using `ArrayField`, yes?

deshraj:
Yes, you are right.

isht3:
@deshraj, the front-end support for this is not yet required, yes?

isht3:
@deshraj The last part of this issue is still remaining. I'll do it ASAP.

deshraj:
You mean the last 3 parts, right?

deshraj:
Please add all of these remaining things in a single Pull request.

PR #1612, opened 2018-03-30
Add domain fields, migrations and add support in the Challenge Creation API.

isht3:
Fields and migrations for the feature.
Adds the fields to the ChallengeSerializer.
Related to #1606


apps/challenges/models.py
```diff
@@ -45,6 +45,12 @@ def __init__(self, *args, **kwargs):
         default=False, verbose_name="Approved By Admin", db_index=True)
     featured = models.BooleanField(
         default=False, verbose_name="Featured", db_index=True)
+    allowed_email_domains = ArrayField(
+        models.CharField(max_length=20, blank=True),
```
deshraj: 20 seems to be too less. Can you please change it to 50?

apps/challenges/migrations/0034_auto_20180330_1024.py
```diff
@@ -0,0 +1,26 @@
+# -*- coding: utf-8 -*-
```
deshraj: Please make sure that you name the migration while running the command `makemigrations`

isht3:
@deshraj Added!

PR #1624, opened 2018-04-03
Update feature to send emails to evalai users using amazon ses

### LINKED ISSUE #1549: Add Feature to notify users about new challenge on EvalAI.

RishabhJain2018:

Fixes #1549


apps/web/views.py
```diff
@@ -63,31 +59,31 @@ def notify_users_about_challenge(request):
...
-                          'notification_email_conformation.html',
-                          {'message': 'All the emails are sent successfully!'})
+            try:
+                email.attach(image)
+                email.send()
+                return render(request,
+                              'notification_email_conformation.html',
+                              {'message': 'All the emails are sent successfully!'})
+            except:
+                return render(request,
```
deshraj: Whenever there is an exception, try to log it with proper severity level.

apps/web/views.py
```diff
@@ -63,31 +59,31 @@ def notify_users_about_challenge(request):
...
+                              'notification_email_data.html',
+                              {'errors': 1})
         else:
             return render(request, 'error404.html')
     else:
-        return render(request, 'error404.html')
+        return render(request, 'error_superuser.html')
```
deshraj: Just return 404. No need to show exactly what happened.

apps/web/views.py
```diff
@@ -49,12 +47,10 @@ def notify_users_about_challenge(request):
             return render(request, template_name)
 
         elif request.method == 'POST':
-            template_name = 'notification_email.html'
-            emails = User.objects.all().exclude(email__isnull=True, email__exact='').values_list('email', flat=True)
-            htmly = get_template('notification_email.html')
+            users = list(User.objects.exclude(email__exact='').values_list('email', flat=True))
```
deshraj: Do we really need to convert to list? Doesn't `values_list` takes care of that? I am not sure about this though. Please add a comment on the line why we are converting to list if we need to do this.

RishabhJain2018: No, for now we don't need that. Earlier `boto` required it to be converted to a list rather than `Queryset` object. :-)

Issue #1492, opened 2018-01-26
Modify database seeding method


deshraj:
## Deliverables

- [ ] Modify the seed command to accept an argument `-nc` which denotes the number of challenges that a developer wants to create for testing.
- [ ] Modify the seed script in such a way that it can be run multiple times (right now, if you run it once and then you try to run it again, then it throws an error).

Obtin:
@deshraj I would like to proceed on this issue. Thank you

deshraj:
@Obtin: sure, go ahead. :)

isht3:
@Obtin Are you working on this?

Obtin:
@isht3 yes :-)

anweshknayak:
@Obtin I see you are working in #1373. Please take one issue at a time. 😃 So that others can work too.

Obtin:
Done with #1373 , Starting working n this. Thank you.

Obtin:
@deshraj I made changes in apps/base/management/commands/seed.py
'    def add_arguments(self, parser):
        parser.add_argument(
            '--nc', dest='nc', nargs='?', default=3,
            help = 'no.of challenges',
        )
'
`    def handle(self, *args, **options):
        self.nc = options['nc']
        self.stdout.write(self.style.SUCCESS('Starting the database seeder. Hang on...'))
        call_command('runscript', 'seed', '--settings', 'settings.dev')
        self.stdout.write(self.style.SUCCESS('Database successfully seeded.'))
`
the ouput is as follows: 
![modify](https://user-images.githubusercontent.com/25207205/36258480-cc6a93ae-1280-11e8-8bdb-d474a3834c10.png)
Further Plan: The 'nc' variable will be passed as an argument in scripts/seed.py file .
 
I am getting an error

> Exception while running run() in 'scripts.seed'
I guess it is due to the problem in `def run ` function.
Please let me know whether its right or I should continue with any other approach.

Obtin:
@deshraj please have a look at it.

isht3:
Hi @Obtin is the seeding exception issue resolved? :smile:

Obtin:
@isht3 working on it will update you soon. :-)

isht3:
@Obtin Okay, cool. I can work on the seeding exception if you have other issues to work on. :smile:

Obtin:
@isht3 yeah sure, feel free to work on it.  :-)

deshraj:
[...]

PR #1626, opened 2018-04-04
Fix the seed script from throwing Exception while running run().

isht3:
This PR fixes the following error.

It adds an option for destroying the current database. 

I'll be adding the `-nc` command support and the pretty printing from the issue #1492 next.


isht3:
@deshraj Do put in your thoughts! :smile:

PR #1642, opened 2018-04-08
Docs: Minor Typo fix

deshraj:


RishabhJain2018:
@deshraj I think we should also add that the time in yaml file should be in UTC format.

deshraj:
@RishabhJain2018 can you please edit it in this branch?

RishabhJain2018:
sure!

PR #1660, opened 2018-04-21
FIX: Limit number of concurrent submissions for a participant.

### LINKED ISSUE #1618: Limit number of concurrent submissions for a participant
## Current Scenario

Currently, a participant can make any number of submissions until he reaches the limit of doing max submissions per day. We would like to add another constraint where a participant can do at max 3  submissions and if he wants to do more then he has to wait until the status of the previously done submissions gets changed to `failed` or `canceled` or `finished`. This will help us to make sure that the participant cannot abuse the system by just making an unlimited number of submissions if the limit of doing max submissions per day is quite large. 

## Deliverables

- [ ] Add a new field `max_concurrent_submissions_allowed` integer field with default value set as 3.
- [ ] Add a constraint that 

```
if max_submissions_per_day < max_concurrent_submissions_allowed:
    set max_concurrent_submissions_allowed = max_submissions_per_day
```
Note: This would require to override the `save()` method I guess.
- [ ] Update the submission API to check:

```
if (number_of_submissions_with_running_status + number_of_submissions_with_submitted_status + number_of_submissions_with_submitting_status) == max_concurrent_submissions_allowed:
    # Don't allow the participant to do another submission
else:
    # Proceed and allow him to make submission
```

- [ ] When the participant is exceeding the limit then show a message like this:
`You have X submissions that are already being processed. Please wait for sometime to make another submission.`

isht3:
Fixes #1618


isht3:
@deshraj I did it!

deshraj:
@isht3 the code looks good. What all remains is test case.

PR #1672, opened 2018-04-26
Change pk --> Index so that pk is not exposed.

### LINKED ISSUE #1671: We should not expose the PK for any phase or challenge_phase_split in the url.
## Current Scenario
The pk of the challenge phase and challenge phase split appears in the url.

### Step 1: 
write a script to modify the value of id for different challenge phases of the same challenge. For VQA Challenge with pk 80, it has 3 phases with id = pk = (100, 101, 102) and 4 challenge phase splits with id = pk = (41,42,43,44)

What we might want to do here is that:
1. add a new field with name `index` with default as None
2. Write script to modify the field for each challenge phase based on the previous script that we shared
3. Modify the challenge creation API to add support for populating this field

After step 1,
  `challenge phase pk = (100,101,102) and challenge phase id = (1,2,3)
  challenge phase split pk = (41,42,43,44) and challenge phase split id = (1,2,3,4)`

### Step 2:

  Modify all the URLs that work like this. See the below example:

  2.1 Old URL: `/api/80/submit/phase/<pk>`
  2.2 New URL: `/api/80/submit/phase/<index>`
  2.3 Modify the API definition to take the index for filtering instead of pk for challenge phase.

isht3:
Fixes #1671


apps/analytics/views.py
```diff
@@ -107,15 +108,21 @@ def get_submission_count(request, challenge_pk, duration):
...
     challenge = get_challenge_model(challenge_pk)
 
-    challenge_phase = get_challenge_phase_model(challenge_phase_pk)
+    try:
+        challenge_phase = ChallengePhase.objects.get(
+            challenge=challenge, phase_id=challenge_phase_id)
+    except ChallengePhase.DoesNotExist:
+        response_data = {
+            'error': 'Challenge Phase {} does not exist'.format(challenge_phase_id)}
+        return Response(response_data, status=status.HTTP_404_NOT_FOUND)
```
RishabhJain2018: @isht3 Please reduce this redundant code everywhere by compiling it into a function. For reference please take a look at https://github.com/Cloud-CV/EvalAI/blob/master/apps/base/utils.py#L43

isht3: I tried and it didn't work. Since the Response is returned when the ChallengePhase does not exist, it throws an error. And if we only take the get statement out we cannot do the error handling.

apps/analytics/views.py
```diff
@@ -107,15 +107,20 @@ def get_submission_count(request, challenge_pk, duration):
...
+def get_challenge_phase_submission_analysis(request, challenge_pk, challenge_phase_id):
     """
     API to fetch
     1. The submissions count for challenge phase.
     2. The participated team count for challenge phase.
     """
     challenge = get_challenge_model(challenge_pk)
 
-    challenge_phase = get_challenge_phase_model(challenge_phase_pk)
+    challenge_phase = get_challenge_phase_from_phase_id(challenge, challenge_phase_id)
```
RishabhJain2018: @isht3 Please add this function in `apps/base/utils.py` instead of the above one.
`def get_model_object_by_id(model_name):
    """
    A function to get the model object by id in which challenge model acts as a foreign key.
    """

    def get_model_by_id(idx, challenge):
        try:
            model_object = model_name.objects.get(challenge=challenge, phase_id=idx)
            return model_object
        except model_name.DoesNotExist:
            raise NotFound('{} {} does not exist'.format(model_name.__name__, idx))
    get_model_by_id.__name__ = 'get_{}_object'.format(model_name.__name__.lower())
    return get_model_by_id`
Please test it on your local once, I think things will work fine with it.

apps/challenges/models.py
```diff
@@ -94,6 +94,7 @@ def is_active(self):
```
RishabhJain2018: Please add `phase_id` and `phase_split_id` in the in the `apps/challenges/admin.py` in `list_display` field.

apps/challenges/utils.py
```diff
@@ -20,3 +20,11 @@ def get_file_content(file_path, mode):
     if os.path.isfile(file_path):
         with open(file_path, mode) as file_content:
             return file_content.read()
+
+
+def get_challenge_phase_from_phase_id(challenge, challenge_phase_id):
```
RishabhJain2018: Please remove this function.

apps/challenges/migrations/0037_challenge_foreign_key_to_challenge_phase_split.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
```
RishabhJain2018: Please remove this file and corresponding model.

apps/challenges/migrations/0038_challenge_foreign_key_to_dataset_split.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
```
RishabhJain2018: Please remove this and corresponding models.

apps/analytics/views.py
isht3: @RishabhJain2018 Every time I try to import Challenge models onto `base/utils.py` it gives me `ImportError: cannot import name ChallengePhase` when doing `from challenges.models import ChallengePhase`. Also would you please explain why we are using the above decorator instead of the other function?

apps/challenges/models.py
isht3: Okay!

apps/analytics/views.py
RishabhJain2018: @isht3 I am not sure why you are getting this error as I checked and it is working for me. Can you please check once.
Moreover,  I am asking you to add this decorator so that we don't have to handle the error response separately everywhere.

isht3: @RishabhJain2018 The CI tests are failing because of that. It's because of the import loop from having `base.utils` inside the `challenge.models`. Is it okay if I move this function to `challenge.utils` to avoid the import loops?

RishabhJain2018:
Closing this issue for now since we converged other design decisions.

PR #1674, opened 2018-04-27
Improve backend logic for adding members to Participant and Host Team

deshraj:
This Pull Request brings in the following additions:

- [x] Change 'Invite Members' to 'Add members' on UI since the feature is actually adding members than that of inviting.
- [x] Add logic for restricting users who are not part of the host or participant team
- [x] Improve the test suite
- [x] Add new test cases for the new logic
- [x] Improve the code structure in tests
- [x] Add a FAQ to test in Docker containers


apps/hosts/utils.py
```diff
@@ -15,4 +15,8 @@ def is_user_a_host_of_challenge(user, challenge_pk):
     return Challenge.objects.filter(pk=challenge_pk, creator_id__in=challenge_host_teams).exists()
 
 
+def is_user_part_of_host_team(user, host_team):
```
isht3: This is to make sure only the creator of the host team can add participants to it, yes?

apps/participants/views.py
```diff
@@ -114,6 +114,15 @@ def invite_participant_to_team(request, pk):
         response_data = {'error': 'User does not exist with this email address!'}
         return Response(response_data, status=status.HTTP_406_NOT_ACCEPTABLE)
 
+    if not is_user_part_of_participant_team(request.user, participant_team):
```
isht3: This cannot be triggered using the UI, yes?

apps/hosts/utils.py
deshraj: Yeah this added to check if any person is part of the host team or not and hence based on the result returned from this, the person will be allowed to add people to his team.

apps/participants/views.py
deshraj: I didn't get your question here.

RishabhJain2018:
One corner case that I think is missing here is that `A participant should not be given right to add a challenge host`.

Why?
Suppose if there are a large number of users who are adding challenge host to their team by using the email address then a large number of `participant teams` will appear on the challenge hosts dashboard which can make him confused about his own teams. Moreover, he will not be able to remove himself from the teams since the challenge will be going on.

What I think should be the way?
- Any participant can't add a challenge host to his team.
- A challenge host can create a team and participate in the challenge.

@deshraj Please share your thoughts.

deshraj:
It totally makes sense. Thanks Rishabh. I will incorporate the changes and will update the pull request by tonight.

RishabhJain2018:
yes, sure 😄

isht3:
I ran it locally. It's working! :smile:

deshraj:
@RishabhJain2018 I tested the use case that you mentioned and found out that it will never happen that a participant will be able to invite add challenge host to his team since we are already assigning a team to the challenge host when he creates the challenge (see https://github.com/Cloud-CV/EvalAI/blob/master/apps/challenges/views.py#L892)

deshraj:
I am going to merge this pull request now since there were no requested changes else than what @RishabhJain2018 said. Please feel free to comment if there is some missing case. Thanks!

PR #1692, opened 2018-06-02
Update submission worker to create challenge specific queues

### LINKED ISSUE #1691: Update submission worker to create challenge specific queues.
### Deliverables:

The submission worker should be updated so that challenge specific queues can be created instead of one single queue for all the challenges.

**Challenge Queue name:** submission_challenge_<challenge_pk>

Make sure to use the exchange type as **direct**.(refer [here](https://www.rabbitmq.com/tutorials/tutorial-four-python.html))

RishabhJain2018:
Fix #1691 . @deshraj Please review the PR.

The removed code from the submission worker file (submission_worker.py) points to the feature that whenever there is a change in the evaluation script or a new challenge is added, then it will load it's configuration into the worker without actually reloading it.
I have removed that code as it is creating one unnecessary queue with every challenge which isn't required. Moreover, we are also not supporting the feature of automatically reloading the submission worker. Hence, it is of no use.


scripts/workers/submission_worker.py
```diff
@@ -212,22 +213,21 @@ def extract_challenge_data(challenge, phases):
     EVALUATION_SCRIPTS[challenge.id] = challenge_module
 
 
-def load_active_challenges():
+def load_active_challenge(challenge_pk):
```
deshraj: Let's rename this method to `load_challenge` since we only load a challenge now and not check whether it is active or not.

scripts/workers/submission_worker.py
```diff
@@ -212,22 +213,21 @@ def extract_challenge_data(challenge, phases):
...
-    q_params = {'approved_by_admin': True}
+    q_params = {'approved_by_admin': True, 'pk': challenge_pk}
     q_params['start_date__lt'] = timezone.now()
     q_params['end_date__gt'] = timezone.now()
 
     # make sure that the challenge base directory exists
     create_dir_as_python_package(CHALLENGE_DATA_BASE_DIR)
 
-    active_challenges = Challenge.objects.filter(**q_params)
+    active_challenge = Challenge.objects.filter(**q_params)
```
deshraj: Rename this to `challenge`

deshraj: Also, instead of filter, we can use the get method with proper exception handling.

RishabhJain2018: sure 👍

isht3:
@RishabhJain2018 @deshraj I messed things up while rebasing the changes. I would normally `reset HEAD`. Can you give me the right advice to revert back? :cry:

isht3:
Also, I tested this PR. So this PR works perfectly for handling a single challenge and it's submissions but to add support for multiple challenges we're going to use a different container for each challenge, for that

* A master script for looking over this multiple operations would be at `scripts/master_worker.py` which would load the active challenges.
* It also creates a new container for each of the challenge using docker-py like `CMD ["./docker/wait-for-it.sh", "django:8000", "--", "python", "scripts/workers/submission_worker.py",  "--challenge_pk", "11" ]` which listens in on it's own unique queue (RabbitMQ here, but the same logic for SQS).
* Creates new containers for new challenges without the need to reload the worker.

Please let me know if my logic here is wrong.

RishabhJain2018:
@isht3 I have reset the changes that you did :-). Can you please test the PR again?

isht3:
@RishabhJain2018 Yup, tested it. Everything works as intended. I'll finish my other to do's and start implementing the master worker.

RishabhJain2018:
Closing this since I implemented it again using SQS.

Issue #1696, opened 2018-06-05
gulp-deprecation error while running docker-compose command


HargovindArora:
While running the `docker-compose up --build` command to install and compose the docker container to set up the project the command is exiting with an error. I'm attaching the screenshots of the errors that were occurring even after so many tries.

![docker_gulp](https://user-images.githubusercontent.com/22341493/40987613-9c4663e2-6906-11e8-91c6-49e07cf77e46.png)

![docker_gulp_2](https://user-images.githubusercontent.com/22341493/40987625-a6f54e20-6906-11e8-9dca-b757cc02d94d.png)

![screenshot from 2018-06-05 21-18-31](https://user-images.githubusercontent.com/22341493/40987639-b1bcf4b6-6906-11e8-9180-b7c9b00253e1.png)

Please resolve the above issue as it might be a hurdle for new contributors to set up the project.

spyshiv:
Currently I am working on it. Will update the gulp tasks soon.

HargovindArora:
Ah. Ohkay. And the problem persists even while following trivial installation procedure(non-docker approach). Thanks :)

deshraj:
@HargovindArora the gulp deprecation is a warning and that is not affecting the project in any way as of now. We are working on it as @spyshiv mentioned. Your error seems like while building docker container, your machine is not able to download the package due to some internet issues. Can you please try removing the containers and do it again? It might work then or try changing the internet connection that you are using.

varunagrawal:
Else try restarting the docker daemon. That usually works for me when docker is having a hard time connecting to the internet.

HargovindArora:
@deshraj @varunagrawal I'm trying your given solutions and will let you know if it worked. But "gulp" warning surely did trouble me a lot while following the non-docker installation procedure.

deshraj:
You should check out #1697 for the gulp problems. @spyshiv has already fixed it and the pull request is in the review. 

Please do let us know once your error is resolved.

HargovindArora:
@deshraj As @varunagrawal stated there was an error with docker daemon. I tried restarting it but it didn't help. Therefore I completely removed and reinstalled docker and it finally worked.
But the gulp error isn't letting me to setup using non-docker approach. I'll try it again once #1697 is merged.
Thanks :)

RishabhJain2018:
I think this is resolved. Hence, closing this one.
@HargovindArora Feel free to open up if you find some issues.

PR #1697, opened 2018-06-06
Upgrade(Gulp): Version bump 3.9.1 -> 4.0.0

spyshiv:
Fixed issue #1696 
* Rewrite Gulp Tasks
* Update Docker Config

@deshraj @RishabhJain2018 @aka-jain Please review and test on your mahines.


deshraj:
Sorry @spyshiv, I didn't get time yesterday. I will test this out today for sure. @RishabhJain2018 can you please check at your end that it is working properly?

RishabhJain2018:
@deshraj I have tested it & it is working perfectly on my system.

isht3:
@RishabhJain2018 I have some trouble fetching this branch from spyshiv's fork. I'll figure out what's wrong and then test it.

isht3:
@RishabhJain2018 Yup tested it. Works!

frontend/src/js/environment/config.local.js
```diff
@@ -0,0 +1,2 @@
+angular.module("evalai-config", [])
+.constant("EnvironmentConfig", {"API":"http://localhost:8000/api/"});
```
deshraj: So, are we not using `config.json` anymore for fetching these names?

deshraj: It would be great if we can fetch these as environment variables instead of hard-coding here. @spyshiv do you know a way so that we can do that?

deshraj:
Left some minor comments. Overall it's working. We can merge this PR once we converge on the minor comments.

frontend/src/js/environment/config.local.js
spyshiv: @deshraj Actually we were using [gulp-ng-config](https://github.com/ajwhite/gulp-ng-config/) and that project is also using `gulp-util` so either we have to wait for fixing of [issue](https://github.com/ajwhite/gulp-ng-config/issues/71) of `gulp-util` deprecation or we should go for other alternatives. 
So as a quick fix we can consider this with hardcoded values or we can put `environment` folder in `.gitignore`

deshraj: I think the ideal way would be to put in the `config.json` file that we had earlier. Would be much easier when we want to do the continuous deployment. Can you please move it to the file?

spyshiv: Ok. I'll do that

spyshiv:
@deshraj Please review again.

deshraj:
Looks great. Thanks @spyshiv :)

PR #1704, opened 2018-06-15
Fix #1703 Change error message

### LINKED ISSUE #1703: Show proper error message when Yaml file config has errors
### Issue:
Proper error isn't displayed when yaml config for challenge creation isn't correct.

### Culprit Url:
`/api/challenges/challenge/challenge_host_team/{challenge_host_team_pk}/zip_upload/`

### Error:
ScannerError

### Solution
Log the error & display message as:
`Error in creating challenge. Please check the yaml configuration!`
For reference: https://www.programcreek.com/python/example/100007/yaml.scanner.ScannerError

calenrobinette:
Fixes #1703

### DIFF
--- a/frontend/src/js/controllers/challengeCreate.js
+++ b/frontend/src/js/controllers/challengeCreate.js
@@ -70,7 +70,7 @@
                             utilities.hideLoader();
                             var error = response.data;
                             angular.element(".file-path").val(null);
-                            $rootScope.notify("error", error.error);
+                            $rootScope.notify("Error in creating challenge. Please check the yaml configuration!", error.error);
                             vm.stopLoader();
                         }
                     };


frontend/src/js/controllers/challengeCreate.js
```diff
@@ -70,7 +70,7 @@
                             utilities.hideLoader();
                             var error = response.data;
                             angular.element(".file-path").val(null);
-                            $rootScope.notify("error", error.error);
+                            $rootScope.notify("Error in creating challenge. Please check the yaml configuration!", error.error);
```
RishabhJain2018: Hi @calenrobinette , the first error shows that the notification type is of `error` & second one gives the message which is correct that is `error.error`.
We have to change it at the [backend](https://github.com/Cloud-CV/EvalAI/blob/master/apps/challenges/views.py#L548) after caching the error [here](https://github.com/Cloud-CV/EvalAI/blob/master/apps/challenges/views.py#L546)

PR #1711, opened 2018-06-18
Prevent users from participating in the past challenges 

### LINKED ISSUE #1708: Prevent users from participating in the past challenges using the API.
# Current Behaviour

You can participate in any past challenge using the end-point `/api/challenges/challenge/{}/participant_team/{}` which is not the desired behavior.

# Expected Behaviour

We should place a check in the  view to see if it's a request to participate in a past challenge and then block it.

isht3:
Fixes #1708


apps/challenges/views.py
```diff
@@ -168,6 +168,10 @@ def add_participant_team_to_challenge(request, challenge_pk, participant_team_pk
         response_data = {'error': 'Challenge does not exist'}
         return Response(response_data, status=status.HTTP_406_NOT_ACCEPTABLE)
 
+    if challenge.end_date < timezone.now():
+        response_data = {'error': 'Sorry, you cannot participate in a past challenge.'}
```
RishabhJain2018: Please change the error message to : `Sorry, cannot accept participant team since challenge is not active`

tests/unit/challenges/test_views.py
```diff
@@ -486,6 +486,17 @@ def test_particular_challenge_for_mapping_with_participant_team_does_not_exist(s
         self.assertEqual(response.data, expected)
         self.assertEqual(response.status_code, status.HTTP_406_NOT_ACCEPTABLE)
 
+    def test_particular_challenge_for_mapping_with_past_challenge(self):
+        self.challenge2.end_date = timezone.now() - timedelta(days=1)
```
RishabhJain2018: Please change the name to `test_participant_team_for_mapping_with_past_challenge`

tests/unit/challenges/test_views.py
```diff
@@ -486,6 +486,17 @@ def test_particular_challenge_for_mapping_with_participant_team_does_not_exist(s
...
+        self.challenge2.save()
+        self.url = reverse_lazy('challenges:add_participant_team_to_challenge',
+                                kwargs={'challenge_pk': self.challenge2.pk,
+                                        'participant_team_pk': self.participant_team3.pk})
+        expected = {'error': 'Sorry, you cannot participate in a past challenge.'}
```
RishabhJain2018: Please modify this too.

apps/challenges/views.py
```diff
@@ -168,6 +168,10 @@ def add_participant_team_to_challenge(request, challenge_pk, participant_team_pk
         response_data = {'error': 'Challenge does not exist'}
         return Response(response_data, status=status.HTTP_406_NOT_ACCEPTABLE)
 
+    if challenge.end_date < timezone.now():
```
RishabhJain2018: Do add a check for future challenges.
`if challenge.end_date < timezone.now() or challenge.start_date > timezone.now():`
I think the condition will be ^ now. Please test it once.

RishabhJain2018:
Don't forget to add a test case.

apps/challenges/views.py
isht3: @RishabhJain2018 Sounds good! :D

PR #1719, opened 2018-06-28
Make submission metadata field optional in database

### LINKED ISSUE #1718: Make submission metadata field optional in database.
### **Deliverables:**

- [x] Make the `Project URL`, `Publication URL`, `Method Name`, `Method Description` optional in the database

- [x] Update the frontend where we enter the above 4 fields and show in brackets `(optional)`

RishabhJain2018:
Fix #1718 . @deshraj Please review the PR.


frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -581,11 +581,23 @@
             parameters.callback = {
                 onSuccess: function(response) {
                     var details = response.data;
-                    vm.submissionResult = details;
 
                     for (var i = 0; i < details.results.length; i++) {
                         vm.submissionVisibility[details.results[i].id] = details.results[i].is_public;
+                        if (details['results'][i]['method_name'] == "null") {
```
deshraj: You probably need a better way of comparison. Think about the cases where the value can be `""`, `"    "`, `null`, `"null"`. Similar is the case with other fields.

RishabhJain2018: @deshraj Since the database stores the values as `null` even when the participant leaves the field as empty, also if participant fills in `null` value then also it won't break.

PR #1726, opened 2018-06-28
Fix API for total submission done

RishabhJain2018:


apps/analytics/views.py
```diff
@@ -118,7 +118,7 @@ def get_challenge_phase_submission_analysis(request, challenge_pk, challenge_pha
     challenge_phase = get_challenge_phase_model(challenge_phase_pk)
 
     submissions = Submission.objects.filter(
-        challenge_phase=challenge_phase, challenge_phase__challenge=challenge)
+        challenge_phase=challenge_phase, challenge_phase__challenge=challenge, created_by=request.user.pk)
```
deshraj: This check has to be done at the participant team level and not at the level of `created_by`. Please change this.

deshraj: Also, please add test cases for this.

deshraj:
Please add a test case where a team A has done 1 submission and other team B has done two submissions, then if I want to get the data from the API, then it should return 1 instead of 3.

RishabhJain2018:
@deshraj Please review and merge it.

PR #1729, opened 2018-06-29
Add blank field for optional models

RishabhJain2018:
Add `blank=True` in optional models for `submission metadata` and `team_url`.

Reference to why we are doing this: https://docs.djangoproject.com/en/2.0/ref/models/fields/#django.db.models.Field.null


RishabhJain2018:
@deshraj Please review and merge this!

PR #1681, opened 2018-05-03
Python3 compatibility

dhuber23b:

PR #1732, opened 2018-06-30
Fix Python3 migration error along with adding its support

RishabhJain2018:

### DIFF
--- a/apps/challenges/migrations/0019_added_leaderboard_data_table.py
+++ b/apps/challenges/migrations/0019_added_leaderboard_data_table.py
@@ -10,7 +10,7 @@ import django.db.models.deletion
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('jobs', '0004_submission_output'),
+        ('jobs', '0001_squashed_0005_upload_unique_random_filename'),
         ('challenges', '0018_added_challenge_phase_split_table'),
     ]
 
--- a/scripts/workers/submission_worker.py
+++ b/scripts/workers/submission_worker.py
@@ -1,4 +1,9 @@
+
 from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+from __future__ import unicode_literals
+
 import contextlib
 import django
 import importlib
@@ -112,7 +117,7 @@ def download_and_extract_file(url, download_location):
         response = None
 
     if response and response.status_code == 200:
-        with open(download_location, 'w') as f:
+        with open(download_location, 'wb') as f:
             f.write(response.content)
 
 
@@ -128,7 +133,7 @@ def download_and_extract_zip_file(url, download_location, extract_location):
         response = None
 
     if response and response.status_code == 200:
-        with open(download_location, 'w') as f:
+        with open(download_location, 'wb') as f:
             f.write(response.content)
         # extract zip file
         zip_ref = zipfile.ZipFile(download_location, 'r')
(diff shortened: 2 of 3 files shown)


scripts/workers/submission_worker.py
```diff
@@ -186,6 +191,7 @@ def extract_challenge_data(challenge, phases):
     evaluation_script_url = return_file_url_per_environment(evaluation_script_url)
     # create challenge directory as package
     create_dir_as_python_package(challenge_data_directory)
+
```
taranjeet: Can we please remove this line from here? It has no significance related to Python3 migration actually.

scripts/workers/submission_worker.py
```diff
@@ -337,9 +343,8 @@ def run_submission(challenge_id, challenge_phase, submission, user_annotation_fi
 
             leaderboard_data_list = []
             for split_result in submission_output['result']:
-
```
taranjeet: Similarly for this. I think you have run some auto linter tool

scripts/workers/submission_worker.py
```diff
@@ -337,9 +343,8 @@ def run_submission(challenge_id, challenge_phase, submission, user_annotation_fi
...
                 # get split_code_name that is the key of the result
-                split_code_name = split_result.items()[0][0]
+                split_code_name = list(split_result.keys())[0]
```
taranjeet: I am not able to understand why is this change being done @RishabhJain2018 
As per my understanding, both will give the same result. Any warnings being thrown by Python3

apps/challenges/migrations/0019_added_leaderboard_data_table.py
```diff
@@ -10,7 +10,7 @@
 class Migration(migrations.Migration):
```
taranjeet: @RishabhJain2018 : Can you give me some more context about this PR ? Why is the migration required and what is the current status ?

RishabhJain2018: While I was trying to give support for the challenges that use python3, I encountered the following error of migrations:
screen shot 2018-07-09 at 7 31 15 pm
Then I researched on it and found that
- *Issue*: 
1. Python 3 detects the missing migration file in jobs file. (https://github.com/javrasya/django-river/issues/30) whereas Python2 doesn’t.
*Fixes Tried:*
1. I tried squashing all the migration files without manually changing the file(defected migration file) then it wasn’t getting squashed as it is throwing Node Not Found Error.
2. I tried squashing all the files by changing the migration file(defected migration file) manually but it created more dependency errors in the other apps.
3. I tried changing the defected migration file manually and it works like a charm. I know we should not change the migration file manually but in emergency we can do it (I think we can go with this ) (https://www.algotech.solutions/blog/python/django-migrations-and-how-to-manage-conflicts/, https://docs.djangoproject.com/en/1.10/howto/writing-migrations/).

@taranjeet Please suggest a more cleaner way to move ahead with this.

scripts/workers/submission_worker.py
RishabhJain2018: I have used the PR #1681 and built upon it.

isht3:
Tested this on docker and it seems to be working as intended. Can someone share the Visual Dialog Challenge config so I can test this with that too?

RishabhJain2018:
@deshraj I am merging it and testing it on the staging server. If there will be an error then I will revert it :)

PR #1742, opened 2018-07-07
Fix wrong number of submissions displayed in challenge analytics section

### LINKED ISSUE #1741: Challenges: Wrong number of submissions displayed in challenge analytics section
Currently, the count of submissions for each phase in a challenge is wrong in the Challenge Analytics section. It looks like there is something off with the API returning response.

RishabhJain2018:
Fix #1741 . @deshraj Please review the PR.

### DIFF
--- a/apps/analytics/serializers.py
+++ b/apps/analytics/serializers.py
@@ -3,14 +3,16 @@ from rest_framework import serializers
 
 class ChallengePhaseSubmissionCount(object):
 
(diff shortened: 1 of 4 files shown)


RishabhJain2018:
Hmm...I get your point. But I did this to avoid the different network calls for every feature and also it will reduce the query of the same thing upon the database.

RishabhJain2018:
Since we converged for a new API endpoint. Hence I am closing this PR.

PR #1743, opened 2018-07-09
Settings: Allow all hostnames to request files in dev env

deshraj:
## Current Scenario 

While creating a challenge using docker in dev environment, django server throws 400 BAD REQUEST as the response code when worker tries to fetch challenge related files and throws the following error:

```
Traceback (most recent call last):
  File "scripts/workers/submission_worker.py", line 303, in run_submission
    submission_output = EVALUATION_SCRIPTS[challenge_id].evaluate(
AttributeError: 'module' object has no attribute 'evaluate'
```

This PR fixes the above mentioned error. Thanks @abhiskk for mentioning this issue. Please feel free to comment if you still face the issue after this merge.

### DIFF
--- a/settings/dev.py
+++ b/settings/dev.py
@@ -5,6 +5,8 @@ from .common import *  # noqa: ignore=F405
 
 DEBUG = True
 
+ALLOWED_HOSTS = ["*"]
+
 DATABASES = {
     'default': {
         'ENGINE': 'django.db.backends.postgresql_psycopg2',


RishabhJain2018:
@deshraj I think the above mentioned issue is because the evaluation script details are present inside a folder in `/tmp` directory.
I don't think `ALLOWED_HOSTS` is related to this.

deshraj:
> I think the above mentioned issue is because the evaluation script details are present inside in /tmp/random_folder_name/challenge_id/evaluation_script directory.

@RishabhJain2018 I didn't get you on this. Can you please elaborate more on this?

RishabhJain2018:
Hmm...this issue occurs when we run the submission worker and it is not able to find the `evaluate` method in the `main.py` file.
The possible reason for not finding the `evaluate` method is that the evaluation script details are present inside `/tmp/random_folder_name/challenge_id/evaluation_script` directory. (This is the mistake that most challenge hosts do).

deshraj:
Yeah I agree what you are saying but here the error is different. It's due to getting 400 BAD REQUEST when the worker tries to fetch files from the django container. I tried on my local machine with our EvalAI-Starters repository and faced this issue. Some other challenge hosts have also reported similar issues.

deshraj:
Merging it for now since this is critical. We can revisit this in future.

PR #1744, opened 2018-07-11
Restrict API to give results for only participant team

RishabhJain2018:
This PR will restrict the `get_challenge_phase_submission_count_by_team` function to give the results for a participant team only.

### DIFF
--- a/apps/analytics/views.py
+++ b/apps/analytics/views.py
@@ -108,11 +108,9 @@ def get_submission_count(request, challenge_pk, duration):
 @api_view(['GET', ])
 @permission_classes((permissions.IsAuthenticated, HasVerifiedEmail, IsChallengeCreator))
 @authentication_classes((ExpiringTokenAuthentication,))
-def get_challenge_phase_submission_analysis(request, challenge_pk, challenge_phase_pk):
+def get_challenge_phase_submission_count_by_team(request, challenge_pk, challenge_phase_pk):
     """
-    API to fetch
-    1. The submissions count for challenge phase.
-    2. The participated team count for challenge phase.
+    API to fetch the submissions count in a challenge phase for a participant team.
     """
     challenge = get_challenge_model(challenge_pk)
 
@@ -122,14 +120,10 @@ def get_challenge_phase_submission_analysis(request, challenge_pk, challenge_pha
 
     submissions = Submission.objects.filter(
         challenge_phase=challenge_phase, challenge_phase__challenge=challenge, participant_team=participant_team)
-    submission_count = submissions.count()
-    submissions = Submission.objects.filter(
-        challenge_phase=challenge_phase, challenge_phase__challenge=challenge)
-    participant_team_count = submissions.values_list(
-        'participant_team', flat=True).distinct().count()
+    participant_team_submission_count = submissions.count()
 
     challenge_phase_submission_count = ChallengePhaseSubmissionCount(
-        submission_count, participant_team_count, challenge_phase.pk)
+        participant_team_submission_count, challenge_phase.pk)
     try:
         serializer = ChallengePhaseSubmissionCountSerializer(challenge_phase_submission_count)
         response_data = serializer.data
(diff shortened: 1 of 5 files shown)


apps/analytics/views.py
```diff
@@ -108,11 +108,9 @@ def get_submission_count(request, challenge_pk, duration):
 @api_view(['GET', ])
 @permission_classes((permissions.IsAuthenticated, HasVerifiedEmail, IsChallengeCreator))
 @authentication_classes((ExpiringTokenAuthentication,))
-def get_challenge_phase_submission_analysis(request, challenge_pk, challenge_phase_pk):
+def get_challenge_phase_submission_count_by_team(request, challenge_pk, challenge_phase_pk):
     """
-    API to fetch
-    1. The submissions count for challenge phase.
-    2. The participated team count for challenge phase.
+    API to fetch the submissions count in a challenge phase for a participant team.
```
deshraj: Please replace this with `Returns number of submissions done by a participant team in a challenge phase`

tests/unit/analytics/test_views.py
```diff
@@ -429,35 +429,33 @@ def test_get_challenge_phase_submission_analysis_when_challenge_phase_does_not_e
...
 
-        self.url = reverse_lazy('analytics:get_challenge_phase_submission_analysis',
+        self.url = reverse_lazy('analytics:get_challenge_phase_submission_count_by_team',
                                 kwargs={'challenge_pk': self.challenge.pk,
                                         'challenge_phase_pk': self.challenge_phase.pk})
 
         expected = {
-                "submission_count": 3,
-                "participant_team_count": 2,
+                "participant_team_submission_count": 3,
```
deshraj: The number `3` should not be hardcoded here. It should be fetched from the model. Something like `participant_team_1.submissions.count()`

tests/unit/analytics/test_views.py
```diff
@@ -429,35 +429,33 @@ def test_get_challenge_phase_submission_analysis_when_challenge_phase_does_not_e
...
 
-        self.url = reverse_lazy('analytics:get_challenge_phase_submission_analysis',
+        self.url = reverse_lazy('analytics:get_challenge_phase_submission_count_by_team',
                                 kwargs={'challenge_pk': self.challenge.pk,
                                         'challenge_phase_pk': self.challenge_phase.pk})
 
         expected = {
-                "submission_count": 1,
-                "participant_team_count": 2,
+                "participant_team_submission_count": 1,
```
deshraj: Please don't hardcode `1` here as well.

apps/analytics/views.py
```diff
@@ -122,14 +120,10 @@ def get_challenge_phase_submission_analysis(request, challenge_pk, challenge_pha
...
-    submission_count = submissions.count()
-    submissions = Submission.objects.filter(
-        challenge_phase=challenge_phase, challenge_phase__challenge=challenge)
-    participant_team_count = submissions.values_list(
-        'participant_team', flat=True).distinct().count()
+    participant_team_submission_count = submissions.count()
 
     challenge_phase_submission_count = ChallengePhaseSubmissionCount(
-        submission_count, participant_team_count, challenge_phase.pk)
+        participant_team_submission_count, challenge_phase.pk)
```
deshraj: Please rename the variable to `participant_team_submissions`

tests/unit/analytics/test_views.py
RishabhJain2018: Hmm..I did this intentionally so that it will be easy in future to get a sense of the test case as there are two test cases one in which `participant_team_submission_count` is 3 and other one in which `participant_team_submission_count` is 1. Should I change it back?

tests/unit/analytics/test_views.py
RishabhJain2018: same reason as above.

tests/unit/analytics/test_views.py
deshraj: The reason why I am saying to remove hard coding is that let's say sometime in future we change the number of submissions done by a particular team in the tests, then your all tests will break for no reason which should not happen.

RishabhJain2018: yes! Agreed. 👍

PR #1748, opened 2018-07-12
Modify condition for showing remaining submissions

### LINKED ISSUE #1746: Bug in showing remaining submissions
Currently, if the number of total remaining submissions is less than that of the limit set for doing the total number of submissions in a day, then ideally total submissions that can be done should be equal to the total remaining submissions which is not the case right now. Please see the screenshot below for illustration:

screen shot 2018-07-11 at 10 22 14 pm

RishabhJain2018:
Fix #1746 @deshraj Please review the PR.


tests/unit/jobs/test_views.py
```diff
@@ -481,6 +481,25 @@ def test_get_remaining_submission_when__submission_made_today(self):
...
+    def test_get_remaining_submission_when_today_submissions_is_more_than_max_submissions(self):
+        self.url = reverse_lazy('jobs:get_remaining_submissions',
+                                kwargs={'challenge_phase_pk': self.challenge_phase.pk,
+                                        'challenge_pk': self.challenge.pk})
+        setattr(self.challenge_phase, 'max_submissions_per_day', 9)
+        setattr(self.challenge_phase, 'max_submissions', 5)
+        self.challenge_phase.save()
+
+        expected = {
+            'remaining_submissions_today_count': 3,
```
deshraj: No hard coding please. :(

apps/jobs/views.py
```diff
@@ -346,6 +346,10 @@ def get_remaining_submissions(request, challenge_phase_pk, challenge_pk):
         # calculate the remaining submissions from total submissions.
         remaining_submission_count = max_submissions_count - \
             (submissions_done_count - failed_submissions_count)
+
+        if remaining_submissions_today_count >= remaining_submission_count:
```
deshraj: Just check for `>`

deshraj: @RishabhJain2018 can you please incorporate this change as well?

PR #1749, opened 2018-07-12
Frontend: Bug fix related to submission count

deshraj:


deshraj:
My bad. Seems like a cache issue on my machine. Already fixed in #1744

PR #1752, opened 2018-07-25
Replace RabbitMQ with SQS

deshraj:
This PR involves the following changes:

- [x] Replace RabbitMQ with SQS
- [x] Add ElasticMQ to mock SQS on the local machine
- [x] Update the docker settings
- [x] Add ElasticMQ in dev docker
- [x] Fix the seed script
- [x] Update README and Docs with the updated installation instructions
- [x] Remove RabbitMQ dependency in CI and add ElasticMQ in CI
- [x] Handle the case of gracefully killing the submission worker (see https://stackoverflow.com/questions/18499497/how-to-process-sigterm-signal-gracefully?answertab=active#tab-top)
- [x] Modify tests to work with the new codebase


deshraj:
@RishabhJain2018: Can you please test this on your local machine? 

@isht3: you should try this on your machine and try to understand it since this will be a part of your next sprint.

requirements/worker.txt
```diff
@@ -0,0 +1,4 @@
+numpy==1.14.5
+networkx==2.1
```
RishabhJain2018: @deshraj This must on line 1 😛

scripts/workers/submission_worker.py
```diff
@@ -299,6 +286,7 @@ def run_submission(challenge_id, challenge_phase, submission, user_annotation_fi
     submission.save()
     try:
         successful_submission_flag = True
+        print(EVALUATION_SCRIPTS)
```
RishabhJain2018: We should replace it with logger statements.

RishabhJain2018:
@isht3 No, you don't need the credentials for this. :)

isht3:
Also, I added Logger statements to wherever necessary. :+1:

PR #1755, opened 2018-07-30
Leaderboard sort fix

live-wire:
This fixes the challenge-leaderboard sorting issue (while switching sort between different columns).

cc: @RishabhJain2018 @deshraj


live-wire:
Turns out the issue was because I was checking for null index like this:

`if (!index)`
This also holds true when `index = 0` 🙈 
@RishabhJain2018

PR #1766, opened 2018-08-14
Add Timezone while displaying time in challenges

RishabhJain2018:
This PR will add the timezone reference in all the places where we show time in EvalAI.

@deshraj @isht3 @live-wire Please review the PR?
screen shot 2018-08-14 at 6 51 22 pm


frontend/src/js/app.js
```diff
@@ -13,5 +13,5 @@ angular
 		'toaster',
 		'angularTrix',
 		'angularMoment',
-		'ngclipboard'
+		'ngclipboard',
```
deshraj: Why there is a `,` with the last element in the array. I am sure this will cause error.

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -343,6 +343,12 @@
             onSuccess: function(response) {
                 var details = response.data;
                 vm.phases = details;
+                // console.log(vm.phases);
```
deshraj: Please remove `console.log()`

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -343,6 +343,12 @@
...
+                for (var i=0; i<vm.phases.results.length; i++) {
+                    vm.phases.results[i].start_date = strftime('%b %d, %Y %H:%M:%S %Z', new Date(vm.phases.results[i].start_date));
```
deshraj: Instead of showing the full timezone name can we show the short form such as EST, IST etc.

deshraj:
Also, why the `Random challenge` has start date in EDT timezone and end date in `EST` timezone. It looks a lot weird.

frontend/src/js/controllers/challengeCtrl.js
RishabhJain2018: I was also looking for the same but didn't found anything simpler than this. I will see if I can reduce it to a short form.

RishabhJain2018:
Hmm...I think it is related to Daylight saving time when it started there wasn't any Daylight saving and at the time of the challenge deadline, it is there.

deshraj:
If this is the case then sounds good but you might want to change the dates and cross check again.

RishabhJain2018:
Please take a look at the Visual Dialog, VizWiz and VLN challenge dates :)

deshraj:
@RishabhJain2018 were you able to shorten the timezone names to EDT, PST etc? If yes, can you please paste a screenshot here

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -343,6 +343,11 @@
             onSuccess: function(response) {
                 var details = response.data;
                 vm.phases = details;
+                for (var i=0; i<vm.phases.results.length; i++) {
+                    vm.phases.results[i].start_date = strftime('%b %d, %Y %H:%M:%S %Z', new Date(vm.phases.results[i].start_date));
+                    vm.phases.results[i].end_date = strftime('%b %d, %Y %H:%M:%S %Z', new Date(vm.phases.results[i].end_date));
+
```
deshraj: Don't need this newline. Please remove.

bower.json
```diff
@@ -31,7 +31,8 @@
     "angular-trix": "^1.0.2",
     "angular-moment": "^1.2.0",
     "moment": "^2.21.0",
-    "ngclipboard": "^2.0.0"
+    "ngclipboard": "^2.0.0",
```
deshraj: Since you are adding new dependencies, it would be great if you can drop your node_modules folder and try building it again for the sanity check.

deshraj:
@RishabhJain2018 can you please update this PR?

deshraj:
Although I have approved this PR, it would be great if we can find a way to show EST instead of full `Eastern Standard Time` etc for better display on the UI.

deshraj:
@RishabhJain2018 I can still see `Eastern Standard Time` instead of `EST`. Can you please fix it? I think you should be able to deal with it using `moment()` in JS.

RishabhJain2018:
Closing this PR.

PR #1774, opened 2018-08-22
Convert string time to integer time to avoid NaN.

RishabhJain2018:


frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -1026,7 +1026,7 @@
                             vm.message = details;
                             vm.showClock = true;
                             vm.countDownTimer = function() {
-                                vm.remainingTime = vm.message.remaining_time;
+                                vm.remainingTime = Math.floor(vm.message.remaining_time);
```
deshraj: Just curious: When the string is not a number then `Math.floor()` will return 0 or something else?

RishabhJain2018: My bad, it will return NaN only.

RishabhJain2018:
Closing this since I am unable to reproduce it. Will revisit it later.

PR #1778, opened 2018-08-24
Fix #1543: Added password errors due to spaces in signup module

### LINKED ISSUE #1543: Issue due to spaces in password while signing up
While signing up (new user) for EvalAI, I was required to set up a password for my account. The problem here is whether to allow spaces in password or not. While the frontend javascript checks donot check the presence or absence of spaces therefore a password like "abc abc " is accepted, but upon submitting, it shows that password is shorter than 8 characters, because of trimming of spaces at backend. This seems like an ambiguous message for the user.
ISSUE 2: Create a password of 8 spaces i.e. "        " and try setting it as password, there will again be an ambiguity with the error message shown.
ISSUE 3: There is a lower limit check on size of password, and no upper limit check on size of password, which may cause password overflow error, if user tries to enter a very long password to compromise with the system.
I can fix these both issues to display appropiate message to user, if this issue is accepted by the community.

parth-verma:
The given PR contains the following fixed which should resolve #1543:
- Added a max length of 32 to password fields
- Added descriptive error messages under each input field as returned by backend unlike earlier version where all the messages were shown at the same spot thus creating confusion as to which input field the error corresponds to.
- The `django-rest-auth` module trims whitespaces in passwords, therefore, I have created a PR for the same in the relevant repo and meanwhile have updated the requirements to correspond to the git repo which contains the fixed code, which can be then changed back to the pypi package once the package is updated.


party98:
@isht3 Please help me with this. The travis build seems to be failing on account of some `programming error`. Namely, it complains that the relation `django_site` doesn't exist. The same error occurs when running the tests locally on the PR as well as the `master` branch.

isht3:
@party98 Are you running the docker setup? Were the migrations successfully done?

requirements/common.txt
```diff
@@ -5,7 +5,6 @@ django-import-export==0.5.1
 djangorestframework==3.5.2
 djangorestframework-expiring-authtoken==0.1.4
 django-cors-headers==1.3.1
-django-rest-auth[with_social]==0.9.2
```
RishabhJain2018: @party98 Can you please tell me why you have removed this file?

party98: This package handles password authentication. During this process, it strips the password before processing, thus removing spaces at the beginning and end of the password input and thereby mutating the password. As you should be able to see, I have replaced this with a fork of the same package which does not remove the spaces. 
I have also made a PR for the given package containing the same fix.

RishabhJain2018: @party98 So, how does this explains that Travis fail is due to it?

party98: Travis should not be failing due to this, as the replaced package is the fork of the same package with just one line changed

party98: I'll try to look into the cause of the failure.

deshraj:
Closing this due to inactivity. Please feel free to reopen it if you are still working on this.

PR #1792, opened 2018-09-06
Add commands to build and push django container to AWS ECR

RishabhJain2018:


requirements/dev.txt
```diff
@@ -9,6 +9,7 @@ django-extensions==1.7.8
 django-silk==1.0.0
 django-spaghetti-and-meatballs==0.2.2
 flake8==3.0.4
+google-compute-engine==2.8.3
```
deshraj: Why do we need this?

settings/test.py
```diff
@@ -22,4 +22,5 @@
     }
 }
 
+DEV = False
```
deshraj: We don't need a new variable `DEV`. 

`settings.DEBUG = True` means dev environment only.

requirements/dev.txt
RishabhJain2018: Earlier it threw that this module isn't found. I am going to remove this line only.

settings/test.py
RishabhJain2018: Please check the condition  https://github.com/Cloud-CV/EvalAI/pull/1792/files#diff-aa6d983093bc40e4ce5d8205f511db8dR78

deshraj: This variable will cause confusion for no reason. We already have the `TEST` variable which you can make use of in the code snippet.

evalai/urls.py
```diff
@@ -75,7 +75,7 @@
                ]
 
 # DJANGO-SPAGHETTI-AND-MEATBALLS URLs available during development only.
-if settings.DEBUG:
+if settings.DEBUG or not settings.TEST:
```
deshraj: I am not sure if we need an additional check here.

scripts/deploy/production.sh
```diff
@@ -0,0 +1,17 @@
+BRANCH="production"
```
deshraj: We don't need this script now, right?

.travis.yml
```diff
@@ -39,7 +45,9 @@ script:
 
 after_success:
   - coveralls --rcfile=.coveragerc
-
+  - pip install awscli
+  - bash scripts/deploy/staging.sh
+  - bash scripts/deploy/production.sh
```
deshraj: We dont need the two scripts now, right?

docker-compose-staging.yml
```diff
@@ -0,0 +1,76 @@
...
+
+  worker:
+    image: ${AWS_ACCOUNT_ID}.dkr.ecr.us-east-1.amazonaws.com/evalai-staging-worker:${TRAVIS_COMMIT}
+    build:
+      context: ./
+      dockerfile: docker/prod/worker/Dockerfile
+    depends_on:
+      - django
+    environment:
+      # TODO: Sort these entries
```
deshraj: Please remove this.

docker-compose-staging.yml
```diff
@@ -0,0 +1,76 @@
+version: "3"
+services:
+
+  django:
+    image: ${AWS_ACCOUNT_ID}.dkr.ecr.us-east-1.amazonaws.com/evalai-staging-backend:${TRAVIS_COMMIT}
+    container_name: django
+    environment:
+      # TODO: Sort these entries
```
deshraj: Please remove this line.

scripts/deploy/deploy.sh
```diff
@@ -0,0 +1,35 @@
+if [ "$TRAVIS_PULL_REQUEST" != "false" -o "$TRAVIS_BRANCH" != "staging" -o "$TRAVIS_BRANCH" != "production" ]; then
+    echo "Skipping deploy to staging or production server; The request or commit is not on staging or production branch"
+    exit 0
+fi
+
+if [ "$TRAVIS_PULL_REQUEST" != "false" -a "TRAVIS_BRANCH" != "staging"]; then
```
deshraj: Please modify line 6-30 and put them in a function(). The logic could be something like this

```
if branch == staging:
    deploy to evalai-staging-* ECR repos with staging credentials
else if branch == production:
   deploy to evalai-* ECR repos with prod credentials
```

PR #1804, opened 2018-09-12
Add Feature to view Private leaderboard

RishabhJain2018:
Also, @isht3 Can you please test it on your local machine?


RishabhJain2018:
screen shot 2018-09-12 at 12 14 48 am
screen shot 2018-09-12 at 12 14 33 am

apps/jobs/views.py
```diff
@@ -266,8 +262,23 @@ def leaderboard(request, challenge_phase_split_id):
                                                 distinct_sorted_leaderboard_data,
                                                 request,
                                                 pagination_class=StandardResultSetPagination())
-    response_data = result_page
-    return paginator.get_paginated_response(response_data)
+
+    challenge_host_user = is_user_a_host_of_challenge(request.user, challenge_obj.pk)
+
+    # Check if the user is challenge host or not
+    if challenge_host_user is True:
```
deshraj: Change this condition to `if challenge_host_user` only :)

apps/jobs/views.py
```diff
@@ -266,8 +262,23 @@ def leaderboard(request, challenge_phase_split_id):
                                                 distinct_sorted_leaderboard_data,
                                                 request,
                                                 pagination_class=StandardResultSetPagination())
-    response_data = result_page
-    return paginator.get_paginated_response(response_data)
+
+    challenge_host_user = is_user_a_host_of_challenge(request.user, challenge_obj.pk)
+
+    # Check if the user is challenge host or not
```
deshraj: This comment is useless. Instead write something like `Show the Private leaderboard only if the user is a challenge host`

apps/jobs/views.py
```diff
@@ -266,8 +262,23 @@ def leaderboard(request, challenge_phase_split_id):
...
+    if challenge_host_user is True:
+        response_data = result_page
+        return paginator.get_paginated_response(response_data)
+
+    # Check if challenge phase leaderboard is public for participant user or not
+    elif challenge_phase_split.visibility != ChallengePhaseSplit.PUBLIC:
+        response_data = {'error': 'Sorry, leaderboard is not public yet for this Challenge Phase Split!'}
```
deshraj: Please change the error to `Sorry, the leaderboard is not public`

apps/jobs/views.py
```diff
@@ -266,8 +262,23 @@ def leaderboard(request, challenge_phase_split_id):
...
+        return Response(response_data, status=status.HTTP_400_BAD_REQUEST)
+
+    # If the leaderboard is public then display it
```
deshraj: This is an obvious comment. Please remove it.

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -365,6 +365,12 @@
             onSuccess: function(response) {
                 var details = response.data;
                 vm.phaseSplits = details;
+                for(var i=0; i<details.length; i++) {
+                    if (details[i].visibility !== 3) {
```
deshraj: Checking directly with number can be confusing at times. Can you please check with JS object? 

Something like:

```
var challengePhaseVisibility = {
    PUBLIC: 3,
    PRIVATE: 2,
    ABC: 1
}
if details[i].visibility == challengePhaseVisibility.PUBLIC
```

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -365,6 +365,12 @@
...
+                        vm.phaseSplits[i].showPrivate = true;
```
deshraj: I am not sure what this line is doing. Can you please tell what it will exactly do?

frontend/src/views/web/challenge/leaderboard.html
```diff
@@ -22,9 +22,11 @@
...
-                            <span><md-select ng-model="challenge.phaseName" placeholder="Challenge Phase" class="rm-margin">
-                            <md-option ui-sref="web.challenge-main.challenge-page.phase-leaderboard({phaseSplitId:key.id})" value="{{key.id}}" ng-repeat="key in challenge.phaseSplits" ng-show="{{key.visibility}}==3"><span class="font-weight-b">Phase: </span>{{key.challenge_phase_name}}, <span class="font-weight-b">Split: </span> {{key.dataset_split_name}}</md-option>
-                        </md-select></span>
+                            <span>
+                                <md-select ng-model="challenge.phaseName" placeholder="Challenge Phase" class="rm-margin">
+                                    <md-option ui-sref="web.challenge-main.challenge-page.phase-leaderboard({phaseSplitId:key.id})" value="{{key.id}}" ng-repeat="key in challenge.phaseSplits" ng-show="{{key.visibility}}"><span class="font-weight-b">Phase: </span>{{key.challenge_phase_name}}, <span class="font-weight-b">Split: </span> {{key.dataset_split_name}} &nbsp; <span class="chip" ng-show="{{key.showPrivate}}">{{key.text}}</span></md-option>
```
deshraj: Can you please use badges instead of chips here? 

see https://materializecss.com/badges.html#!

frontend/src/js/controllers/challengeCtrl.js
RishabhJain2018: It is used to decide when to show `Private` in the challenge host leaderboard. Please refer here: https://github.com/Cloud-CV/EvalAI/pull/1804/files#diff-c09876db1a2167c9371dbcccf9a69aebR27

frontend/src/views/web/challenge/leaderboard.html
RishabhJain2018: Sure!

frontend/src/js/controllers/challengeCtrl.js
deshraj: Do we really have to store it as a string `Private`. Can we please not show using `ng-if` which will be much neater.

PR #1824, opened 2018-09-27
Fix #1821: Fixes download token button for Firefox

### LINKED ISSUE #1821: Download JSON  is not working in API Authentication Token pop-up box
* the observed bevaviour 

in user profile  **Download JSON** is not working
after clicking it doesn't give any response

screenshot from 2018-09-24 00-44-03

parth-verma:
Fixes #1821. Changed the way token is downloaded as the `click` event that was used earlier seems to be [incompatible](https://stackoverflow.com/questions/24673612/element-click-does-not-work-in-firefox-and-ie-but-works-in-chrome)  with Firefox.


deshraj:
I think this issue is already fixed in #1991

PR #1832, opened 2018-10-04
Fix #1622 Add response argument in onError functions in controllers for frontend

### LINKED ISSUE #1622: Solve inconsistencies in the response handling for the front-end and back-end.
# Current Scenario
Currently, there's a lot of inconsistencies in handling the various response messages on the front-end and on the back-end. 

* Error responses on controllers don't take in the response argument. `onError: function()` --> `onError: function(response)`
* Responses are not properly displayed. E.g `['Server error']` Which is vague and the user gets no feedback.
* Some responses are displayed in a list, which is a strike for the UI.

# Deliverables
* Try to make the success and the error responses from the back-end to be consistent with the Django and all-auth type responses.
* Make the success and error message display in the front-end to be clean and consistent all throughout the application.

nagpalm7:
Fix #1622 : Add response argument in onError functions in controllers for frontend wherever possible @isht3

### DIFF
--- a/frontend/src/js/controllers/challengeCtrl.js
+++ b/frontend/src/js/controllers/challengeCtrl.js
@@ -1038,9 +1038,10 @@
                         }
                     }
                 },
-                onError: function() {
+                onError: function(response) {
+                    var details = response.data;
                     vm.stopLoader();
-                    $rootScope.notify("error", "Some error occured. Please try again.");
+                    $rootScope.notify("error", details.error);
                 }
             };
             utilities.sendRequest(parameters);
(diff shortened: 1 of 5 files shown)


nagpalm7:
@isht3 please review the PR #1622

sudo
```diff
@@ -0,0 +1 @@
+151.101.4.162 registry.npmjs.org /etc/hosts
```
deshraj: I don't think you need to add this. :)

nagpalm7: I couldn't find this in the diff. Can you tell me the file name. If its related to something else can you explain please!

nagpalm7:
@deshraj i have resolved the issue. Please review the changes

Issue #1833, opened 2018-10-04
Remove discussions tab from challenge page


RishabhJain2018:
Currently the discussions tab is always shown whether a link in it exists or not but we would like to show the discussions tab only when there is a link in it from the backend.

nagpalm7:
I want to work on this issue!!

RishabhJain2018:
Sure! Go ahead.

spyshiv:
ytdytfy
![asset-37-4-x 3x](https://user-images.githubusercontent.com/7015220/47180574-7d64ef00-d33e-11e8-9abd-5511f16e143f.png)

hbuyb
lmnijnui

PR #1835, opened 2018-10-05
hide discussions tab if link is null from backend

nagpalm7:
#1833  Remove discussions tab from challenge page
- [x] Hide discussions tab from challenge page if link is null from backend @RishabhJain2018


nagpalm7:
Thanks

PR #1847, opened 2018-10-18
Fix #1782: Corrected improper API Naming for Safari 

### LINKED ISSUE #1782: Unable to login via Safari
### The expected behaviour
Successfully be able to login to the dashboard on Safari

### The observed behaviour
Unable to maintain a session. After 5 seconds of logging in, a timeout error is generated and the site logouts automatically.

### Steps to reproduce the problem
Try logging in via safari

### Which operating system and/or browser you're using, if relevant
macOS 10.13.6
Safari 11.1.2

### (if possible) copies of the console/server logs
[Error] Failed to load resource: the server responded with a status of 401 (Unauthorized) (challenge_host_team, line 0)

image

vkartik97:
Closes #1782 
screen shot 2018-10-18 at 8 57 00 pm
 Safari Version 11.1.2 (13605.3.8)

### DIFF
--- a/frontend/src/js/controllers/dashCtrl.js
+++ b/frontend/src/js/controllers/dashCtrl.js
@@ -91,7 +91,7 @@
         utilities.sendRequest(parameters);
 
         //check for host teams.
-        parameters.url = 'hosts/challenge_host_team';
+        parameters.url = 'hosts/challenge_host_team/';
         parameters.method = 'GET';
         parameters.token = userKey;
         parameters.callback = {
--- a/settings/common.py
+++ b/settings/common.py
@@ -139,6 +139,8 @@ USE_L10N = True
 
 USE_TZ = True
 
+APPEND_SLASH = False
+
 # Static files (CSS, JavaScript, Images)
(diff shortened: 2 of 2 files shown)


RishabhJain2018:
I appreciate that you almost resolved the issue but can you please try looking for the fix on safari version 12(latest version) as it doesn't work on it.
Please keep me posted regarding it.

deshraj:
I am not sure if this is a perfect solution but if it is working as of now, we can merge it. 

@vkartik97 can you please share some resource that justifies this as the solution to this problem.
Also, @RishabhJain2018 can you please check that all the other REST APIs are working after doing this change on all the browsers (Chrome, Mozilla, and Safari)?

RishabhJain2018:
Sure! I'll test it thoroughly once @vkartik97 fixes it.

vkartik97:
@deshraj 
https://stackoverflow.com/questions/22696560/angularjs-authorization-header-not-added-in-safari

RishabhJain2018:
@vkartik97 Please update it so that we can merge it.

PR #1853, opened 2018-10-21
Fix #1850: Made execution time show only if greater than 0

### LINKED ISSUE #1850: Display execution time only when greater than zero.
Currently, all the types of execution time is shown on the frontend under `my submissions`. We would like to show only the execution time which is greater than zero.

Helpful link: https://stackoverflow.com/questions/18403835/how-to-make-less-than-or-greater-than-comparison-in-angularjs?answertab=active#tab-top

dzpzee:
In "my submissions" page

#1850


frontend/src/views/web/challenge/my-submission.html
```diff
@@ -92,7 +92,7 @@
                                             <td ng-if="key.publication_url == ''"> None </td>
 
                                             <td class="val-style capitalize" ng-class="key.status">{{key.status}}</td>
-                                            <td class="align-center">{{key.execution_time}}</td>
+                                            <td class="align-center" ng-show="key.execution_time > 0">{{key.execution_time}}</td>
```
RishabhJain2018: Please add a default value `None` which will be displayed when the execution time is < 0.

dzpzee: Done.

dzpzee: Done.

dzpzee:
Ok. I think that worked.

RishabhJain2018:
LGTM :) Thanks for the fix!

PR #1862, opened 2018-10-24
Add API to populate leaderboard

RishabhJain2018:


apps/jobs/urls.py
```diff
@@ -17,4 +17,6 @@
         views.leaderboard, name='leaderboard'),
     url(r'^submission/(?P<submission_id>[0-9]+)$',
         views.get_submission_by_pk, name='get_submission_by_pk'),
+    url(r'^challenge/(?P<challenge_pk>[0-9]+)/submission/$',
```
deshraj: let's rename the URL to `/api/jobs/challenge/<challenge_pk>/update_submission/`

apps/jobs/urls.py
```diff
@@ -17,4 +17,6 @@
...
+        views.submit_leaderboard_data, name='submit_leaderboard_data'),
```
deshraj: Change the function name to `update_submission`

PR #1880, opened 2018-11-04
Challenges: Add capability for pseudo remote evaluation

deshraj:
This PR adds the following capabilities:

- [x] Ability to do remote evaluation where the submission worker sends the evaluation job to other API endpoint hosted by challenge organizer
- [x] Update the Challenge model to support `remote_evaluation` as boolean field
- [x] Update ZipChallengeSerializer 
- [x] Update Challenge creation docs for the new field


scripts/workers/submission_worker.py
```diff
@@ -288,24 +288,34 @@ def run_submission(challenge_id, challenge_phase, submission, user_annotation_fi
...
     # main directory does not gets polluted
     temp_run_dir = join(submission_data_dir, 'run')
     create_dir(temp_run_dir)
 
-    stdout_file_name = 'temp_stdout.txt'
-    stderr_file_name = 'temp_stderr.txt'
-
-    stdout_file = join(temp_run_dir, stdout_file_name)
-    stderr_file = join(temp_run_dir, stderr_file_name)
+    stdout_file = join(temp_run_dir, 'temp_stdout.txt')
```
RishabhJain2018: Just to confirm, we are making this change so as to remove the unnecessary variables `stdout_file_name`, `stderr_file_name`, right?

scripts/workers/submission_worker.py
```diff
@@ -519,7 +529,6 @@ def main():
 
     while True:
         for message in queue.receive_messages():
-            # Print out the body of the message
```
RishabhJain2018: Why we are removing this comment in this PR?

scripts/workers/submission_worker.py
deshraj: Yep. These variables were defined for no reason.

scripts/workers/submission_worker.py
deshraj: This comment is pretty much useless. You can directly figure out what is going on in that line and don't need this comment.

RishabhJain2018: I know, just wanted to highlight it so that we won't wonder about it in future.
Rest Looks goo!

deshraj:
> @deshraj I think we also need to update the challenge creation using zip file API, if we are updating the documentation as the challenge hosts will try to add that key and challenge creation will fail.

I think we just need to add the field of `remote_evaluation` to the `ZipChallengeSerializer` to make the challenge creation API work with the new field. Isn't it the case? Moreover, since the default value is `False` for this field so it will take care if the field is not supplied in the config file. 

Please let me know if I am missing something here.

RishabhJain2018:
> I think we just need to add the field of remote_evaluation to the ZipChallengeSerializer to make the challenge creation API work with the new field. Isn't it the case?
True.
Looks good!

PR #1881, opened 2018-11-04
Fix #1879: Remove RabbitMQ from settings and docs

### LINKED ISSUE #1879: Remove RabbitMQ config settings
Since we have removed RabbitMQ as a message broker, we would also like to remove the settings for it in the `settings/common.py`.

deshraj:


docs/source/faq(developers).md
```diff
@@ -15,9 +15,9 @@ Django is the heart of the application, which powers our backend. We use Django
 
 We use Django Rest Framework for writing and providing REST APIs. It's permission and serializers have helped write a maintainable codebase.
 
-##### RabbitMQ
+##### SQS
```
RishabhJain2018: Can we please change it to `AWS Simple Queue Service`

PR #1882, opened 2018-11-05
Modify documentation for updating submission record API

RishabhJain2018:


apps/jobs/views.py
```diff
@@ -528,8 +528,18 @@ def get_submission_by_pk(request, submission_id):
                 description='Submission error file content'
             ),
             'result': openapi.Schema(
-                type=openapi.TYPE_STRING,
-                description='Submission results'
+                type=openapi.TYPE_OBJECT,
+                description='Submission results (The keys in this string must correspond to the leaderboard schema format. There can be n number of keys.)',
+                properties={
+                    'key 1': openapi.Schema(
```
deshraj: Let's rename the `key 1` to `metric-1` and similar for the `key 2`. It will become much easier in that case for someone to understand what the format is.

apps/jobs/views.py
```diff
@@ -528,8 +528,18 @@ def get_submission_by_pk(request, submission_id):
                 description='Submission error file content'
             ),
             'result': openapi.Schema(
-                type=openapi.TYPE_STRING,
-                description='Submission results'
+                type=openapi.TYPE_OBJECT,
+                description='Submission results (The keys in this string must correspond to the leaderboard schema format. There can be n number of keys.)',
```
deshraj: Replace `description='Submission results (The keys in this string must correspond to the leaderboard schema format. There can be n number of keys.)',` with 
`description='Submission results in dict format (Keys in this dict must correspond to the metrics in the leaderboard of challenge phase split. API will throw an error if any metric is missing)',`

deshraj: Also, just to confirm: Is it supposed to be array of dicts or just dict?

RishabhJain2018: It is just dict.

RishabhJain2018:
@deshraj here is the screenshot.
screenshot 2018-11-04 at 10 19 00 pm

apps/jobs/views.py
```diff
@@ -528,10 +528,45 @@ def get_submission_by_pk(request, submission_id):
...
-                description='Submission results'
-            )
-        }),
+                type=openapi.TYPE_ARRAY,
+                description='Submission results in array format'
+                ' (Keys in this dict must correspond to the challenge phase split.'
+                ' API will throw an error if any split and results metric is missing)',
+                items=openapi.Schema(
+                    type=openapi.TYPE_OBJECT,
+                    properties={
```
deshraj: Sorry for asking to change one more time. Can you please change the schema to 

```
[
  {
    "split": "split1",
    "accuracies": {
      "score": 90
    }
  },
  {
    "split": "split2",
    "accuracies": {
      "score": 50
    }
  }
]
```

PR #1885, opened 2018-11-05
Submissions: Make changes in update_submission API to allow multiple splits

deshraj:
This API adds the following functionalities:

- [x] Update permission check for whether the user is challenge host or not
- [x] Update checks for if challenge phase split exists or not
- [x] Make the transaction atomic
- [x] Handle more corner cases
- [x] Fix error return style
- [x] Minor typo fixes like `succes` to `success`


apps/jobs/views.py
```diff
@@ -538,66 +540,93 @@ def get_submission_by_pk(request, submission_id):
 })
 @throttle_classes([UserRateThrottle,])
 @api_view(['PUT',])
-@permission_classes((permissions.IsAuthenticated, HasVerifiedEmail, IsChallengeCreator,))
+@permission_classes((permissions.IsAuthenticated, HasVerifiedEmail,))
```
RishabhJain2018: Why have you removed `IsChallengeCreator` permission class from this API?

apps/jobs/views.py
```diff
@@ -538,66 +540,93 @@ def get_submission_by_pk(request, submission_id):
...
 @authentication_classes((ExpiringTokenAuthentication,))
 def update_submission(request, challenge_pk):
     """
     API endpoint to update submission realted attributes
     """
-    challenge_phase_split_pk = request.data['challenge_phase_split']
+    if not is_user_a_host_of_challenge(request.user, challenge_pk):
```
RishabhJain2018: This is similar to using `IsChallengeCreator`, right?

apps/jobs/views.py
```diff
@@ -538,66 +540,93 @@ def get_submission_by_pk(request, submission_id):
...
 
-        serializer = CreateLeaderboardDataSerializer(
-            data=request.data.copy(),
-            context={
-                'challenge_phase_split': challenge_phase_split,
-                'submission': submission,
-                'request': request,
-            }
-        )
+        leaderboar_data_list = []
```
RishabhJain2018: Can we please change `leaderboar_data_list` to `leaderboard_data_list`.

apps/jobs/views.py
```diff
@@ -538,66 +540,93 @@ def get_submission_by_pk(request, submission_id):
...
+            serializer = CreateLeaderboardDataSerializer(
+                data=request.data.copy(),
+                context={
+                    'challenge_phase_split': challenge_phase_split,
+                    'submission': submission,
+                    'request': request,
+                }
+            )
+            if serializer.is_valid():
+                leaderboar_data_list.append(serializer)
```
RishabhJain2018: Can we please change `leaderboar_data_list` to `leaderboard_data_list`.

apps/jobs/views.py
```diff
@@ -538,66 +540,93 @@ def get_submission_by_pk(request, submission_id):
...
+            else:
+                return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
-        if serializer.is_valid():
-            serializer.save()
-        else:
-            return Response(serializer.errors, status=status.HTTP_406_NOT_ACCEPTABLE)
+        try:
+            with transaction.atomic():
+                for serializer in leaderboar_data_list:
```
RishabhJain2018: Can we please change `leaderboar_data_list` to `leaderboard_data_list`.

apps/jobs/views.py
```diff
@@ -538,66 +540,93 @@ def get_submission_by_pk(request, submission_id):
 })
 @throttle_classes([UserRateThrottle,])
 @api_view(['PUT',])
-@permission_classes((permissions.IsAuthenticated, HasVerifiedEmail, IsChallengeCreator,))
+@permission_classes((permissions.IsAuthenticated, HasVerifiedEmail,))
 @authentication_classes((ExpiringTokenAuthentication,))
 def update_submission(request, challenge_pk):
     """
     API endpoint to update submission realted attributes
     """
```
RishabhJain2018: Can we please add a commented format of the data this API is going to receive, like what we have in submission worker. It will be helpful when we will look at this code in future.

apps/jobs/views.py
```diff
@@ -538,66 +542,117 @@ def get_submission_by_pk(request, submission_id):
...
+        except IntegrityError:
+            logger.exception('Failed to update submission_id {} related metadata'.format(submission_pk))
+            response_data = {'error': 'Failed to update submission_id {} related metadata'.format(submission_pk)}
+            return Response(response_data, status=status.HTTP_400_BAD_REQUEST)
 
     submission.status = submission_status
     submission.completed_at = timezone.now()
     submission.stdout_file.save('stdout.txt', ContentFile(stdout_content))
     submission.stderr_file.save('stderr.txt', ContentFile(stderr_content))
     submission.submission_result_file.save('submission_result.json', ContentFile(submission_result))
```
RishabhJain2018: @deshraj Just wanted to confirm, This will output all the data in the results into the `results.json` file which will be visible to participants.
As of now, we don't show participants their accuracies on a particular split but this will generate a file with content like
```
[{"split":"split2","accuracies":{"score":50}},{"split":"split1","accuracies":{"score":50}}]
```
into it. I think this isn't the expected bahaviour.

deshraj: Good catch Rishabh. I am glad that we caught it before it could have been deployed to any challenge. I will add a key `show_to_participant` in the API and that will take care of this.

apps/jobs/views.py
```diff
@@ -538,66 +542,117 @@ def get_submission_by_pk(request, submission_id):
...
+     - ``challenge_phase``: challenge phase id, e.g. 123 (**required**)
+     - ``submission``: submission id, e.g. 123 (**required**)
+     - ``stdout``: Stdout after evaluation, e.g. "Evaluation completed in 2 minutes" (**required**)
+     - ``stderr``: Stderr after evaluation, e.g. "Failed due to incorrect file format" (**required**)
+     - ``submission_status``: Status of submission after evaluation
+                (can take one of the following values: FINISHED/CANCELLED/FAILED), e.g. FINISHED (**required**)
+     - ``result``: contains accuracies for each metric, e.g. 
+                [
+                    {
+                        "split": "split1",
```
RishabhJain2018: One more thing, I forgot to mention that can we please change it like `"split": "split1 codename"`. It would be accurate and will avoid confusion.

apps/jobs/views.py
```diff
@@ -538,66 +542,117 @@ def get_submission_by_pk(request, submission_id):
...
+                        "accuracies": {
+                        "score": 90
+                        }
+                    },
+                    {
+                        "split": "split2",
```
RishabhJain2018: Same as above.

apps/jobs/views.py
RishabhJain2018: @deshraj What about this one? Can I go ahead and update it?

deshraj: Yeah, that's a good idea. I will do these changes.

PR #1895, opened 2018-11-09
Added new_auth_token url

gescalona1:
Backend progress for creating a new auth token endpoint.

### DIFF
--- a/apps/accounts/urls.py
+++ b/apps/accounts/urls.py
@@ -6,5 +6,5 @@ urlpatterns = [
 
     url(r'^user/disable$', views.disable_user, name='disable_user'),
     url(r'^user/get_auth_token$', views.get_auth_token, name='get_auth_token'),
-
+    url(r'^user/new_auth_token$', views.new_auth_token, name='new_auth_token'),
 ]
--- a/apps/accounts/urls.py
+++ b/apps/accounts/urls.py
@@ -3,8 +3,7 @@ from django.conf.urls import url
 from . import views
 
 urlpatterns = [
-
     url(r'^user/disable$', views.disable_user, name='disable_user'),
     url(r'^user/get_auth_token$', views.get_auth_token, name='get_auth_token'),
-    url(r'^user/new_auth_token$', views.new_auth_token, name='new_auth_token'),
+    url(r'^user/generate_auth_token$', views.generate_auth_token, name='generate_auth_token'),
 ]
--- a/apps/accounts/views.py
+++ b/apps/accounts/views.py
@@ -51,19 +51,21 @@ def get_auth_token(request):
 @api_view(['GET'])
 @permission_classes((permissions.IsAuthenticated, HasVerifiedEmail))
 @authentication_classes((ExpiringTokenAuthentication, ))
-def new_auth_token(request):
+def generate_auth_token(request):
     try:
         user = User.objects.get(email=request.user.email)
     except User.DoesNotExist:
         response_data = {"error": "This User account doesn't exist."}
         return Response(response_data, status.HTTP_404_NOT_FOUND)
-    oldtoken = Token.objects.get(user=user)
-    if oldtoken:
-        Token.objects.filter(user=user).delete()
-    token = Token.objects.create(user=user)
-    token.save()
+    try:
+        old_auth_token = Token.objects.get(user_id=request.user.pk)
+    except Token.DoesNotExist:
+        response_data = {"error": "This Token account doesn't exist."}
+        return Response(response_data, status.HTTP_404_NOT_FOUND)
+    Token.delete(old_auth_token)
+    new_token = Token.objects.create(user=user)
+    new_token.save()
     response_data = {
-        "oldtoken": "{}".format(oldtoken),
-        "token": "{}".format(token)
+        "message": "The new generated auth token for {} is {}".format(request.user.username, new_token),
     }
     return Response(response_data, status=status.HTTP_200_OK)
--- a/apps/accounts/views.py
+++ b/apps/accounts/views.py
@@ -45,3 +45,19 @@ def get_auth_token(request):
 
     response_data = {"token": "{}".format(token)}
     return Response(response_data, status=status.HTTP_200_OK)
+
+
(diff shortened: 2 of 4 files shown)


apps/accounts/urls.py
```diff
@@ -6,5 +6,5 @@
 
     url(r'^user/disable$', views.disable_user, name='disable_user'),
     url(r'^user/get_auth_token$', views.get_auth_token, name='get_auth_token'),
-
+    url(r'^user/new_auth_token$', views.new_auth_token, name='new_auth_token'),
```
RishabhJain2018: Please rename `new_auth_token` -> `generate_auth_token`

apps/accounts/views.py
```diff
@@ -45,3 +45,25 @@ def get_auth_token(request):
 
     response_data = {"token": "{}".format(token)}
     return Response(response_data, status=status.HTTP_200_OK)
+
+
+@throttle_classes([UserRateThrottle])
+@api_view(['GET'])
+@permission_classes((permissions.IsAuthenticated, HasVerifiedEmail))
+@authentication_classes((ExpiringTokenAuthentication, ))
+def new_auth_token(request):
```
RishabhJain2018: Please rename `new_auth_token` to `generate_auth_token`.

apps/accounts/views.py
```diff
@@ -45,3 +45,25 @@ def get_auth_token(request):
...
+    try:
+        user = User.objects.get(email=request.user.email)
```
RishabhJain2018: Please query user object using the `pk`.

apps/accounts/views.py
```diff
@@ -45,3 +45,25 @@ def get_auth_token(request):
...
+    except User.DoesNotExist:
+        response_data = {"error": "This User account doesn't exist."}
+        return Response(response_data, status.HTTP_404_NOT_FOUND)
+    oldtoken = Token.objects.get(user=user)
```
RishabhJain2018: `oldtoken`  -> `old_auth_token`

apps/accounts/views.py
```diff
@@ -45,3 +45,25 @@ def get_auth_token(request):
...
+    if oldtoken:
```
RishabhJain2018: Same as above.

apps/accounts/views.py
```diff
@@ -45,3 +45,25 @@ def get_auth_token(request):
...
+        Token.objects.filter(user=user).delete()
```
RishabhJain2018: There is no need to query again as you can delete the object you have obtained from the earlier query.

apps/accounts/views.py
RishabhJain2018: Also please add `get` query in a try catch because if the Token isn't found then it will throw an error.

apps/accounts/views.py
```diff
@@ -45,3 +45,25 @@ def get_auth_token(request):
...
+    token = Token.objects.create(user=user)
```
RishabhJain2018: `token` -> `new_token`

apps/accounts/views.py
```diff
@@ -45,3 +45,25 @@ def get_auth_token(request):
...
+    token.save()
```
RishabhJain2018: same as above.

apps/accounts/views.py
```diff
@@ -45,3 +45,25 @@ def get_auth_token(request):
...
+    response_data = {
```
RishabhJain2018: Please return the new token with below message.
`response_data = {'message' : 'The new generated auth token for {} is {}.format(request.user.username, new_token)'}`

apps/accounts/views.py
```diff
@@ -45,3 +45,25 @@ def get_auth_token(request):
...
+        "oldtoken": "{}".format(oldtoken),
+        "token": "{}".format(token)
+    }
+    return Response(response_data, status=status.HTTP_200_OK)
```
RishabhJain2018: Please add test cases for this API.

apps/accounts/views.py
```diff
@@ -45,3 +45,27 @@ def get_auth_token(request):
...
     return Response(response_data, status=status.HTTP_200_OK)
+
+
+@throttle_classes([UserRateThrottle])
+@api_view(['GET'])
+@permission_classes((permissions.IsAuthenticated, HasVerifiedEmail))
+@authentication_classes((ExpiringTokenAuthentication, ))
+def generate_auth_token(request):
+    try:
+        user = User.objects.get(email=request.user.email)
```
RishabhJain2018: Please query user using `pk`

apps/accounts/views.py
```diff
@@ -45,3 +45,27 @@ def get_auth_token(request):
...
+    except User.DoesNotExist:
+        response_data = {"error": "This User account doesn't exist."}
+        return Response(response_data, status.HTTP_404_NOT_FOUND)
+    try:
+        old_auth_token = Token.objects.get(user_id=request.user.pk)
+    except Token.DoesNotExist:
+        response_data = {"error": "This Token account doesn't exist."}
+        return Response(response_data, status.HTTP_404_NOT_FOUND)
+    Token.delete(old_auth_token)
+    new_token = Token.objects.create(user=user)
```
RishabhJain2018: Please add a try-catch here too as it will throw an IntegrityError if the user isn't created.

tests/unit/accounts/test_urls.py
```diff
@@ -33,3 +34,17 @@ def test_disable_user(self):
 
         url = reverse_lazy('accounts:get_auth_token')
         self.assertEqual(str(url), '/api/accounts/user/get_auth_token')
+
+        url = reverse_lazy('accounts:generate_auth_token')
+        self.assertEqual(str(url), '/api/accounts/user/generate_auth_token')
+
+    def test_token_generate(self):
```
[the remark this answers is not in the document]
gescalona1: I do not understand this comment , as in `assertRaises(Exception, ...)`?

RishabhJain2018:
Thanks for the contribution. We will merge this later :)

RishabhJain2018:
Closing this PR for now. We will revisit it later.

PR #1909, opened 2018-11-15
[WIP] Add tests for leaderboard API

### LINKED ISSUE #1829: Add exception handling when key is not present in leaderboard API
The backend server throws 500 internal server error whenever any key in the results stored in the leaderboard data table is not present in the leaderboard schema for that particular challenge. Instead we want to show meaningful error on the UI with proper status code. Please see the relevant error log given below:

Sentry Report: https://sentry.io/cloudcv/evalai-staging/issues/605096490/

```
KeyError: u'spl'
(5 additional frame(s) were not displayed)
...
  File "rest_framework/views.py", line 477, in dispatch
    response = self.handle_exception(exc)
  File "rest_framework/views.py", line 437, in handle_exception
    self.raise_uncaught_exception(exc)
  File "rest_framework/views.py", line 474, in dispatch
    response = handler(request, *args, **kwargs)
  File "rest_framework/decorators.py", line 52, in handler
    return func(*args, **kwargs)
  File "jobs/views.py", line 259, in leaderboard
    item['result'] = [item['result'][index] for index in leaderboard_labels]

KeyError: u'spl'
```

Ram81:


apps/jobs/views.py
```diff
@@ -360,7 +360,14 @@ def leaderboard(request, challenge_phase_split_id):
...
     leaderboard_labels = challenge_phase_split.leaderboard.schema['labels']
     for item in distinct_sorted_leaderboard_data:
-        item['result'] = [item['result'][index] for index in leaderboard_labels]
+        result = []
+        for index in leaderboard_labels:
+            try:
+                result.append(item['result'][index])
+            except:
+                response_data = {'error': 'Sorry, label {} not found in leaderboard schema!'.format(index)}
+                return Response(response_data, status=status.HTTP_400_BAD_REQUEST)
```
RishabhJain2018: Can you please change the status to `404 NOT FOUND`.

apps/jobs/views.py
```diff
@@ -360,7 +360,14 @@ def leaderboard(request, challenge_phase_split_id):
 
     leaderboard_labels = challenge_phase_split.leaderboard.schema['labels']
     for item in distinct_sorted_leaderboard_data:
-        item['result'] = [item['result'][index] for index in leaderboard_labels]
+        result = []
+        for index in leaderboard_labels:
+            try:
+                result.append(item['result'][index])
+            except:
+                response_data = {'error': 'Sorry, label {} not found in leaderboard schema!'.format(index)}
```
RishabhJain2018: Please change response_data to `{'error': 'Sorry, label {} is not found in leaderboard schema'.format(index)}`

apps/jobs/views.py
```diff
@@ -360,7 +360,14 @@ def leaderboard(request, challenge_phase_split_id):
 
     leaderboard_labels = challenge_phase_split.leaderboard.schema['labels']
     for item in distinct_sorted_leaderboard_data:
-        item['result'] = [item['result'][index] for index in leaderboard_labels]
```
RishabhJain2018: Please add the test cases accordingly.

Ram81:
@RishabhJain2018 tests added.

tests/unit/jobs/test_views.py
```diff
@@ -841,3 +862,47 @@ def test_get_submission_by_pk_when_user_is_neither_challenge_host_nor_submission
...
+        }
+
+        self.leaderboard_data = LeaderboardData.objects.create(
+            challenge_phase_split=self.challenge_phase_split,
+            submission=self.submission,
+            leaderboard=self.leaderboard,
+            result=self.result_json
+        )
+
+    def test_leaderboard_schema_key_missing(self):
```
RishabhJain2018: Please change `test_leaderboard_schema_key_missing` to `test_leaderboard_schema_with_missing_key`

tests/unit/jobs/test_views.py
```diff
@@ -841,3 +862,47 @@ def test_get_submission_by_pk_when_user_is_neither_challenge_host_nor_submission
...
+            method_description="Test Description",
+            project_url="http://testserver/",
+            publication_url="http://testserver/",
+        )
+
+        self.submission.is_public = True
+        self.submission.status = Submission.FINISHED
+        self.submission.save()
+
+        # NOTE: result schema has one key missing for test_leaderboard_schema_key_missing
```
RishabhJain2018: Please update the below function name here also.

tests/unit/jobs/test_views.py
```diff
@@ -841,3 +862,47 @@ def test_get_submission_by_pk_when_user_is_neither_challenge_host_nor_submission
...
+            status="finished",
+            input_file=self.challenge_phase.test_annotation,
+            method_name="Test Method",
+            method_description="Test Description",
+            project_url="http://testserver/",
+            publication_url="http://testserver/",
+        )
+
+        self.submission.is_public = True
+        self.submission.status = Submission.FINISHED
```
RishabhJain2018: I think you set the submission status to finished in line 876. Why are you setting it again? Please let me know if I am missing some context here.

tests/unit/jobs/test_views.py
```diff
@@ -841,3 +862,47 @@ def test_get_submission_by_pk_when_user_is_neither_challenge_host_nor_submission
...
+        self.leaderboard_data = LeaderboardData.objects.create(
+            challenge_phase_split=self.challenge_phase_split,
+            submission=self.submission,
+            leaderboard=self.leaderboard,
+            result=self.result_json
+        )
+
+    def test_leaderboard_schema_key_missing(self):
+        self.url = reverse_lazy('jobs:leaderboard',
+                                kwargs={'challenge_phase_split_id': self.challenge_phase_split.id})
```
RishabhJain2018: I think we should add a test case to check if the challeng_ phase_split exists or not.

tests/unit/jobs/test_views.py
Ram81: When a submission is created for first time in `jobs/models.py` save method is written such that it sets status as `SUBMITTED` only.

Ram81: Which is why I have to explicitly set the status

tests/unit/jobs/test_views.py
Ram81: Adding it

tests/unit/jobs/test_views.py
```diff
@@ -841,3 +851,108 @@ def test_get_submission_by_pk_when_user_is_neither_challenge_host_nor_submission
...
+            dataset_split=self.dataset_split,
+            leaderboard=self.leaderboard,
+            visibility=ChallengePhaseSplit.PUBLIC
+        )
+
+        self.submission = Submission.objects.create(
+            participant_team=self.participant_team,
+            challenge_phase=self.challenge_phase,
+            created_by=self.user1,
+            status="finished",
```
RishabhJain2018: Please replace it with `submitted` then.

tests/unit/jobs/test_views.py
```diff
@@ -841,3 +851,108 @@ def test_get_submission_by_pk_when_user_is_neither_challenge_host_nor_submission
...
+
+        self.assertEqual(response.data['count'], expected['count'])
+        self.assertEqual(response.data['next'], expected['next'])
+        self.assertEqual(response.data['previous'], expected['previous'])
+        self.assertEqual(response.data['results'], expected['results'])
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+    def test_get_leaderboard_with_invalid_challenge_phase_split_id(self):
+        self.url = reverse_lazy('jobs:leaderboard',
+                                kwargs={'challenge_phase_split_id': 203})
```
RishabhJain2018: Please don't hardcode the values.

tests/unit/jobs/test_views.py
```diff
@@ -1,5 +1,6 @@
 import os
 import shutil
+import collections
```
RishabhJain2018: Can we please import it alphabetically.

tests/unit/jobs/test_views.py
```diff
@@ -12,7 +13,8 @@
 from rest_framework import status
 from rest_framework.test import APITestCase, APIClient
 
-from challenges.models import Challenge, ChallengePhase
+from challenges.models import Challenge, ChallengePhase, ChallengePhaseSplit
```
RishabhJain2018: Can we please import line 16 and 17 using a tuple.

Ram81:
@deshraj @RishabhJain2018 any idea why Travis CI can't find the credentials?

apps/jobs/views.py
```diff
@@ -360,7 +360,14 @@ def leaderboard(request, challenge_phase_split_id):
...
+        result = []
```
RishabhJain2018: @Ram81 Another better way of doing it to add this check while entering the data in the LeaderboardData Table i.e. in submission worker script.
 So, the flow would be
1. Check for the metrics in the particular leaderboard corresponding to the submission result.
2. If the keys match then the submission will be successful and data will be entered in the LeaderboardData table.
3. If the keys don't match then submission status will be set to failed, the error will be stored in the std_error file, LeaderboardData object won't be create and execution will be terminated.

Please remove this code and the corresponding test cases for it. We can merge the remaining test cases that you have added.

cc: @deshraj

Ram81: Sure

Ram81:
@RishabhJain2018 @deshraj test cases fixed please review

tests/unit/jobs/test_views.py
```diff
@@ -841,3 +855,111 @@ def test_get_submission_by_pk_when_user_is_neither_challenge_host_nor_submission
...
+            publication_url="http://testserver/"
+        )
+
+        self.submission.is_public = True
+        self.submission.status = Submission.FINISHED
+        self.submission.save()
+
+        self.result_json = {
+            'score': 50,
+            'test-score': 50
```
RishabhJain2018: Just a minor change, can we have different value for `test-score`. I want to have different values for both the matrics, like if one is 50.0 then other can be 75.0 or any positive number other than 50.0.
Also, please add float values.

tests/unit/jobs/test_views.py
```diff
@@ -841,3 +855,111 @@ def test_get_submission_by_pk_when_user_is_neither_challenge_host_nor_submission
...
+        }
+
+        self.expected_results = [self.result_json['score'], self.result_json['test-score']]
+        self.filtering_score = 50.0
```
RishabhJain2018: Can we please use this value from the above like self.result_json['score'].

PR #1914, opened 2018-11-19
Fix failing travis builds after using encrypted AWS keys

RishabhJain2018:


docker/dev/docker.env
```diff
@@ -1,8 +1,3 @@
-AWS_SECRET_ACCESS_KEY=x
-AWS_ACCESS_KEY_ID=x
-
-DEBUG=True
-DJANGO_SETTINGS_MODULE=settings.dev
```
deshraj: I am not sure why we need to remove this. Can you please give some context?

RishabhJain2018: If we run tests in docker env then it was taking settings from `environment` rather than taking it from `settings.test`. It was due to the reason as the pytest module considers the precedence order for settings module in [this](https://pytest-django.readthedocs.io/en/latest/configuring_django.html#pytest-ini-settings) way.
Moreover, we are already passing this as an argument in [this](https://github.com/Cloud-CV/EvalAI/blob/master/docker/dev/django/container-start.sh) script.

Just to be sure, we aren't using these settings anywhere else apart from what I described above?

PR #1933, opened 2018-12-02
Add fix for missing submission status field

RishabhJain2018:
Fixes https://sentry.io/cloudcv/evalai/issues/784258070/?query=is:unresolved

### DIFF
--- a/apps/jobs/views.py
+++ b/apps/jobs/views.py
@@ -637,7 +637,7 @@ def update_submission(request, challenge_pk):
         submission_status = request.data.get('submission_status').lower()
     except Exception as error:
         response_data = {'error': '`submission_status` key is missing from request body.'
-                        'Please try again after adding the key!'}
+                                  'Please try again after adding the key!'}
         return Response(response_data, status=status.HTTP_400_BAD_REQUEST)
     stdout_content = request.data.get('stdout', '')
     stderr_content = request.data.get('stderr', '')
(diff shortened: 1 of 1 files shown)


apps/jobs/views.py
```diff
@@ -633,7 +633,12 @@ def update_submission(request, challenge_pk):
 
     challenge_phase_pk = request.data.get('challenge_phase')
     submission_pk = request.data.get('submission')
-    submission_status = request.data.get('submission_status').lower()
```
deshraj: Instead of adding all these lines, can we set it to default to `''` and then it will get handled easily. No need to put try catch for this.

RishabhJain2018: All the other rest fields are taken care of using `''` but this field is important as challenge host must send us the submission status once the submission is submitted by a participant. I don't think we can use `''` in this field. That's why I have added a try-excpt block.
Let me know your thoughts about it.

deshraj: If you set it to `''` as the default field then line https://github.com/Cloud-CV/EvalAI/pull/1933/files#diff-42c752186f5f6b2ed96b92afa242a6f3R650 will already take care of the invalid submission status. 

Currently, it looks like there are extra try-catch blocks which can be handled with just one existing try-catch block. Do you agree with what I am saying? Let me know if you see any concern with the approach I am suggesting.

RishabhJain2018: yeah, I agree. I didn't see that we are handling it later in the code.

deshraj: No worries. Let me know when you are done with that change. We can then merge this PR. :)

RishabhJain2018: Yup, updated it.

PR #1939, opened 2018-12-04
Add error handling while making submission

RishabhJain2018:


deshraj:
@RishabhJain2018 can you please mention what was the original error and what will come now as the error?

RishabhJain2018:
The original error is when someone tries to submit to a challenge phase whose start and end date are in future then it shows this:
screenshot 2018-12-04 at 5 29 54 pm

Now the error shown in the API call will be shown.

deshraj:
Thanks for the clarification Rishabh. In future, can you please add a small description in the PR or an issue for the same if the purpose of the PR is not clear from the title itself? Thanks!

RishabhJain2018:
Sure! I'll do that.

PR #1962, opened 2018-12-16
Create separate requirements for production worker container

RishabhJain2018:


requirements/worker_production.txt
```diff
@@ -0,0 +1,8 @@
+matplotlib==2.2.3
```
deshraj: Cannot we do like `-r worker.txt`instead of replicating the same requirements here?

PR #1963, opened 2018-12-16
Fix: Test case runtime warnings

Ram81:


Ram81:
@RishabhJain2018 @deshraj to fix those warnings I have to make `auto_now_add=False`, but can we allow submitted_at=null ?

RishabhJain2018:
I am not in favour of making `auto_now_add=False` and the other change too.
@Ram81 Can you please try this solution https://stackoverflow.com/a/45080605?

Ram81:
@RishabhJain2018 I tried all three, none of them worked. Issue seems to be with `auto_now_add=True`

RishabhJain2018:
Closing this now since #1987 fixes the same thing.

PR #1961, opened 2018-12-16
Fix #1957 Add tests for get_broker_urls API

inishchith:
Added tests for `get_broker_urls` API.
I've added 2 tests
- `test_get_broker_urls_when_user_is_not_superuser`: when the request is made by a user who is not a super user hence [tesing](https://github.com/Cloud-CV/EvalAI/blob/master/apps/challenges/views.py#L1335).
- `test_get_broker_urls_when_challenge_is_approved`: when the request is made by a superuser hence [testing](https://github.com/Cloud-CV/EvalAI/blob/master/apps/challenges/views.py#L1338), when challenges are approved.

The reason why i've added superuser to `BaseAPITestClass` is the convenience of adding further tests related to superuser.

@RishabhJain2018 Please do let me know if any changes required .

Closes #1957

PR #1970, opened 2018-12-18
Fix #1965 Update response format for the API 

### LINKED ISSUE #1965: Update response format for the API
The response formats in the API's here https://github.com/Cloud-CV/EvalAI/blob/master/apps/challenges/views.py#L1330 should be of JSON type (Currently, it returns list) and consistent with other API's. For e.g. the `result` key should be used for showing the results and `error` key should be used for displaying the error.

Please feel free to ask if you have any doubts.

inishchith:
I've made changes to the response format as discussed in #1965 Thread.
Also made subsequent changes to #1961 

Closes #1965


inishchith:
@RishabhJain2018 Please review and let me know if any changes required .

apps/challenges/views.py
```diff
@@ -1345,6 +1345,9 @@ def get_broker_urls(request):
     else:
         challenges = Challenge.objects.filter(**q_params)
         response_data = challenges.values_list('queue', flat=True)
+        response_data = {
+                'result': list(response_data)
```
deshraj: I think we can just return the `{'results': response_data}` instead of converting it to the list. Also, please note that the key is `results` and not `result`.

apps/challenges/views.py
```diff
@@ -1366,6 +1369,7 @@ def get_broker_url_by_challenge_pk(request, challenge_pk):
         except Challenge.DoesNotExist:
             response_data = {'error': 'Challenge {} does not exist'.format(challenge_pk)}
             return Response(response_data, status=status.HTTP_400_BAD_REQUEST)
-
-        response_data = [challenge.queue]
+        response_data = {
+                'result': list(challenge.queue)
```
deshraj: Same comment as above.

inishchith: Done!

apps/challenges/views.py
inishchith: Done!

inishchith:
@deshraj I've made the required changes. Please do let me know if any other changes required.

RishabhJain2018:
Same comment for this one.
Please feel free to take up any other issue you want.

RishabhJain2018:
@inishchith Please resolve the conflicts here too. Let's get this merged too 🎉

inishchith:
@RishabhJain2018 I had updated the PR, can you have a look again?

Thanks!

PR #1971, opened 2018-12-19
Import/Export: Minor improvements in import-export for different apps

deshraj:
This PR introduces following changes:

- [x] Add export to User model
- [x] Change list_filter for different models for better admin experience
- [x] Make the use of `'` instead of `"` consistent in admin files
- [x] Add searching by username in Auth Token's Table.


RishabhJain2018:
@deshraj This looks good to me. Can we please also add searching by username in `Token` table. That will also be a minor change.

PR #1976, opened 2018-12-31
Updated the .gitignore file to avoid future merge conflicts arising from JetBrains IDE's

rshrc:
Previously the repository whenever cloned and opened in any JetBrains IDE, automatically created the .idea/ folder which lead to merge conflicts. Updating the .gitignore file now allows us to ignore that. Many developers including me use PyCharm or WebStorm for exploring and contributing to the source code of EvalAI.

IMPORTANT NOTES (please read, then delete):

This PR is not addressing any currently existing Issue. This is just an addition to ease the use of IDE's without resulting into merge conflicts.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,29 +58,7 @@ ssl/
 
 # Code editors
 .vscode
-.idea/**/workspace.xml
-.idea/**/tasks.xml
-.idea/**/usage.statistics.xml
-.idea/**/dictionaries
-.idea/**/shelf
-
-# Generated files
-.idea/**/contentModel.xml
-
-# Sensitive or high-churn files
-.idea/**/dataSources/
-.idea/**/dataSources.ids
-.idea/**/dataSources.local.xml
-.idea/**/sqlDataSources.xml
-.idea/**/dynamic.xml
-.idea/**/uiDesigner.xml
-.idea/**/dbnavigator.xml
-
-# CMake
-cmake-build-*/
-
-# Mongo Explorer plugin
-.idea/**/mongoSettings.xml
+.idea/
 
 # File-based project format
 *.iws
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -58,3 +58,38 @@ ssl/
 
 # Code editors
 .vscode
+.idea/**/workspace.xml
```
deshraj: It will be a good idea to just put `.idea/` in the gitignore rather than specifying the path of each of the files/folders inside that directory.

.gitignore
```diff
@@ -58,3 +58,38 @@ ssl/
...
+# Sensitive or high-churn files
+.idea/**/dataSources/
+.idea/**/dataSources.ids
+.idea/**/dataSources.local.xml
+.idea/**/sqlDataSources.xml
+.idea/**/dynamic.xml
+.idea/**/uiDesigner.xml
+.idea/**/dbnavigator.xml
+
+# CMake
```
vkartik97: Please remove `CMake` Directories as it is not used in PyCharm.

.gitignore
rshrc: Okay I will do it. This is me first time resolving a PR.. actually I looked at the JetBrains github, and they gave all these https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore which covers all the IDE's.. even WebStorm

.gitignore
rshrc: I took https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore this as a reference.. Here they show support for WebStorm too. Is CMake used in any of the other IDE's?

vkartik97: `CMake` is mostly used in `CLion`

.gitignore
rshrc: There has been no response. Is this PR reviewed yet?

.gitignore
rshrc: @vkartik97 hey did you check the changes I've made that you've requested. Thanks

PR #1985, opened 2019-01-06
[WIP] Fix #1929 Two Loaders working at same time 

### LINKED ISSUE #1929: Two Loaders working at same time
### Observed Behaviour
When a user wants to participate in a challenge, there are two loaders that are working. As seen in the attached screenshot.

### Expected Behaviour
There should be only one loader working.

### Steps to Reproduce
Go to a challenge and participate in a challenge. When you register a team for the challenge, you can observe two loaders working at the same time.

### Would you like to work on it?
Yes

@deshraj @RishabhJain2018 

image

Sanji515:
**Fix #1929 **

Now there is a separate loader for Create Team and pagination.


frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -147,17 +147,14 @@
                                                 vm.loaderTitle = '';
                                                 vm.loaderContainer = angular.element('.exist-team-card');
 
-                                                // show loader
-                                                vm.startLoader("Loading Teams");
-                                                // loader end
```
Sanji515: Now instead of using `startLoader()` function for showing loader I've just set `ng-show` variable to True which was already set to True in line no. 146

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -147,17 +147,14 @@
...
 
                                                 parameters.url = 'challenges/challenge/' + vm.challengeId + '/participant_team/' + vm.teamId;
                                                 parameters.method = 'POST';
                                                 parameters.callback = {
                                                     onSuccess: function() {
                                                         vm.isParticipated = true;
                                                         $state.go('web.challenge-main.challenge-page.submission');
-                                                        vm.stopLoader();
+                                                        vm.isExistLoader = false;
                                                     },
```
Sanji515: Same as above for hiding the loader we just set the `ng-show` variable to false.

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -166,7 +163,7 @@
                                                             error = response.data["error"];
                                                         }
                                                         $rootScope.notify("error", error);
-                                                        vm.stopLoader();
+                                                        vm.isExistLoader = false;
                                                     }
```
Sanji515: Same as above.

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -180,7 +177,6 @@
                                                 vm.loaderContainer = angular.element('.exist-team-card');
 
 
-                                                vm.startLoader("Loading Teams");
                                                 if (url !== null) {
```
Sanji515: Same as above.

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -208,7 +204,7 @@
                                                         } else {
                                                             vm.isPrev = '';
                                                         }
-                                                        vm.stopLoader();
+                                                        vm.isExistLoader = false;
```
Sanji515: Same as above.

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -831,6 +827,20 @@
...
+                vm.newContainer.addClass('low-screen');
+            };
+
+            // stop loader
+            vm.stopLoader = function() {
+                vm.isLoader = false;
+                vm.loaderTitle = '';
+                vm.newContainer.removeClass('low-screen');
+            };
+
```
Sanji515: For the right-side loader i.e. for _**Create Team**_ I've changed the `ng-show` variable from `challenge.isExistLoader`(which is already set for pagination _**"left-side loader"**_) to `challenge.isLoader` so for right side loader I've defined two functions(**startLoader & stopLoader**) for showing the loader.

frontend/src/views/web/challenge/participate.html
```diff
@@ -62,7 +62,7 @@
                 <div class="col s12 m6">
                     <div class=" ev-card-panel ev-z-depth-5">
                         <div class="ev-md-container ev-panel-title"><strong>Create a New Team</strong></div>
-                        <div class="loader-container card-loader-container" id="loader" ng-show="challenge.isExistLoader" class="fade">
+                        <div class="loader-container card-loader-container" id="loader" ng-show="challenge.isLoader" class="fade">
```
Sanji515: This is the `ng-show` variable where I've changed from `isExistLoader` to `isLoader` to differentiate the showing functionality of left-loader and right-loader.

Sanji515:
@deshraj

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -831,6 +827,20 @@
             vm.loaderTitle = '';
             vm.newContainer = angular.element('.new-team-card');
 
+            // show loader
+            vm.startLoader = function(msg) {
```
RishabhJain2018: Can you please tell us why are you creating this function?

Sanji515: because the functions `startLoader` and `stopLoader` which are defined in `frontend/src/js/services/loaderService.js` are toggling `isExistLoader` which is for the left loader not for the right loader, so for right loader in `frontend/src/views/web/challenge/participate.html` I've given the `ng-show="challenge.isLoader"`, therefore this function which I'm creating here just toggling the `isLoader` for the right loader.

Sanji515: @RishabhJain2018

RishabhJain2018: okay! but it can easily be done by removing right loader code in ` frontend/web/views/participate.html`. There is no need to add this function.
Please let me know if I am not clear.

Sanji515: @RishabhJain2018 As I know that we need two loader one when we create new team(displayed on the right side) and another when we go for another page for listing the team(displayed on the left side), so according to me if we want to display two loaders at different places then we need that code or if we want only single loader in both cases to display at center(or any particular place) then we don't need that code in `frontend/src/views/web/challenge/participate.html`

RishabhJain2018: We would ideally want a single loader in the center for both the cases you are talking. That's why I was asking to remove the code and have a single loader only.

Sanji515: @RishabhJain2018 I've done as you have told me, Please review.

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -1027,14 +1017,19 @@
...
             vm.showSubmissionNumbers = false;
+            vm.maxExceeded = false;
             parameters.url = "jobs/" + vm.challengeId + "/phases/" + phaseId + "/remaining_submissions";
             parameters.method = 'GET';
             parameters.callback = {
                 onSuccess: function(response) {
                     var status = response.status;
                     var details = response.data;
                     if (status === 200) {
-                        if (details.remaining_submissions_today_count > 0) {
```
Sanji515: This line actually came from master branch... I just pasted the code from master branch of `frontend/src/js/controllers/challengeCtrl.js` and then did my updates.

PR #1987, opened 2019-01-06
[WIP] Fix #1986  Test warning fix

### LINKED ISSUE #1986: Remove warnings generated during testing.
Datetime and Pagination related warnings are generated during testing.

Ayukha:
Fixes #1986

### DIFF
--- a/apps/analytics/views.py
+++ b/apps/analytics/views.py
@@ -89,13 +89,14 @@ def get_submission_count(request, challenge_pk, duration):
     q_params = {'challenge_phase__id__in': challenge_phase_ids}
     since_date = None
     if duration.lower() == 'daily':
-        since_date = timezone.now().date()
+        # Get the midnight time of the day
+        since_date = timezone.now().replace(hour=0, minute=0, second=0, microsecond=0)
 
     elif duration.lower() == 'weekly':
-        since_date = (timezone.now() - timedelta(days=7)).date()
+        since_date = (timezone.now() - timedelta(days=7)).replace(hour=0, minute=0, second=0, microsecond=0)
 
     elif duration.lower() == 'monthly':
-        since_date = (timezone.now() - timedelta(days=30)).date()
+        since_date = (timezone.now() - timedelta(days=30)).replace(hour=0, minute=0, second=0, microsecond=0)
     # for `all` we dont need any condition in `q_params`
     if since_date:
         q_params['submitted_at__gte'] = since_date
--- a/apps/challenges/views.py
+++ b/apps/challenges/views.py
@@ -84,7 +84,7 @@ def challenge_list(request, challenge_host_team_pk):
         return Response(response_data, status=status.HTTP_406_NOT_ACCEPTABLE)
 
     if request.method == 'GET':
-        challenge = Challenge.objects.filter(creator=challenge_host_team, is_disabled=False).order_by('-id')
+        challenge = Challenge.objects.filter(creator=challenge_host_team, is_disabled=False).order_by('id')
         paginator, result_page = paginated_queryset(challenge, request)
         serializer = ChallengeSerializer(
             result_page, many=True, context={'request': request})
@@ -294,7 +294,7 @@ def get_featured_challenges(request):
         featured=True,
         published=True,
         approved_by_admin=True,
-        is_disabled=False).order_by('-id')
+        is_disabled=False).order_by('id')
     paginator, result_page = paginated_queryset(challenge, request)
     serializer = ChallengeSerializer(result_page, many=True, context={'request': request})
     response_data = serializer.data
@@ -356,7 +356,7 @@ def get_challenges_based_on_teams(request):
         host_team_ids = get_challenge_host_teams_for_user(request.user)
         q_params['creator__id__in'] = host_team_ids
 
-    challenge = Challenge.objects.filter(**q_params).order_by('-id')
+    challenge = Challenge.objects.filter(**q_params).order_by('id')
     paginator, result_page = paginated_queryset(challenge, request)
     serializer = ChallengeSerializer(
         result_page, many=True, context={'request': request})
--- a/apps/challenges/views.py
+++ b/apps/challenges/views.py
@@ -84,7 +84,7 @@ def challenge_list(request, challenge_host_team_pk):
         return Response(response_data, status=status.HTTP_406_NOT_ACCEPTABLE)
 
     if request.method == 'GET':
-        challenge = Challenge.objects.filter(creator=challenge_host_team, is_disabled=False).order_by('id')
+        challenge = Challenge.objects.filter(creator=challenge_host_team, is_disabled=False).order_by('-id')
         paginator, result_page = paginated_queryset(challenge, request)
         serializer = ChallengeSerializer(
             result_page, many=True, context={'request': request})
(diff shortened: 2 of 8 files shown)


Ayukha:
@RishabhJain2018  Please review.

apps/analytics/views.py
```diff
@@ -89,13 +89,13 @@ def get_submission_count(request, challenge_pk, duration):
     q_params = {'challenge_phase__id__in': challenge_phase_ids}
     since_date = None
     if duration.lower() == 'daily':
-        since_date = timezone.now().date()
+        since_date = timezone.now().replace(hour=0, minute=0, second=0, microsecond=0)
```
RishabhJain2018: @Ayukha Can you add a reason why you need to make this change?

apps/challenges/views.py
```diff
@@ -84,7 +84,7 @@ def challenge_list(request, challenge_host_team_pk):
         return Response(response_data, status=status.HTTP_406_NOT_ACCEPTABLE)
 
     if request.method == 'GET':
-        challenge = Challenge.objects.filter(creator=challenge_host_team, is_disabled=False)
+        challenge = Challenge.objects.filter(creator=challenge_host_team, is_disabled=False).order_by('id')
```
RishabhJain2018: Why are we adding the `order_by` using `id`?

apps/challenges/views.py
```diff
@@ -294,7 +294,7 @@ def get_featured_challenges(request):
         featured=True,
         published=True,
         approved_by_admin=True,
-        is_disabled=False)
+        is_disabled=False).order_by('id')
```
RishabhJain2018: Same as above.

apps/hosts/views.py
```diff
@@ -28,7 +28,7 @@ def challenge_host_team_list(request):
 
     if request.method == 'GET':
         challenge_host_team_ids = ChallengeHost.objects.filter(user=request.user).values_list('team_name', flat=True)
-        challenge_host_teams = ChallengeHostTeam.objects.filter(id__in=challenge_host_team_ids)
+        challenge_host_teams = ChallengeHostTeam.objects.filter(id__in=challenge_host_team_ids).order_by('id')
```
RishabhJain2018: same as above.

apps/challenges/views.py
```diff
@@ -356,7 +356,7 @@ def get_challenges_based_on_teams(request):
         host_team_ids = get_challenge_host_teams_for_user(request.user)
         q_params['creator__id__in'] = host_team_ids
 
-    challenge = Challenge.objects.filter(**q_params)
+    challenge = Challenge.objects.filter(**q_params).order_by('id')
```
RishabhJain2018: same as above.

apps/hosts/views.py
```diff
@@ -105,7 +105,7 @@ def challenge_host_list(request, challenge_host_team_pk):
             challenge_host_status = challenge_host_status.split(',')
             filter_condition.update({'status__in': challenge_host_status})
 
-        challenge_host = ChallengeHost.objects.filter(**filter_condition)
+        challenge_host = ChallengeHost.objects.filter(**filter_condition).order_by('id')
```
RishabhJain2018: same as above.

settings/common.py
```diff
@@ -297,3 +298,9 @@
 REDOC_SETTINGS = {
     'SPEC_URL': ('docs.yaml', {'format': '.yaml'}),
 }
+
+# Datetime warning settings
```
RishabhJain2018: Also, instead of `# Datetime warning settings` please add briefly why is the below code added.

apps/participants/views.py
```diff
@@ -37,7 +37,7 @@ def participant_team_list(request):
     if request.method == 'GET':
         participant_teams_id = Participant.objects.filter(user_id=request.user).values_list('team_id', flat=True)
         participant_teams = ParticipantTeam.objects.filter(
-            id__in=participant_teams_id)
+            id__in=participant_teams_id).order_by('id')
```
RishabhJain2018: same as above.

apps/participants/views.py
```diff
@@ -74,7 +74,7 @@ def get_participant_team_challenge_list(request, participant_team_pk):
         return Response(response_data, status=status.HTTP_404_NOT_FOUND)
 
     if request.method == 'GET':
-        challenge = Challenge.objects.filter(participant_teams=participant_team)
+        challenge = Challenge.objects.filter(participant_teams=participant_team).order_by('id')
```
RishabhJain2018: same as above.

apps/analytics/views.py
Ayukha: Submitted_at is a datetime field so it was throwing warnings on receiving only a date. I used this code to pass the datetime of midnight rather than just today's date in since_date.

apps/challenges/views.py
Ayukha: It removes the Pagination related warnings.
I used [this](https://stackoverflow.com/questions/44033670/python-django-rest-framework-unorderedobjectlistwarning) link as a reference.

RishabhJain2018: How adding `order_by id` removes these warnings?

Ayukha: The issue is due to a new control for ordering added in Django 1.11 
https://docs.djangoproject.com/en/1.11/topics/pagination/#required-arguments

> 
```
def _check_object_list_is_ordered(self):
        """
        Warn if self.object_list is unordered (typically a QuerySet).
        """
        if hasattr(self.object_list, 'ordered') and not self.object_list.ordered:
            warnings.warn(
                'Pagination may yield inconsistent results with an unordered '
                'object_list: {!r}'.format(self.object_list),
                UnorderedObjectListWarning
            )`

```

> 

so i used the suggested [way](https://docs.djangoproject.com/en/1.11/topics/pagination/#required-arguments) to remove the error .

RishabhJain2018: Cool! This looks good. Also, please order by in `descending` order.

Ayukha: On changing the `order by` field to `descending` order test for `test_get_challenge_when_mode_is_host` is Failing.
Should I keep it in `ascending`?

RishabhJain2018: yeah, keep it ascending in that.

settings/common.py
```diff
@@ -297,3 +298,9 @@
 REDOC_SETTINGS = {
     'SPEC_URL': ('docs.yaml', {'format': '.yaml'}),
 }
+
+#Prevents Datetime warning by showing errors 
+warnings.filterwarnings(
```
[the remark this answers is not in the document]
Ayukha: @deshraj  `filterwarnings()` helped me to debug the code as it converted warnings into errors. I think it would be helpful for other developers if they face similar issues in future development.

apps/challenges/views.py
RishabhJain2018: I believe you misunderstood me here. We should keep the order `ascending` in the API for which the tests are failing.

settings/dev.py
```diff
@@ -26,7 +27,7 @@
 ACCOUNT_DEFAULT_HTTP_PROTOCOL = 'http'
 
 # DJANGO-SPAGHETTI-AND-MEATBALLS SETTINGS
-INSTALLED_APPS += [ # noqa: ignore=F405
+INSTALLED_APPS += [  # noqa: ignore=F405
```
RishabhJain2018: Please remove this change!

settings/dev.py
```diff
@@ -47,9 +48,15 @@
 
 MEDIA_URL = "/media/"
 
-MIDDLEWARE += [ # noqa: ignore=F405
+MIDDLEWARE += [  # noqa: ignore=F405
```
RishabhJain2018: Please remove this change!

Ayukha: @RishabhJain2018  Please review.

Issue #1984, opened 2019-01-06
Absence of decorators for authenticating and checking user status.


charanjit-singh:
I'd like to propose code standard optimisation:

**Observation**

- [Line 45](https://github.com/Cloud-CV/EvalAI/blob/master/apps/web/views.py), view notify_users_about_challenge will only rendered if and only if user is authenticated and it passes test of being a superuser. There are so many nested if else s, which can be eliminated by:

- [ ] Using Decorators: login_required and user_passes_test.

- [ ] Creating decorators for handling server error and page_not_found error.

Please note that there is no optimisation in performance. But the code management is effected.
For example we may want to handle 500 error for every view. So we cant use if else for handling. Usage of decorators can help there.

Thanks!

vkartik97:
@charanjit-singh Great Observation! 
What do you think about using `permissions.IsAuthenticated` instead of  `login_required`?

charanjit-singh:
Yes! using permissions will keep it in DRF's way.

charanjit-singh:
And Instead of rendering 404.html in views, we should Raise `Http404 `Exception.

`from django.http import Http404`

PR #1988, opened 2019-01-07
Update views.py

charanjit-singh:
Using decorators instead of nested if elses for authenticated requests.

Fix: #1984

### DIFF
--- a/apps/web/views.py
+++ b/apps/web/views.py
@@ -41,41 +41,37 @@ def internal_server_error(request):
     response.status_code = 500
     return response
 
-
+@permission_classes((permissions.IsAuthenticated,))
 def notify_users_about_challenge(request):
     """
     Email New Challenge Details to EvalAI Users
     """
-    if request.user.is_authenticated() and request.user.is_superuser:
-        if request.method == 'GET':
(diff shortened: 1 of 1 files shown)


apps/web/views.py
```diff
@@ -41,42 +41,38 @@ def internal_server_error(request):
     response.status_code = 500
     return response
 
-
+@permission_classes((permissions.IsAuthenticated,))
```
RishabhJain2018: @charanjit-singh Using this decorator doesn't solve our purpose since this API can only be accessed by `superuser`.

apps/web/views.py
```diff
@@ -41,42 +41,38 @@ def internal_server_error(request):
...
-                              'notification_email_conformation.html',
-                              {'message': 'All the emails are sent successfully!'})
-            except SMTPException:
-                logger.exception(traceback.format_exc())
-                return render(request,
-                              'notification_email_data.html',
-                              {'errors': 1})
-        else:
-            return render(request, 'error404.html')
+    if request.method == 'GET':
```
RishabhJain2018: The condition for verifying if a user is `superuser` or not is missing.

PR #1993, opened 2019-01-12
Settings: Change default timezone to be ET timezone

deshraj:
Currently, we have the default timezone to be set to UTC which makes it harder for EvalAI Admin to see when a particular entry was created in the database since there is a conversion from UTC to ET involved in this process. This PR solves this issue where it visualizes the time in ET instead of UTC on Django Admin.

### DIFF
--- a/settings/common.py
+++ b/settings/common.py
@@ -131,7 +131,7 @@ AUTH_PASSWORD_VALIDATORS = [
 
 LANGUAGE_CODE = 'en-us'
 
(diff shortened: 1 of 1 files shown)


vkartik97:
@deshraj 
If this is only for Django Admin, can we just use a middleware to change the timezone on `/admin` instead of changing timezone in settings?

RishabhJain2018:
Closing this for now as it is not the priority! We might visit it later.

PR #1999, opened 2019-01-14
Fix #1996: Django Updated to 1.11.18

### LINKED ISSUE #1996: Fix potential vulnerabilities due to outdated requirements
There are some outdated requirements which can open potential security threats. Hence we should update the vulnerable package versions as suggested by Github. Here is the list of upgrades to be done (as suggested by Github):

`django==1.11.18`

Please make sure to verify the validity and compatibility of suggestions with EvalAI codebase.

Ayukha:
Updates Django to 1.11.18 (#1996)


Ayukha:
@RishabhJain2018  I ran all the tests before opening the PR. And also checked the[ Django release notes](https://docs.djangoproject.com/en/1.11/releases/1.11.18/#django-1-11-18-release-notes) for any major update. But I did'nt tested it manually.
May I do that if required?

RishabhJain2018:
That's good. Please check it manually also, so as to make sure things are not breaking.

Ayukha:
I just did a manual check. Everything seems to work fine.

PR #2013, opened 2019-01-19
Fix minor bug in get_remaining_submissions_api

RishabhJain2018:


apps/jobs/views.py
```diff
@@ -487,6 +487,16 @@ def get_remaining_submissions(request, challenge_phase_pk, challenge_pk):
         # Calculate the remaining submissions for today.
         remaining_submissions_today_count = (max_submissions_per_day_count -
                                              submissions_done_today_count)
+
+        if remaining_submission_count < remaining_submissions_this_month_count:
```
deshraj: Instead of adding this if else block, can you please change it to something like this:

```
remaining_submission_count = <some integer value>
remaining_submissions_this_month_count = <some integer value>
remaining_submissions_today_count = <some integer value>

remaining_submissions_this_month_count = math.min(remaining_submission_count, remaining_submissions_this_month_count)
remaining_submissions_today_count = math.min(remaining_submissions_this_month_count, remaining_submissions_today_count)
```

deshraj: Please let me know if I am missing any case in this.

PR #2016, opened 2019-01-19
Fix #2014: Populate challenge's queue field with a random string

### LINKED ISSUE #2014: Update seed script to add broker url in challenge
## Current Behaviour:

Currently, when the challenge is created using the seed script, the `queue` field is not populated.

## Expected Behaviour:
When the seed script is run, the `queue` field must be populated with the random string of 75 characters in length.

yashdusing:
Fixes #2014 and adds random string to queue field of challenge model.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # Byte-compiled / optimized / DLL files
 __pycache__/
+.pytest_cache/
 *.py[cod]
 *$py.class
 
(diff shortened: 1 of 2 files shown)


yashdusing:
@RishabhJain2018 Could you please verify ?

RishabhJain2018:
@yashdusing Can you please fix the Travis Build?

PR #2024, opened 2019-01-21
[WIP] Fix #2014 : Fill challenge's queue field with a random string

### LINKED ISSUE #2014: Update seed script to add broker url in challenge

yashdusing:
Fix #2014 : Fill challenge's queue field with a random string


yashdusing:
@RishabhJain2018 Could you please verify now ?
Also I've left my proposal for review on your gitter could you please take a look at the proposal when you have the time to.

scripts/seed.py
```diff
@@ -134,6 +136,7 @@ def create_challenge(title, start_date, end_date, host_team):
     """
     evaluation_script = open(
         os.path.join(settings.BASE_DIR, 'examples', 'example1', 'sample_evaluation_script.zip'), 'rb')
+    random_string = ''.join(random.choice(string.ascii_letters) for _ in xrange(75))
```
RishabhJain2018: Please change it to `queue`

scripts/seed.py
```diff
@@ -149,6 +152,7 @@ def create_challenge(title, start_date, end_date, host_team):
         anonymous_leaderboard=False,
         start_date=start_date,
         end_date=end_date,
+        queue=random_string,
```
RishabhJain2018: `queue=queue`

scripts/seed.py
RishabhJain2018: Also, there is no `xrange` in Python3. Please change it to `range`.

yashdusing: About that, at the start of the file, there’s xrange = range for python so this should work ever for python 3.

RishabhJain2018: I see. But it would be nice if we follow only Python 3 syntax here. Can you please change it so that I can merge it.

yashdusing:
@RishabhJain2018 Made the 'range' change

PR #2030, opened 2019-01-22
[WIP] Fixes #2021: Adding HTTP interception to services

### LINKED ISSUE #2021: Add a http response interceptor to catch all responses with status code 401
## Current Behaviour

Currently, all the 401 responses are handled separately in different controllers which increases redundancy in the code.

## Expected Behaviour

A single function which will handle all the requests.

## Deliverables

- [ ] Create an interceptor service in `services.js` in order to catch all the response and check for 401 status code and return the response if there is 401 code in the response.
The below code can be utilized to create an interceptor service.

```
    angular
        .module('evalai')
        .service('utilities', utilities)
        .factory('authHttpResponseInterceptor',['$q','$location',function($q, $location){
            return {
                responseError: function(rejection) {
                    if (rejection.status === 401) {
                        localStorage.clear();
                        $location.path('/auth/login');
                    }
                    return $q.reject(rejection);
                }
            };
        }])
        .config(['$httpProvider',function($httpProvider) {
            //Http Intercpetor to check auth failures for xhr requests
            $httpProvider.interceptors.push('authHttpResponseInterceptor');
        }]);
```

- [ ] Remove all the 401 response codes from different controllers.

- [ ] Test the system end to end after removing the code.

yashdusing:
Fix #2021: Adding HTTP interception to services to handle 401 errors

### DIFF
--- a/frontend/src/js/controllers/analyticsCtrl.js
+++ b/frontend/src/js/controllers/analyticsCtrl.js
@@ -49,12 +49,6 @@
 
                     // navigate to permissions denied page
                     $state.go('web.permission-denied');
-                } else if (status == 401) {
-                    alert("Timeout, Please login again to continue!");
-                    utilities.resetStorage();
-                    $state.go("auth.login");
-                    $rootScope.isAuth = false;
-
                 }
             }
         };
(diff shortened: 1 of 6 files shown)


yashdusing:
@RishabhJain2018 I think that removing the 401 codes will not show alert messages anymore. Is that alright ?

yashdusing:
@deshraj @RishabhJain2018 
Tested the functionality by reducing the token expiration period as suggested. User experience is same as before.
Also, I added in an alert message because directly redirecting to login page seems like a poor user experience. Please tell me if you want it changed to before

PR #2035, opened 2019-01-23
Fix #2025: Fix directly editing host challenge bug

### LINKED ISSUE #2025: Editing hosted challenges directly doesn't work
- The expected behaviour : 
     Editing any hosted challenge properties should go smoothly
- The observed behaviour :
     Editing gives off an error
- Reason : localStorage which should contain the current challenge is not updated when we directly go through the 'Hosted Challenges' section. Only the 'All Challenges' section updates the localStorage
- Steps to reproduce the problem :
     1. Create a challenge (don't go to the 'All Challenges' section)
     2. Go to Hosted Challenges page and select your hosted challenge
     3. Edit the properties

screenshot_20190121_203454
screenshot-0 0 0 0-8888-2019 01 21-20-36-11

Would I like to take this up ? 
Yes.
Thanks!

yashdusing:
Fixes #2025 : Fix bug which occurs when host challenge is directly edited without going through the 'all challenges' pages (due to which the 'ChallengeCreator' wasn't updated)

### DIFF
--- a/frontend/src/js/controllers/hostedChallengeCtrl.js
+++ b/frontend/src/js/controllers/hostedChallengeCtrl.js
@@ -15,6 +15,8 @@
         utilities.showLoader();
 
         vm.challengeList = [];
+        vm.challengeCreator = {};
+        vm.currentList = {};
 
         var parameters = {};
         parameters.url = 'hosts/challenge_host_team/';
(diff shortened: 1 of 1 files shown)


yashdusing:
@RishabhJain2018 @deshraj Could you please verify the PR ?

frontend/src/js/controllers/hostedChallengeCtrl.js
```diff
@@ -32,6 +34,22 @@
...
+                            
+                            vm.currentList = data.results;
+                            for (var i in vm.currentList) {
+
+                                var descLength = vm.currentList[i].description.length;
+                                if (descLength >= 50) {
+                                    vm.currentList[i].isLarge = "...";
+                                } else {
+                                    vm.currentList[i].isLarge = "";
+                                }
```
RishabhJain2018: @yashdusing I don't think we need the code from line 41 - 46. Am I missing something here?

frontend/src/js/controllers/hostedChallengeCtrl.js
```diff
@@ -15,6 +15,8 @@
         utilities.showLoader();
 
         vm.challengeList = [];
+        vm.challengeCreator = {};
+        vm.currentList = {};
```
RishabhJain2018: `vm.currentList` --> `vm.results`

frontend/src/js/controllers/hostedChallengeCtrl.js
```diff
@@ -32,6 +34,22 @@
                             for (var j=0; j<data.results.length; j++){
                                 vm.challengeList.push(data.results[j]);
                             }
+                            
+                            vm.currentList = data.results;
+                            for (var i in vm.currentList) {
```
RishabhJain2018: Also, please use the above loop only. I don't think we need this loop also.

PR #2044, opened 2019-01-26
Fix #2041 : Added basic slack functionality

### LINKED ISSUE #2041: Add feature for slack notification
## Current Behaviour

Currently, when a challenge is created, submission is made or a contact message is received, EvalAI Admin doesn't get any notification.

## Expected Behaviour

A slack notification should be added for all the three events.

## Deliverables
- [ ] Create a common function in `base/utils` which will take input a slack webhook URL and a message to be sent to the slack.
- [ ] Add the notification for following events:
	- [ ] When a challenge zip is uploaded
	- [ ] When a contact message is received
	- [ ] When a submission is made
- [ ] Write test cases.

## Starter Code
```
def send_slack_notification(webhook=settings.SLACK_WEBHOOKS['default'], message=""):
    '''Send slack notification to any workspace
    
    Keyword Arguments:
        webhook {string} -- slack webhook URL (default: {settings.SLACK_WEBHOOKS['default']})
        message {str} -- JSON/Text message to be sent to slack (default: {""})
    '''

    try:
        response = requests.post(
            webhook,
            data=json.dumps({'text': str(message)}),
            headers={'Content-Type': 'application/json'}
        )
    except Exception as e:
        logger.info('Exception raised while sending slack notification. \n Exception message: {}'.format(e))
```

Please feel free to ask if you have any doubts!

yashdusing:
Fix #2041 : First version of slack functionality is added 

Currently, working on writing the test case


yashdusing:
@RishabhJain2018 Could you please verify if this is what was needed. 
Also, I haven't included any webhook in settings so by default should it be blank ?

apps/challenges/views.py
```diff
@@ -927,15 +927,18 @@ def create_challenge_using_zip_file(request, challenge_host_team_pk):
             for email in emails:
                 user = User.objects.get(email=email)
                 host = Participant(
-                            user=user,
-                            status=Participant.ACCEPTED,
-                            team=participant_host_team,
-                    )
+                    user=user,
```
vkartik97: @yashdusing 
I think the editor or Automatic PEP8 Formatter such as `autopep8` has made a lot of unnecessary changes which are not needed in this PR.

yashdusing: Yeah, I’ll clean that up in a while 😅

yashdusing:
@vkartik97 I didn’t understand what you meant by ‘without blocking the endpoint return’ . Doesn’t the ‘request’ module send HTTP request asynchronously ?

vkartik97:
@yashdusing Please refer http://docs.python-requests.org/en/latest/user/advanced/#blocking-or-non-blocking

vkartik97:
@yashdusing 
>With the default Transport Adapter in place, Requests does not provide any kind of non-blocking IO.

Ref: http://docs.python-requests.org/en/latest/user/advanced/#blocking-or-non-blocking

yashdusing:
@vkartik97 Doesn’t the same thing apply for all calls to backend api as well ?

vkartik97:
@yashdusing 
No as the endpoint has an `External Dependency` i.e Slack API and not Project Dependency. If at any point Slack API takes longer to respond, the endpoint will block the call and will proceed if there is any Success or Timeout, which would increase the overall Response Time.

vkartik97:
@yashdusing 
No as the endpoint has an `External Dependency` i.e Slack API and not Project Level i.e Internal Dependency. If at any point Slack API takes longer to respond, the endpoint will block the call and will proceed if there is any Response or Timeout, which would increase the overall Response Time.

yashdusing:
@vkartik97 Oh okay. Makes sense.
So, there are other libraries like requests-threads which is listed in the link you mentioned which could be used

vkartik97:
I would suggest to first experiment with `threading.Thread`.

yashdusing:
@deshraj @RishabhJain2018 @vkartik97 Could you please check ?

apps/base/utils.py
```diff
@@ -69,3 +75,30 @@ def decode_data(data):
...
+            webhook,
+            data=json.dumps({'text': str(message)}),
+            headers={'Content-Type': 'application/json'}
+        )
+    except Exception as e:
+        logger.info(
+            'Exception raised while sending slack notification. \n Exception message: {}'.format(e))
+
+
+def send_slack_notification(webhook=settings.SLACK_WEBHOOKS['default'], message=""):
```
vkartik97: @yashdusing 
`send_slack_notification` should only run if the Project is not running in `Dev`. Add a condition for it

settings/common.py
```diff
@@ -282,6 +282,9 @@
 # For inviting users to participant and host teams.
 ADMIN_EMAIL = "admin@cloudcv.org"
 CLOUDCV_TEAM_EMAIL = "EvalAI Team <team@cloudcv.org>"
+SLACK_WEBHOOKS = {
+    'default': ''
```
vkartik97: @yashdusing 
How are you testing the feature locally if the webook is blank?

yashdusing:
@vkartik97 I'm using a personal webhook which I haven't displayed in the PR

yashdusing:
@vkartik97 I'll add in the condition for DEV environment in a while. I'm also trying to look for a way to return values from a thread without using global variables (for the test case) but the code works even without it.

vkartik97:
@yashdusing 
You can use Pass by Reference to Inject a List(Mutable),  to `send_slack_notification`. Then, store the Response Code at 0th Index of List. You can then simply check if the Response Code Matches with Required Code or not for Testing, after a delay of some seconds(Preferable 2).
For Mocking POST, try `response` module.

For Example, 
```python
def threader(i):
    i[0] = 10
    print("Executed!")

i = [20]
def handle_sub_view(i):  
    threading.Thread(target=threader, args=[i]).start()
handle_sub_view(i)
time.sleep(2)
print(i[0])

# Output:
# Executed!
# 10
```

@deshraj  @RishabhJain2018  Can you please check if this method is efficient?

deshraj:
@yashdusing @vkartik97 Thanks for all the suggestions. Karthik did point out some valid questions in the thread (thanks @vkartik97 for thinking about the corner cases). But I would recommend to first use the simple code for sending the notifications to slack without using threading. The ultimate plan is to use Celery to queue these kind of jobs and let them run in background. 

@yashdusing so the set of todos for you in this PR should be:

- [ ] Integrate slack request code shared by @RishabhJain2018
- [ ] Send slack message for different events (like: submission done, challenge zip uploaded, new user signup) 
- [ ] Mock requests for dev. so that you can write tests for this

Once we have these functionalities, the next pull request will contain the following changes:

- [ ] Integrate celery (I believe Rishabh already have a PR for that. Just needs some refining)
- [ ] Put these slack notifications in celery tasks

Please let me know if you have questions. :)

yashdusing:
@deshraj @RishabhJain2018 Could you please review the PR ?

yashdusing:
The test case I've written is failing because there is no webhook provided in settings. It would work correctly if a webhook is provided

yashdusing:
@vkartik97 Correct me if I am wrong but isn't the library you provided used for mimicking the responses ?
If so, then I don't get the point of using a test case. The test case I've written is helpful only to check whether the slack API is up and the internet connection is working (since it is a simple requests usage).

vkartik97:
@yashdusing 
- The reason of using Test is just to ensure that if in future, anyone tries to modify `send_slack_notification`, the changes should not break the feature. 
- Mocking because Unit Test is efficient when run on `Isolated Environment`, where a Fixed State(sometimes reusable) will be used for Assert Tests.

- >The test case I've written is helpful only to check whether the slack API is up and the internet connection is working

    If you are sending `Testing Slack notification functionality`  via Post Request on Webhook URL defined in Settings.py (Ref: [tests/unit/base/test_utils.py#L121](https://github.com/Cloud-CV/EvalAI/blob/5bef35b2dab06aa739c9e5b6aba2c5508f2f3122/tests/unit/base/test_utils.py#L121) ), this will be actually be used for Posting Message using `Slack API` whenever a Test is run even on `Travis`.

yashdusing:
@deshraj @RishabhJain2018 @vkartik97 Please review the PR. I added in the mock responses for the unit tests

apps/base/utils.py
```diff
@@ -69,3 +74,23 @@ def decode_data(data):
...
+    '''Send slack notification to any workspace
+
+    Keyword Arguments:
+        webhook {string} -- slack webhook URL (default: {settings.SLACK_WEBHOOKS['default']})
+        message {str} -- JSON/Text message to be sent to slack (default: {""})
+    '''
+    if settings.DEBUG is False:
```
deshraj: You don't need this check here. Please remove it. In dev also, we might want to test the slack notifications sometimes and then this condition will prohibit us from testing that.

apps/base/utils.py
```diff
@@ -69,3 +74,23 @@ def decode_data(data):
...
+        try:
+            response = requests.post(
+                webhook,
+                data=json.dumps({'text': str(message)}),
+                headers={'Content-Type': 'application/json'}
+            )
+            return response
+        except Exception as e:
```
deshraj: Instead of doing exception handling, it would make more sense to check based on the response status.

apps/base/utils.py
```diff
@@ -69,3 +74,23 @@ def decode_data(data):
...
+            logger.info(
+                'Exception raised while sending slack notification. \n Exception message: {}'.format(e))
```
deshraj: Please send the slack message in logger.info if the request didn't go through properly.

apps/challenges/views.py
```diff
@@ -936,6 +936,9 @@ def create_challenge_using_zip_file(request, challenge_host_team_pk):
 
             zip_config.challenge = challenge
             zip_config.save()
+
+            send_slack_notification(
+                message="A *new challenge* has been uploaded to EvalAI")
```
deshraj: This message is not informative. Please stringify a dict that contains 'link of the zip uploaded', 'primary key of uploaded config' and 'Challenge name' (if a challenge is created).

apps/jobs/views.py
```diff
@@ -174,6 +174,7 @@ def challenge_submission(request, challenge_id, challenge_phase_id):
             submission = serializer.instance
             # publish message in the queue
             publish_submission_message(challenge_id, challenge_phase_id, submission.id)
+            send_slack_notification(message="A *new submission* has been uploaded")
```
deshraj: Similarly, add the submission details here. The dict should look like this:
```
{'challenge_phase': 'Test Phase 1', 'challenge': 'Random Challenge', 'challenge_id': 2, 'event': 'New submission', 'participant': 'deshraj'}
```

apps/web/views.py
```diff
@@ -98,6 +99,7 @@ def contact_us(request):
         if serializer.is_valid():
             serializer.save()
             response_data = {'message': 'We have received your request and will contact you shortly.'}
+            send_slack_notification(message="A *new contact message* is received")
```
deshraj: Add the contact message details inside the slack message. Otherwise, it's not useful at all.

requirements/common.txt
```diff
@@ -18,4 +18,5 @@ Pillow==3.4.2
 psycopg2==2.7.3.2
 PyYaml==4.2b1
 proc==0.10.1
+responses==0.5.1
```
deshraj: Nice. This is what we were looking for. But do you think we need this in prod config? If not, then please put it in dev.txt

yashdusing: @deshraj I'm not really sure if we need this in prod. If 'travis' uses the prod build for testing then we will need it for sure. If in the prod build, we were to remove the responses codes and put in the actual webhook then it might work for dev.txt only

yashdusing:
@deshraj @RishabhJain2018 Could you review the PR again please ?

requirements/common.txt
deshraj: We actually don't need this in prod. Feel free to move it to dev.py

apps/base/utils.py
deshraj: I think it should just work since Travis is also just a machine running all the tests. It should ideally work on all the machines. Can you please spend some more time on this to figure it out? Thanks for your patience.

settings/common.py
```diff
@@ -282,6 +282,9 @@
 # For inviting users to participant and host teams.
 ADMIN_EMAIL = "admin@cloudcv.org"
 CLOUDCV_TEAM_EMAIL = "EvalAI Team <team@cloudcv.org>"
+SLACK_WEBHOOKS = {
+    'default': 'http://testslackwebhook.com/webhook'
```
deshraj: Let's not hardcode it. Can you please fetch this from the environment ?

yashdusing: @deshraj I didn't understand what you meant by fetching from the environment. Could you please elaborate

apps/base/utils.py
yashdusing: Fixed this. I missed adding in mocks for test cases where send_slack_notifications was indirectly used

settings/common.py
yashdusing: did you mean this (https://stackoverflow.com/questions/4906977/how-do-i-access-environment-variables-from-python) ?

yashdusing:
@deshraj @RishabhJain2018 could you please check the PR and confirm the doubts that I have ?

yashdusing:
@deshraj @RishabhJain2018 I changed the settings so that instead of hard-coding the url, I fetched it from the OS environment

settings/common.py
```diff
@@ -284,7 +284,7 @@
 CLOUDCV_TEAM_EMAIL = "EvalAI Team <team@cloudcv.org>"
 
 slack_webhook = os.environ.get('SLACK_WEBHOOK')
-if slack_webhook == None:
+if slack_webhook is None:
```
vkartik97: @yashdusing 
Instead of this check, you can use
```python
slack_webhook = os.getenv('SLACK_WEBHOOK', 'http://testslackwebhook.com/webhook')
```
Please Ref: https://docs.python.org/3/library/os.html#os.getenv

yashdusing: Thanks. Simplifies the logic :) :+1:

RishabhJain2018:
@yashdusing Can you please resolve the conflicts? Let's get this merged 🎉

yashdusing:
@RishabhJain2018 I've resolved the merge conflicts

yashdusing:
@RishabhJain2018 please review

RishabhJain2018:
@yashdusing if possible could you please resolve the conflicts here. We'd like to get it merged.

PR #2071, opened 2019-02-02
Fix #1740 : Bug fix in API for fetching challenge phase details

### LINKED ISSUE #1740: Modify challenge phase query for displaying challenge phase detail
**Current Scenario:**
The function [here](https://github.com/Cloud-CV/EvalAI/blob/master/apps/challenges/views.py#L398) will **not** throw the error even if there are no challenge phase corresponding to the challenge.

**Steps to Reproduce:** You will only be able to check it only at the API level.
Let us suppose there exists a challenge with id 34 and phase ids 120, 121. Now when you will hit the [url ](https://github.com/Cloud-CV/EvalAI/blob/master/apps/challenges/urls.py#L17)with challenge id 34 and phase id say 1 then it won't give error if phase id 1 exists(for any other challenge). 

Ideally, it this [url](https://github.com/Cloud-CV/EvalAI/blob/master/apps/challenges/urls.py#L17) should only accept the phase ids 120 and 121 and should throw error for every other phase id.

**Deliverables:**
Modify the challenge phase query [here](https://github.com/Cloud-CV/EvalAI/blob/master/apps/challenges/views.py#L406) to include the `challenge` parameter in it.

What should be added:
`challenge_phase = ChallengePhase.objects.get(challenge=challenge, pk=pk)`
Also, Update the response message to
`response_data = {'error': 'Challenge Phase {} does not exist'.format(pk)}`

Ayukha:
Modifies challenge phase query for displaying challenge phase detail (closes #1740 )


Ayukha:
@RishabhJain2018 Please review.

apps/challenges/views.py
```diff
@@ -523,9 +523,9 @@ def challenge_phase_detail(request, challenge_pk, pk):
         return Response(response_data, status=status.HTTP_406_NOT_ACCEPTABLE)
 
     try:
-        challenge_phase = ChallengePhase.objects.get(pk=pk)
+        challenge_phase = ChallengePhase.objects.get(challenge=challenge, pk=pk)
     except ChallengePhase.DoesNotExist:
-        response_data = {"error": "ChallengePhase does not exist"}
+        response_data = {"error": "Challenge Phase {} does not exist".format(pk)}
```
RishabhJain2018: Please modify the error message to `{"error": "Challenge Phase {} does not exist for challenge".format(pk, challenge.pk)}`

tests/unit/challenges/test_views.py
```diff
@@ -2107,14 +2107,12 @@ def test_update_challenge_phase_when_user_is_its_creator(self):
...
-                "challenge_pk": self.challenge.pk,
-                "pk": self.challenge_phase.pk + 2,
-            },
-        )
-        expected = {"error": "ChallengePhase does not exist"}
+        self.url = reverse_lazy('challenges:get_challenge_phase_detail',
+                                kwargs={'challenge_pk': self.challenge.pk,
+                                        'pk': self.challenge_phase.pk + 2})
+        expected = {
+            'error': 'Challenge Phase {} does not exist'.format(self.challenge_phase.pk + 2)
```
RishabhJain2018: same as above!

PR #2085, opened 2019-02-07
Challenges: Create docker repository for participant team in docker based challenges

deshraj:
Included changes:

- [x] Add util function to create ECR repository
- [x] Add util function to create a federated user
- [x] Modify participation API to create ECR repository when user participates in docker based challenge
- [x] Update boto3 and botocore


apps/challenges/utils.py
```diff
@@ -20,3 +25,118 @@ def get_file_content(file_path, mode):
...
+            repositoryNames=[
+                name,
+            ]
+        )
+        repository = response['repositories'][0]
+    except ClientError as e:
+        if e.response['Error']['Code'] == 'RepositoryNotFoundException':
+            response = client.create_repository(repositoryName=name)
+            repository = response['repository']
+            created = True
```
RishabhJain2018: Please add an else block with logger.exception in case the create operation doesn't succeed.

PR #2091, opened 2019-02-08
Fix #1733: submission cannot be made when the challenge is not active or approved by admin

### LINKED ISSUE #1733: Error in receiving message from submission queue with error 'NoneType' object has no attribute 'get'
https://sentry.io/cloudcv/evalai/issues/498341967/

```
Error in receiving message from submission queue with error 'NoneType' object has no attribute 'get'
```

Note that this error occurs whenever you try to make submission to a challenge which is not active/not approved. This leads to making the submission status stuck to `SUBMITTED` state only. Instead we would like to change it to canceled and show user a readable error message that 'Sorry your submission cannot be processed since the challenge is not active or is not approved by EvalAI Admin.'

nikhilmudholkar:
Solved issue #1733

### DIFF
--- a/apps/challenges/urls.py
+++ b/apps/challenges/urls.py
@@ -21,6 +21,8 @@ urlpatterns = [
         name='get_all_challenges'),
     url(r'^featured/$', views.get_featured_challenges,
         name='get_featured_challenges'),
+    url(r'^challenge/phase/(?P<pk>[0-9]+)/$',
+        views.get_challenge_phase_by_pk, name='get_challenge_phase_by_pk'),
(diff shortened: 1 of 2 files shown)


nikhilmudholkar:
I had a look at build log. I might be wrong but do we need to update the test cases as well? If I am wrong then where is this implementation going wrong? Can anyone help me out here?

nikhilmudholkar:
@RishabhJain2018 can you help me out here? I am kind of stuck here

RishabhJain2018:
I'll review the PR's soon till then you can work on the newer issues. :)

PR #2106, opened 2019-02-10
Fix #2080: Deleting challenge redirects to dashboard

### LINKED ISSUE #2080: After deleting a challenge, redirect to Dashboard.
**Current Behaviour:**
After deleting a challenge, the page does not change and the user can still modify the contents of the page which obviously won't persist. 
**Expected:**
As the challenge has been deleted, it makes more sense to redirect the user to some relevant page like the dashboard or the Hosted Challenges page.
@RishabhJain2018 

**Deliverables:** 

- [x] Add necessary frontend code to redirect the user to the aforementioned page after confirming the deletion.

ArshdeepSingh98:


frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -1274,6 +1274,7 @@
                 };
 
                 utilities.sendRequest(parameters);
+                $location.path("web/dashboard");
```
deshraj: Can you please use `$state.go()` instead of using `$location.path`?

ArshdeepSingh98: done using state, please review https://github.com/Cloud-CV/EvalAI/pull/2106/commits/680f204cfce05749883b8c0ddb4efdb86d76e31a

ArshdeepSingh98: https://github.com/ArshdeepSingh98/EvalAI/pull/1/commits/73d87a065151460c0372414f12d8ec7684ff656d
Please review,
Thanks

PR #2118, opened 2019-02-12
Fix #2101: Added test cases for the filtering bug in Leaderboard API

### LINKED ISSUE #2101: Add unit tests for leaderboard API
## Deliverable

- [ ] Add unit tests for the corner case mentioned in PR #2100 in the leaderboard API.

yashdusing:
Fix #2101 : Added unit test cases for corner cases from #2100

### DIFF
--- a/tests/unit/jobs/test_views.py
+++ b/tests/unit/jobs/test_views.py
@@ -1212,6 +1212,18 @@ class ChallengeLeaderboardTest(BaseAPITestClass):
             publication_url="http://testserver/"
         )
 
+        self.submission2 = Submission.objects.create(
+            participant_team=self.participant_team,
+            challenge_phase=self.challenge_phase,
+            created_by=self.user1,
+            status="submitted",
+            input_file=self.challenge_phase.test_annotation,
+            method_name="Test Method",
+            method_description="Test Description",
+            project_url="http://testserver/",
+            publication_url="http://testserver/"
+        )
+
         self.private_submission = Submission.objects.create(
             participant_team=self.host_participant_team,
             challenge_phase=self.private_challenge_phase,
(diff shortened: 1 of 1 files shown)


yashdusing:
@RishabhJain2018 @deshraj Could you please review ?

yashdusing:
@Ram81 read the comment I made for the test case that failed.

yashdusing:
The expected submission should have been the lower ‘X’ value submission as was solved in the PR, but the challenge host still sees the higher ‘X’ value submission (this is not the case for the frontend)

Ram81:
Doesn't this depend on sorting order, correct me if I am wrong ?

yashdusing:
@Ram81 I don’t think so. I’ll look over the functionality but if you were to see the referenced PR (#2100), it is evident that the challenge host should see only the public submissions. And, in the test cases, in total, there are only 2 submissions (1 private and 1 public)

yashdusing:
@Ram81 if you meant that the highest value submission is shown, then yes that is true. But, it has to be a public submission. The test case which fails gets the private submission (which is highest but shouldn’t be visible to challenge host)

Ram81:
This is mentioned in the issue
```
The entries on the leaderboard viewed by the challenge host and the normal user should be the same (if they are public). i.e for this case challenge host should be able to see entry A whereas normal users should neither see A or B.
```
And this was the consideration
```
1. Now, make two submissions to a challenge phase `Z` named as A & B. Keep A as private and B as public. Also, the value of x is higher (in magnitude) for submission A than for B.
```
That's why I wanted to confirm

yashdusing:
I’ll look over the functionality once just to be sure but as it stands normal users and challenge hosts can see highest public submissions

Ram81:
@yashdusing any updates on this ?

yashdusing:
@Ram81 I’ll work on this after a week. I have my exams currently. IIRC the behaviour in test case seems different to that in UI for some reason.

Issue #2109, opened 2019-02-11
Add unit tests for get_challenge_phase_by_pk API


RishabhJain2018:
There are missing unit tests for the API added in #2108

Sanji515:
@RishabhJain2018 I want to work on this issue but I haven't written unit tests before.

RishabhJain2018:
Sure, you can work on it! For reference, you can take a look at the `tests/` folder in the repository. Also, feel free to ask if you have some doubts.

Sanji515:
@RishabhJain2018 I've added the unit tests but how can I verify if they are correct.

kurianbenoy:
@Sanji515 haven't you put a PR for this issue?

Sanji515:
@kurianbenoy Hii!!, yup! I've already dropped a PR 3 days ago.

PR #2119, opened 2019-02-12
Fix: #2109 Add unit tests for get_challenge_phase_by_pk API

Sanji515:
I've written the tests for three cases:
- When `challenge phase` exists for the same pk.
- When `challenge phase` does not exist.
- And the last one when the user is not authenticated.


Sanji515:
@RishabhJain2018 @deshraj Please review.

tests/unit/challenges/test_views.py
```diff
@@ -3005,3 +3005,52 @@ def test_star_challenge(self):
...
+        expected = {
+            "id": self.challenge_phase.id,
+            "name": self.challenge_phase.name,
+            "description": self.challenge_phase.description,
+            "leaderboard_public": self.challenge_phase.leaderboard_public,
+            "start_date": "{0}{1}".format(self.challenge_phase.start_date.isoformat(), 'Z').replace("+00:00", ""),
+            "end_date": "{0}{1}".format(self.challenge_phase.end_date.isoformat(), 'Z').replace("+00:00", ""),
+            "challenge": self.challenge_phase.challenge.pk,
+            "is_public": self.challenge_phase.is_public,
+            "is_active": True,
```
Ram81: @Sanji515 can't we fetch this from `self.challenge_phase` instead of hardcoding it?

tests/unit/challenges/test_views.py
```diff
@@ -3005,3 +3005,52 @@ def test_star_challenge(self):
...
+            "codename": "Phase Code Name",
```
Ram81: Same here

tests/unit/challenges/test_views.py
Sanji515: @Ram81 Actually the `response.data` is a dictionary and `self.challenge_phase` is a `queryset` so for comparing we need both of them to be a dictionary.

Sanji515: @Ram81 We can't fetch `is_active` from `self.challenge_phase` because it`(is_active)` came from `serializers.py` not from `models.py`.

tests/unit/challenges/test_views.py
Sanji515: @Ram81 done.

Sanji515:
@Ram81 @RishabhJain2018 Please review.

tests/unit/challenges/test_views.py
```diff
@@ -3643,3 +3643,53 @@ def test_star_challenge(self):
...
+            "is_public": self.challenge_phase.is_public,
+            "is_active": True,
+            "codename": self.challenge_phase.codename,
+            "slug": self.challenge_phase.slug,
+        }
+        response = self.client.get(self.url, {})
+        self.assertEqual(response.data, expected)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+    def test_get_challenge_phase_by_pk_when_user_is_not_authenticated(self):
```
RishabhJain2018: Hey, @Sanji515 I just noticed. The API [here](https://github.com/Cloud-CV/EvalAI/blob/master/apps/challenges/views.py#L1868) doesn't need authentication. Hence this test case isn't required. Moreover, please add the test case for this endpoint URL also.

Sanji515: Hi @RishabhJain2018, I don't get it actually, of which endpoint URL you want me to write tests?

RishabhJain2018: This URL: https://github.com/Cloud-CV/EvalAI/blob/master/apps/challenges/urls.py#L144

Sanji515:

tests/unit/challenges/test_views.py
```diff
@@ -3679,3 +3679,127 @@ def test_star_challenge(self):
...
+            password="secret_psassword",
+        )
+
+        EmailAddress.objects.create(
+            user=self.user1,
+            email="user1@test.com",
+            primary=True,
+            verified=True,
+        )
+
```
RishabhJain2018: One test case is missing when the `challenge_pk` is invalid. Can you please add that.

tests/unit/challenges/test_views.py
```diff
@@ -3679,3 +3679,127 @@ def test_star_challenge(self):
...
+        self.assertEqual(response.data, expected)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+    def test_get_challenge_phases_by_challenge_pk_when_user_is_not_challenge_host(self):
+        """
+        This is the case in which a user is not a challenge host
+        """
+        self.client.force_authenticate(user=self.user1)
+        expected = {"error": "Sorry, you are not authorized to access these challenge phases."}
+
```
RishabhJain2018: Please remove this line.

Sanji515:
@RishabhJain2018 I've applied all the suggested changes, please review.

PR #2121, opened 2019-02-14
Add field in challenge model to restrict pushing large docker images

RishabhJain2018:


apps/challenges/models.py
```diff
@@ -60,6 +60,8 @@ def __init__(self, *args, **kwargs):
         default=False, verbose_name="Is Docker Based", db_index=True)
     slug = models.CharField(
         max_length=200, db_index=True, default='')
+    max_docker_image_size = models.PositiveIntegerField(
+        default=50, null=True, blank=True)
```
deshraj: Two things:

1. We should store the value in bytes rather than in GB or MB. It will be much easier to compare since most of the docker commands give result in bytes and we will be able to avoid extra conversion. 
2. The default value should be quite high. Something 32 GB sounds fine to me.

PR #2128, opened 2019-02-15
Fix #2127: Update the Django version

### LINKED ISSUE #2127: Fix potential vulnerabilities due to outdated requirements
There are some outdated requirements which can open potential security threats. Hence we should update the vulnerable package versions as suggested by Github. Here is the list of upgrades to be done (as suggested by Github):

`django==1.11.20`

Please make sure to verify the validity and compatibility of suggestions with EvalAI codebase.

tawAsh1:
Fixes #2127 .

Django version
old 1.11.18
new 1.11.20

I have confirmed that there is no part in the code base to display decimal numbers exceeding 200 digits.

I hasten to send you the PR..

### DIFF
--- a/requirements/common.txt
+++ b/requirements/common.txt
@@ -2,7 +2,7 @@ asgi-redis==1.4.3
 boto3==1.9.88
 botocore==1.12.0
 commonmark==0.5.4
-django==1.11.18
+django==1.11.20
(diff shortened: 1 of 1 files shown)


RishabhJain2018:
Thanks for the fix @tawAsh1!

PR #2129, opened 2019-02-16
Add feature to limit the concurrent submission message processing from SQS

RishabhJain2018:


apps/jobs/views.py
```diff
@@ -745,3 +745,31 @@ def update_submission(request, challenge_pk):
...
+    challenge = get_challenge_model(challenge_pk)
+
+    if submission_status not in [Submission.SUBMITTED, Submission.RUNNING, Submission.FAILED,
+                                 Submission.CANCELLED, Submission.FINISHED, Submission.SUBMITTING]:
+        response_data = {
+            'error': 'Invalid submission status {}'.format(submission_status)
+        }
+        return Response(response_data, status=status.HTTP_400_BAD_REQUEST)
+
+    if not is_user_a_host_of_challenge(request.user, challenge.id):
```
deshraj: This check should come before you are checking the submission status.

scripts/workers/submission_worker.py
```diff
@@ -547,15 +547,34 @@ def main():
...
+                process_submission_callback(message.body)
+                # Let the queue know that the message is processed
+                message.delete()
+            if killer.kill_now:
+                break
+            time.sleep(0.1)
+    else:
+        maximum_concurrent_submissions = challenges.evaluate_concurrent_max_submissions
+        while True:
+            submissions_done_in_challenge_count = Submission.objects.filter(
```
deshraj: Shouldn't this check be after line 569 in this PR? Right now, you will basically be querying the database many times each second to get the number of running submissions which doesn't make sense.

apps/jobs/views.py
RishabhJain2018: Sure, but I wrote it in order like first all the checks regarding the URL is completed and then this one.

apps/jobs/views.py
```diff
@@ -745,3 +745,40 @@ def update_submission(request, challenge_pk):
     submission.save()
     response_data = {'success': 'Submission result has been successfully updated'}
     return Response(response_data, status=status.HTTP_200_OK)
+
+
+@api_view(['GET', ])
+@throttle_classes([UserRateThrottle])
+@permission_classes((permissions.IsAuthenticated, HasVerifiedEmail))
+@authentication_classes((ExpiringTokenAuthentication,))
+def get_challenge_submissions_by_status(request, challenge_pk):
```
deshraj: Please change the name to `get_submissions_for_challenge()` and add the status as a query param. 

Also, it would be nice if we can start using django filter backend for filtering. See https://django-filter.readthedocs.io/en/latest/guide/rest_framework.html for more info.

apps/jobs/views.py
```diff
@@ -745,3 +745,40 @@ def update_submission(request, challenge_pk):
...
+    submissions_done_in_challenge = Submission.objects.filter(
+        challenge_phase__challenge=challenge.id, status=submission_status)
+
+    submissions_done_in_challenge_count = submissions_done_in_challenge.count()
+
+    serializer = SubmissionSerializer(
+        submissions_done_in_challenge, many=True, context={'request': request})
+    response_data = {
+        'submissions': serializer.data,
+        'submissions_count': submissions_done_in_challenge_count
```
deshraj: Please remove the submissions_count since you can always get it from `len(serializer.data)`

apps/jobs/views.py
RishabhJain2018: Since we need this feature urgently, I've added it in the list and will add it later during GSoC.

apps/jobs/views.py
```diff
@@ -745,3 +745,35 @@ def update_submission(request, challenge_pk):
...
+
+    challenge = get_challenge_model(challenge_pk)
+
+    if not is_user_a_host_of_challenge(request.user, challenge.id):
+        response_data = {'error': 'Sorry, you are not authorized to make this request!'}
+        return Response(response_data, status=status.HTTP_400_BAD_REQUEST)
+
+    submission_status = request.query_params.get('status', None)
+
+    valid_submission_status = [Submission.SUBMITTED, Submission.RUNNING, Submission.FAILED,
```
deshraj: So this status check should be a part of the serializer and not in the view. Serializers are meant to do the data validation. Can you please move this in the serializer?

apps/jobs/views.py
```diff
@@ -745,3 +745,35 @@ def update_submission(request, challenge_pk):
...
+                               Submission.CANCELLED, Submission.FINISHED, Submission.SUBMITTING]
+
+    if submission_status not in valid_submission_status:
```
deshraj: This if block will also go away when we move the check to the serializer.

apps/jobs/views.py
```diff
@@ -745,3 +745,35 @@ def update_submission(request, challenge_pk):
...
+@authentication_classes((ExpiringTokenAuthentication,))
+def get_submissions_for_challenge(request, challenge_pk):
+
+    challenge = get_challenge_model(challenge_pk)
+
+    if not is_user_a_host_of_challenge(request.user, challenge.id):
+        response_data = {'error': 'Sorry, you are not authorized to make this request!'}
+        return Response(response_data, status=status.HTTP_400_BAD_REQUEST)
+
+    submission_status = request.query_params.get('status', None)
```
deshraj: Please use django filter here.

apps/jobs/views.py
deshraj: My bad. This is not a valid comment. Please use django-filter instead of manually validating the status.

RishabhJain2018:
@deshraj Added the issue #2219 for it.

PR #2133, opened 2019-02-17
Cancelled submissions should not be counted towards submission limits

RishabhJain2018:


apps/jobs/models.py
```diff
@@ -119,7 +119,12 @@ def save(self, *args, **kwargs):
                 participant_team=self.participant_team,
                 status=Submission.FAILED).count()
 
-            successful_count = self.submission_number - failed_count
+            cancelled_count = Submission.objects.filter(
```
deshraj: Instead of making a fresh query, you can do something like this:

```
            submissions = Submission.objects.filter(
                challenge_phase=self.challenge_phase,
                participant_team=self.participant_team)
           submission_status_to_exclude = [Submission.CANCELLED, Submission.FAILED]
           num_submissions_to_ignore = submissions.filter(status__in=submission_status_to_exclude)
```

Note that you have to remove the logic on line 117-121.

apps/jobs/views.py
```diff
@@ -420,10 +420,12 @@ def get_remaining_submissions(request, challenge_phase_pk, challenge_pk):
     max_submissions_per_month_count = challenge_phase.max_submissions_per_month
     max_submissions_per_day_count = challenge_phase.max_submissions_per_day
 
+    submission_status_to_exclude = [Submission.FAILED, Submission.CANCELLED]
```
deshraj: It would be great if this can come from some utils function so that we don't have to write this at multiple places. Does that make sense?

apps/jobs/models.py
```diff
@@ -114,12 +114,15 @@ def save(self, *args, **kwargs):
...
                 self.submission_number = 1
 
-            failed_count = Submission.objects.filter(
+            submissions = Submission.objects.filter(
                 challenge_phase=self.challenge_phase,
-                participant_team=self.participant_team,
-                status=Submission.FAILED).count()
+                participant_team=self.participant_team)
+
+            submission_status_to_exclude = [Submission.FAILED, Submission.CANCELLED]
```
deshraj: Can you please import it from constants file ?

PR #2140, opened 2019-02-20
Fix #2103: Redirect users to Profile section on less than medium screen

### LINKED ISSUE #2103: Redirect users to Profile section on less than medium screen
### Observed Behaviour
When a user visits any page on less than medium screen and tries to see `Profile` section, he is redirected to Home page.  You could see the Profile section from homepage only.

### Expected Behaviour
The user should be directed to the `Profile` page. 

@deshraj @RishabhJain2018

Sanji515:
Here's the screencast of it.

reload-profile-page-demo


Sanji515:
@RishabhJain2018 @gautamjajoo Please review.

PR #2145, opened 2019-02-20
Fix #1486: Deleting the challenges config entry when not used

### LINKED ISSUE #1486: Invalid Challenge creation using space
## Issues to be fixed
- [ ] When a user tries to create a challenge, if he uploads an invalid file ( zip without YAML) or any other format, the user gets an error but the files are stored in challenge configurations.

- [ ] This can result in large consumption of disk space and can be exploited by the users.

yashdusing:
Fixes #1486 : Deletes the unnecessary entry created of challenges config when error occurs

## Fixed issues checklist
* [x]  When a user tries to create a challenge, if he uploads an invalid file ( zip without YAML) or any other format, the user gets an error but the files are stored in challenge configurations.
* [x]  This can result in large consumption of disk space and can be exploited by the users.

### DIFF
--- a/apps/challenges/views.py
+++ b/apps/challenges/views.py
@@ -498,6 +498,10 @@ def create_challenge_using_zip_file(request, challenge_host_team_pk):
         uploaded_zip_file_path = serializer.data['zip_configuration']
     else:
         response_data = serializer.errors
+        try:
+            uploaded_zip_file.delete()
+        except Exception as e:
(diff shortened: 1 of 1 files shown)


yashdusing:
@RishabhJain2018 @deshraj could you please review the PR ?

yashdusing:
@RishabhJain2018 please review the PR

PR #2155, opened 2019-02-28
Remove unnecessary logger exceptions in challenge creation API and worker

RishabhJain2018:


apps/challenges/views.py
```diff
@@ -959,12 +944,9 @@ def create_challenge_using_zip_file(request, challenge_host_team_pk):
                 'error': 'Error in creating challenge. Please check the yaml configuration!'}
             return Response(response_data, status=status.HTTP_400_BAD_REQUEST)
         finally:
-            try:
-                shutil.rmtree(BASE_LOCATION)
-                logger.info('Zip folder is removed')
-            except:
-                logger.exception('Zip folder for challenge {} is not removed from location'.format(challenge.pk,
-                                                                                                   BASE_LOCATION))
+            shutil.rmtree(BASE_LOCATION)
```
deshraj: We should catch this exception. If this operation fails, then we should throw an error, no?

scripts/workers/submission_worker.py
```diff
@@ -216,7 +216,6 @@ def extract_challenge_data(challenge, phases):
         challenge_module = importlib.import_module(CHALLENGE_IMPORT_STRING.format(challenge_id=challenge.id))
         EVALUATION_SCRIPTS[challenge.id] = challenge_module
     except:
-        logger.exception('Exception raised while creating Python module for challenge_id: %s' % (challenge.id))
```
deshraj: We should catch this error, no? We will get an error on sentry if we are not able to start the worker.

RishabhJain2018: I removed it since we can always check it in the worker logs.

apps/challenges/views.py
RishabhJain2018: Sure :)

scripts/workers/submission_worker.py
deshraj: True but this can be really painful when you have a lot of challenges and you launch them using one script that loops all the challenges and launches workers for all the challenges. You will then have to manually do `docker ps` and see which all challenge id is missing from the running containers and then do `docker logs <container_id>`. We can avoid all that by just keep this line, no?

RishabhJain2018: I agree.

PR #2164, opened 2019-03-04
Fix challange page navbar

### LINKED ISSUE #2162: Fix challenge page display on mobile
## Current Behaviour
The tabs on the mobile are not properly aligned as shown here -
file

## Expected Behaviour
Fix the alignments.

geekayush:
Fix #2162 
@RishabhJain2018 
### When maximum width is 320px ( like iPhone 5/SE )
se
### When width is in between 320px and 360px ( like Galaxy S5/S9/S9+ )
galaxy
### When width is more than that ( like pixel 2/2XL )
pixel2
### When maximum width is 375px ( exceptional case like iPhone X/XS )
xxs


deshraj:
@geekayush can you please modify your code in such a way that:

1. We don't see the links in the header and just see the hamburger icon that is on the left. 
2. Move all the links to the sidebar that opens while clicking on the link (eg: https://materializecss.com/mobile.html)
3. For all medium and small sized screens, we dont want to show the links in the nav bar. You can easily do this using the classes present in the materializecss library (see https://materializecss.com/helpers.html). You won't have to write the custom media query that you have added in the css if you use the materializecss classes.

geekayush:
@deshraj 

Screen Shot 2019-03-11 at 23 19 31

Screen Shot 2019-03-11 at 23 19 39

I have hidden the links from the header and shifted it to the SideBar

gautamjajoo:
@RishabhJain2018 @galipremsagar This can be merged since it is approved.

Issue #1552, opened 2018-03-10
UI Improvements in Leaderboard


RishabhJain2018:
- [ ] Make the team name as **Bold** on leaderboard.

- [ ] Make `Rank` and `Participant Team` field as left aligned including the title.

- [ ] Make all the other fields as right aligned including the title.

- [ ] In the heading `submission time`, change the text to `Last Submission at` & display the delta between the current time and the last submission time. For Example: If the current time is 10 am, & last submission was made at 6 am, then the field should display `4h` (10am - 6am).

- [ ] Add a border line after every entry in the leaderboard.

Please feel free to ask if you have further doubts or queries.

ParthS007:
I want to work on this @RishabhJain2018

RishabhJain2018:
Sure go ahead @ParthS007 .

ParthS007:
@RishabhJain2018 Now the leaderboard looks like this.
![image](https://user-images.githubusercontent.com/24358501/37252355-c3e3be86-2545-11e8-9d7a-97fd20c18f0f.png)
Can you please help me with the 4th point?

RishabhJain2018:
@ParthS007 There are other fields `(yes/no, number, other, overall)` too in the right side of the screenshot. Please don't remove them. Just make them right aligned. Also, the border line after every entry is very dark. Make it similar to the one just after the heading. 
Moreover, in the fourth point we need to display the time difference between now and when the last submission was made. I think this can be easily handled client side and we don't need to modify the API. Please feel free to ask if you have further doubts.

RishabhJain2018:
@ParthS007 Since we need this feature early, I am taking this up. Please take up any other issue from the list. :-)

PR #2184, opened 2019-03-13
Maintenance Page and alignment Issue in leaderboard header.

galipremsagar:
This pull request contains two changes.
1. Fix to an alignment issue on the header of leaderboards.  This is due to the following commit:

https://github.com/Cloud-CV/EvalAI/commit/7edf7a0b0426cc5ffe84828f9be4542f24a410e3#diff-c09876db1a2167c9371dbcccf9a69aeb

But it was introduced on purpose, I tracked down to this issue #1552(https://github.com/Cloud-CV/EvalAI/issues/1552) where it was given as a task but the screenshot with changes shows correct alignment. I doubt why it is so, because the state of the UI alignment in the table header now doesn't look like the one in the screen-shot.
I have attached UI screenshot before and after the fix now. The reason could also be due to the shifting of date format from "Mar 10, 2018 6:20:52 PM" to "1 year ago".

Before:

Screenshot 2019-03-12 at 3 16 46 PM

After:
Screenshot 2019-03-12 at 9 04 19 PM

Before:

Screenshot 2019-03-12 at 3 17 07 PM

After:

Screenshot 2019-03-12 at 9 06 07 PM

2. Added a Maintenance-page. You may access the page at : http://0.0.0.0:8888/maintenance-page

But this is how it looks like currently: 

Screenshot 2019-03-12 at 9 16 42 PM

This is the present GIF I've used looks like this, however I'm open to any new/good looking GIFs, if you know any please suggest:

maintenance


galipremsagar:
Add two things:

1. Team based filtering in "View all submissions" tab. Here is a gif of how it looks:
ScreenRecording20190320at1

2. Updating the team members list "Participant Teams" tab whenever there is a new team member added. Previously the existTeam was not being updated when there is a new team member addition.

PR #2195, opened 2019-03-15
Submissions: Add API to get signed URL of a file in S3 bucket

deshraj:
#### Request URL

```
https://evalapi.cloudcv.org/api/jobs/submission_files/?bucket=<bucket_name>&key=team_{id}/submission_{id}/...../file.log
```

#### Response

Response codes:
1. 200 -- When signed url is returned
2. 400 -- When file path is invalid
3. 403 -- When user is not part of the participant team
4. 404 -- File not found in S3 bucket

Remaining todos (that should come as a separate PR):

- [x] Create a controller that makes API call to this endpoint
        - If it receives 200 response code from backend server, then redirects to the signed URL
        - Else, Display 404 or appropriate response page
- [x] Create a angular router.

### DIFF
--- a/frontend/src/js/route-config/route-config.js
+++ b/frontend/src/js/route-config/route-config.js
@@ -443,6 +443,13 @@
             title: "Accept challenge invitation"
         };
 
+        var get_submission_related_files = {
+            name: "get-submission-related-files",
+            url: "/:bucket/*key",
+            controller: "SubmissionFilesCtrl",
+            controllerAs: "submission_files",
+        };
+
         // call all states here
         $stateProvider.state(home);
         $stateProvider.state(privacy_policy);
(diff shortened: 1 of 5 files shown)


frontend/src/js/route-config/route-config.js
```diff
@@ -443,6 +443,13 @@
             title: "Accept challenge invitation"
         };
 
+        var get_submission_related_files = {
+            name: "get-submission-related-files",
+            url: "/:bucket/*key",
```
deshraj: Please change the URL to something like `evalai.cloudcv.org/web/submission/files?bucket=<blah>&key=<blah>`

RishabhJain2018: @deshraj I think we should not let users know what the bucket is or key is. We should parse it on our own in the controller.

How about this url -- `evalai.cloudcv.org/web/submission/files/<bucket_name>/<key>`

PR #2201, opened 2019-03-18
Fix #2189: Display error on UI while creating challenge

### LINKED ISSUE #2189: Display error on UI while creating challenge
## Current Behaviour
When there are syntax errors in the `yaml` challenge configs then the challenge creation using zip [API](https://github.com/Cloud-CV/EvalAI/blob/c156f3d54c95861c4ac4eaef1d2e3419bab83a7a/apps/challenges/views.py#L575) fails with 500 Internal Server error which should not be the case.

## Expected Behaviour
The API should give a proper error with 406 error code. The error should be displayed on the UI.

## Resources
For creating a challenge zip, please use [EvalAI-Starters](https://github.com/Cloud-CV/EvalAI-Starters)

Sanji515:
Fix #2189 

The exception that we are getting as `exc` is of exception type: `<class 'yaml.scanner.ScannerError'>`
- Previously, we were sending this exception directly as an error message due to which API fails with 500 Internal Server error.
- But now, we first take the required attribute from the exception and then send them as a proper syntax error message with 406 error code.

Here's the screencast of the same-

Recording #9 (1)


Sanji515:
@RishabhJain2018 @deshraj Please review.

apps/challenges/views.py
```diff
@@ -573,8 +573,22 @@ def create_challenge_using_zip_file(request, challenge_host_team_pk):
...
     except (yaml.YAMLError, ScannerError) as exc:
+        # To get the problem description
+        if hasattr(exc, 'problem'):
+            error_description = exc.problem
+            # To capitalize the first alphabet of the problem description as default is in lowercase
+            error_description = error_description[0:].capitalize()
+        # To get the error line and column number
+        if hasattr(exc, 'problem_mark'):
+            mark = exc.problem_mark
+            line_number = mark.line+1
```
RishabhJain2018: Please change it to `line_number = mark.line + 1`

apps/challenges/views.py
```diff
@@ -573,8 +573,22 @@ def create_challenge_using_zip_file(request, challenge_host_team_pk):
...
+            column_number = mark.column+1
```
RishabhJain2018: Please change it to `column_number = mark.column + 1`

apps/challenges/views.py
```diff
@@ -573,8 +573,22 @@ def create_challenge_using_zip_file(request, challenge_host_team_pk):
...
+        message = (
```
RishabhJain2018: I think it'd be better to display these errors on the page below the `Submit` button on `create challenge page` rather than in notification which disappears after some time. The challenge host won't be able to fix the error if it disappears.

apps/challenges/views.py
```diff
@@ -573,8 +573,22 @@ def create_challenge_using_zip_file(request, challenge_host_team_pk):
...
+            'Your yaml file contain invalid YAML syntax! '
```
RishabhJain2018: Please change the message to -
```
The YAML file in the challenge configuration contains the following error - 
<error>
```

apps/challenges/views.py
Sanji515: done.

apps/challenges/views.py
Sanji515: done.

apps/challenges/views.py
Sanji515: done.

apps/challenges/views.py
Sanji515: done.

RishabhJain2018:
@Sanji515 Please add the screencast for it.

Sanji515:
@RishabhJain2018 here is the screenshot for the same

Screenshot (200)

Sanji515:
@RishabhJain2018 Here is the screencast for the same.

Recording #10

Sanji515:
Hi, @RishabhJain2018  Sorry I forgot of one case i.e. for all other error messages except the 'yaml syntax error' that are to be shown over `toaster` are not showing as previously. I've added that case condition but these changes are not automatically appending with this PR.
How can I append these changes to this PR?

Sorry for the trouble.

Sanji515:
@RishabhJain2018 Please have a look at this.

RishabhJain2018:
Hey, @Sanji515 Thanks for figuring that out. Please open a new PR for that.

PR #2204, opened 2019-03-19
Fixes #2203: Update Python 2 code.

### LINKED ISSUE #2203: Outdated Python 2 code while creating challenge using zip file.
**Observed behavior:** When the serializer is not valid while creating a challenge, the error message never gets sent due to incompatible python 2 code which raises an error. Instead, the control is redirected to the except block, sending a generic message.

**Expected behavior:** The serializer error should be sent in the response when appropriate.

**Reason:** This happens because because of this line https://github.com/Cloud-CV/EvalAI/blob/1dde5467e6cfc4c7338f0107757025f5f80d1fed/apps/challenges/views.py#L940
In Python 3, dictionaries  return "view" objects from the `values()` which are iterable but don't support indexing. In python 2, they returned lists.
@RishabhJain2018 @deshraj

KhalidRmb:
Fix #2203 
**Changes:** Convert the `response_data.values() `object to a list before accessing the message given by serializer.error, which itself is a list.


KhalidRmb:
@deshraj @RishabhJain2018 Please Review!

apps/challenges/views.py
```diff
@@ -937,7 +937,7 @@ def create_challenge_using_zip_file(request, challenge_host_team_pk):
     except:
         try:
             if response_data:
-                response_data = {'error': response_data.values()[0]}
+                response_data = {'error': list(response_data.values())[0][0]}
```
rishabhbhardwaj: @KhalidRmb Looks like an issue with indexing. You are picking the first char of the message string. Please check.
```suggestion
                response_data = {'error': list(response_data.values())[0]}
```

RishabhJain2018:
@KhalidRmb Any updates on this??

apps/challenges/views.py
KhalidRmb: @rishabhbhardwaj Actually in this particular case the response_data is storing serializer.errors which itself is a dictionary. (Refer https://github.com/encode/django-rest-framework/blob/master/rest_framework/serializers.py). And that's why I'm accessing the 0 element twice.

KhalidRmb: Noted. I'll make the change.

RishabhJain2018:
@rishabhbhardwaj Can you please review this oen.

rishabhbhardwaj:
@KhalidRmb Thanks for the changes.
@RishabhJain2018 LGTM.

RishabhJain2018:
@KhalidRmb Can you please open this PR again with the same change?

PR #2205, opened 2019-03-19
Fix #1800: Fixing submission worker to load challenge data when submission made by challenge host

### LINKED ISSUE #1800: Load challenge data in the process when challenge is created
## Description

Currently, when someone creates a challenge on EvalAI, then gets the access to make submissions to that challenge in order to test it. But the catch here is that the submission worker is not ready to accept the submissions for that challenge since it needs to load the evaluation script and other required files. Hence whatever submission the challenge host does, it stays in the submitted state.

## How to fix it

- [ ] Add support for loading the required challenge related files whenever the challenge is created making sure that submission worker is reloaded or is loading the required files in memory

shiv6146:
Have fixed #1800 this issue by adding a new boolean field `is_host_submission` to the submissions model which will be False by default and will be set to True iff a submission is made by the host to one of their hosted challenges, in which case the submission worker is not ready to accept the submission as its not reloaded so making use of this field in `submission_worker.py` to determine and call `load_challenge()` if needed.


RishabhJain2018:
Closing this since it is a very old PR.

PR #2217, opened 2019-03-20
Fix #2206: Add error message when leaderboard schema is missing in challenge config

### LINKED ISSUE #2206: Display error message when Leaderboard schema is missing while creating challenge.
**Expected behavior**: A message like the following `"There is no leaderboard schema in the YAML configuration file. Please add it and then try again!"` with 406 status code should be returned when the schema is missing from the YAML file. This provides information to the user as to what needs to be done.

**Observed behavior:** A 500 internal server error.

**Deliverables:**
- [ ] Change the API to return appropriate response when the schema is not present.

@RishabhJain2018 @deshraj

shiv6146:
Have fixed #2206 by adding a new check after checking for leaderboard but before getting default_order_by

# Deliverables

- [x] Change the API to return appropriate response when the schema is not present.


rishabhbhardwaj:
@shiv6146 Please update with master.

rishabhbhardwaj:
@RishabhJain2018 LGTM

shiv6146:
@RishabhJain2018 Done :+1:

RishabhJain2018:
Thanks @rishabhbhardwaj for the review!

PR #2221, opened 2019-03-22
Fix [Remaining condition]: Display error on UI while creating challenge

Sanji515:
Recording #11


Sanji515:
@RishabhJain2018 Please review.
thanks!

RishabhJain2018:
Hey @Sanji515 Thanks for the effort but I think we should display the error at the bottom only which will be persistent and can help the challenge host rather than a notification which disappears. Hence, I'm closing it.

PR #2075, opened 2019-02-05
Removes database dependency from worker container

RishabhJain2018:
To Run the remote worker - 
1. Add the host auth token in the `docker.env` file
2. Change the queue name of the challenge to `evalai_submission_queue` in Django backend.
3. Change the result format in the evaluation script of the challenge. It should be acceptable by `update_challenge_data` API

PR #2239, opened 2019-03-27
Removes database dependency from submission worker

RishabhJain2018:
Duplicate of #2075.

To Run the remote worker - 
1. Add the host auth token in the `docker.env` file
2. Change the queue name of the challenge to `evalai_submission_queue` in Django backend.
3. Change the result format in the evaluation script of the challenge. It should be acceptable by `update_challenge_data` API


docker/dev/docker.env
```diff
@@ -14,3 +16,4 @@ POSTGRES_PASSWORD=postgres
 POSTGRES_HOST=db
 POSTGRES_PORT=5432
 SENDGRID_API_KEY=x
+QUEUE_NAME=evalai_submission_queue
```
deshraj: Please add a newline here.

Issue #2238, opened 2019-03-27
Update method to store queue name in Challenge model


deshraj:
Currently, we store the queue name for each challenge using something like this:

```
challenge_queue_name = "{}-{}-{}".format(challenge_title, challenge_pk, random_challenge_id)[:75]
```

Where the challenge title can be any string. But SQS does not allow special characters in the queue name and hence we should validate the queue name before storing it. 

## Deliverables

- [ ] Write a utils function that generates a valid SQS queue name given a string as an input
- [ ] Use the function written in the previous step to generate the queue name for a challenge
- [ ] Write tests to check if the function created in the first step is robust enough

tawAsh1:
Can i work on this?

deshraj:
@tawAsh1 Go ahead. :)

pushkalkatara:
Hi,

Is this issue resolved?

Sanji515:
Hey @pushkalkatara , I think it is already been added [here](https://github.com/Cloud-CV/EvalAI/blob/master/apps/base/utils.py#L196). Hence this issue can be closed @RishabhJain2018 @deshraj 

Feel free to take another issue @pushkalkatara

Sanji515:
@pushkalkatara but the tests still left for that function, so if you want then you can work on writing tests for it.

Thanks!

pushkalkatara:
@Sanji515 The tests were also written in the pull requests. Due to conflicts, it was not merged and the [PR](https://github.com/Cloud-CV/EvalAI/pull/2244/commits/f863e09d8c1a628209ee1b9e6c695133f2c2428f) was closed due to inactivity. Should I add the tests @deshraj?

RishabhJain2018:
Sure @pushkalkatara .

pushkalkatara:
@RishabhJain2018 ive added the tests. please review #2450

PR #2244, opened 2019-03-27
Fix: #2238 Add queue name generator

tawAsh1:
Fix: #2238 Add queue name generator.

https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_CreateQueue.html
Queue name should be less than or equal to 80 character and alphanumeric / - / _.

I implements queue name generator which generates valid string.
When string longer than 80 characters, truncates challenge_title to 43 characters.


RishabhJain2018:
@tawAsh1 I've already added the function for it [here](https://github.com/Cloud-CV/EvalAI/blob/master/apps/base/utils.py#L194). Can you resolve the conflicts?

PR #2252, opened 2019-04-01
Fix worker for dev and prod env for loading challenge

RishabhJain2018:
We load all the challenges in one worker container for the dev environment, while we load a single challenge in worker container for the prod. environment. Hence this PR loads the challenges based upon the environment.
I've to add the conditions for environments since we're also handling the case of running concurrent submissions in a challenge from the worker queue. For the dev environment where multiple challenges are loaded simultaneously the field, `max_concurrent_submission_evaluation` doesn't make sense whereas in prod. env it makes a lot of sense.


deshraj:
Can you please add a description for the purpose of the PR? I am not sure what this PR is doing. It would be great if you can follow this convention of adding PR description everytime.

RishabhJain2018:
sure :)

deshraj:
So basically this will not allow me to test the `max_concurrent_submission_evaluation` in the development environment, right? 

The better solution would be to have a settings/env variable that controls whether we want to enable the setting of `max_concurrent_submission_evaluation` or not in the dev environment.

You might want to think more about this and see what seems the best way to solve this problem.

docker/dev/docker.env
```diff
@@ -2,6 +2,7 @@ AWS_ACCESS_KEY_ID=x
 AWS_SECRET_ACCESS_KEY=x
 AWS_DEFAULT_REGION=us-east-1
 AWS_ACCOUNT_ID=x
+CHALLENGE_PK=1
```
deshraj: Please change the env variable name to something more meaningful.

scripts/workers/submission_worker.py
```diff
@@ -641,18 +656,29 @@ def main():
     if challenge_pk:
         q_params["pk"] = challenge_pk
 
-    challenges = Challenge.objects.filter(**q_params)
-    for challenge in challenges:
-        load_challenge(challenge)
+    if settings.DEBUG or settings.TEST:
+        if settings.LIMIT_CONCURRENT_SUBMISSION_PROCESSING:
```
deshraj: Might be a better idea to put this in env variable and then load from there.

scripts/workers/submission_worker.py
```diff
@@ -641,18 +656,29 @@ def main():
...
+            if not challenge_pk:
+                logger.exception(
+                    "Please add CHALLENGE_PK as environment variable"
```
deshraj: Please be more explicit in the error message. If I am a new developer, then I would not know what is going on even after seeing this error.

docker/dev/docker.env
RishabhJain2018: I think we should not change it now as it'll involve changes at various places for e.g. in deployment script, production, staging env variables files etc.

PR #2260, opened 2019-04-03
Frontend: Update landing page

deshraj:
This PR improves the home page of EvalAI. We will continue to bring in more UI improvements over the next couple of days. Keep an eye out. :) 
screencapture-localhost-8888-2019-04-03-04_50_48


deshraj:
Updated screenshot. 

screencapture-localhost-8888-2019-04-03-05_37_49

PR #2270, opened 2019-04-05
Fixes #2269: Remove django.setup() from remote_submission_worker.

### LINKED ISSUE #2269: django.setup() not required for Remote submission worker.
Currently, the remote worker doesn't use the Django app directly as the submission worker does, and only references the models through the public API.

And also the error: `django.core.exceptions.ImproperlyConfigured: Requested setting LOGGING_CONFIG, but settings are not configured. You must either define the environment variable DJANGO_SETTINGS_MODULE or call settings.configure() before accessing settings.` is shown while starting the worker.
**Deliverables:**
- [x] Remove the statement `django.setup()` from `remote_submission_worker.py`.

@RishabhJain2018 @deshraj

KhalidRmb:
Fix #2269


KhalidRmb:
@RishabhJain2018 @deshraj PTAL.

PR #2272, opened 2019-04-06
Fix #2248: Prevent the user to edit the uploaded filename

### LINKED ISSUE #2248: Prevent the user to edit the uploaded filename
### Observed Behaviour
Currently a user is able to edit the uploaded file name during submission to a challenge or when uploading zip file to host a challenge.

### Expected Behaviour
DIsable user to edit the filename or enter any text in the field.

@RishabhJain2018 @deshraj 

Capture

ThongVoHien:
Fix #2248: Remove the option of changing the name of uploaded file


RishabhJain2018:
Closing this PR since the issue is already resolved and the PR contains code which is not relevant. Thanks for the effort though @ThongVoHien.

PR #2277, opened 2019-04-10
Fix worker to load multiple challenges in dev env

RishabhJain2018:
This PR fixes -
1. Removes `CHALLENGE_PK` from docker.env file to load multiple challenges in a single worker for dev env.
2. Add a check if a module is created or not for a challenge and then creates a module for another challenge while loading multiple challenges in dev env.


scripts/workers/submission_worker.py
```diff
@@ -38,8 +39,8 @@
     "DJANGO_SETTINGS_MODULE", "settings.dev"
 )
 DJANGO_SERVER = os.environ.get("DJANGO_SERVER", "localhost")
-LIMIT_CONCURRENT_SUBMISSION_PROCESSING = os.environ.get(
-    "LIMIT_CONCURRENT_SUBMISSION_PROCESSING"
+LIMIT_CONCURRENT_SUBMISSION_PROCESSING = eval(
```
deshraj: Instead of `eval()`, it would definitely be a better idea to do `int()` since this is always going to be integer. By doing `eval()`, you are opening doors for bugs later in the codebase.

scripts/workers/submission_worker.py
```diff
@@ -264,7 +266,11 @@ def load_challenge(challenge):
     # make sure that the challenge base directory exists
     create_dir_as_python_package(CHALLENGE_DATA_BASE_DIR)
     phases = challenge.challengephase_set.all()
-    extract_challenge_data(challenge, phases)
+    challenge_module = extract_challenge_data(challenge, phases)
+    if isinstance(challenge_module, ModuleType):
+        return
+    else:
+        sys.exit(1)
```
deshraj: Why exiting the process? So basically what will happen here is that if the module is not loaded then the process will exit without showing any error. This doesn't sound right to me.

scripts/workers/submission_worker.py
```diff
@@ -249,6 +250,7 @@ def extract_challenge_data(challenge, phases):
             CHALLENGE_IMPORT_STRING.format(challenge_id=challenge.id)
         )
         EVALUATION_SCRIPTS[challenge.id] = challenge_module
+        return challenge_module
```
deshraj: Sorry but I don't think I understand exactly how returning the challenge module here will help solve the problem of loading multiple modules. Can you please elaborate more on how this is solving the problem?

RishabhJain2018: That was a temporary hack but I found a permanent fix for this while loading the modules dunamically: https://docs.python.org/3/library/importlib.html#importlib.invalidate_caches

RishabhJain2018:
Closing this PR.

PR #2280, opened 2019-04-12
Fix #2220:Add three languages to challenge overview, Evaluation Criteria, Terms and Condition

### LINKED ISSUE #2220: Add support for displaying challenges in different languages
## Current Scenario
We display all the content and the challenges in one language only i.e. English.

## Expected Scenario
The user should be able to choose the language in which the content should be displayed.

## Hint:
https://www.toptal.com/angular-js/internationalize-your-angularjs-app

sarthak212:
languages

@RishabhJain2018 Right Now I am using Yandex API to convert these content to other languages but I think google API is better than this if you have another API in your mind please tell me and how many languages do we have to add furthermore and do we have to convert keywords like Challenge name, Overview, Evaluation criteria etc to different languages.


galipremsagar:
I see that Yandex translate api is being used, and since it is a free version we will be limited after a certain number of characters(see below).
Screenshot 2019-05-01 at 11 50 42 PM

Hence this is not a feasible approach for us. This PR can be closed @RishabhJain2018 .

RishabhJain2018:
Thanks for the efforts @sarthak212.

PR #2284, opened 2019-04-17
Fix update submission API for TypeError in result

RishabhJain2018:
**Description:**

The update submission API will give an error if the data in the `result` key is not of type string.

### DIFF
--- a/apps/jobs/views.py
+++ b/apps/jobs/views.py
@@ -778,9 +778,10 @@ def update_submission(request, challenge_pk):
         if successful_submission:
             try:
                 results = json.loads(submission_result)
-            except ValueError:
+            except (ValueError, TypeError) as exc:
                 response_data = {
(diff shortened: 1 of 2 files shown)


apps/jobs/views.py
```diff
@@ -778,9 +778,10 @@ def update_submission(request, challenge_pk):
         if successful_submission:
             try:
                 results = json.loads(submission_result)
-            except ValueError:
+            except (ValueError, TypeError) as exc:
                 response_data = {
-                    "error": "`result` key contains invalid data. Please try again with correct format!"
+                    "message": "`result` key contains invalid data. Please try again with correct format.",
+                    "error": str(exc),
```
deshraj: Why are we sending two keys? I thought we planned on keeping one key in the response_data dict.

If success, then key will be `message` and if there is error, we will have the key `error`

PR #2297, opened 2019-04-30
Fix #2248: Disable edit for uploaded submission filename

### LINKED ISSUE #2248: Prevent the user to edit the uploaded filename

thanos-pakou:
Fix #2248 
Prevents the user from editing the file name or entering any text in the input field (by making input field "readonly").


RishabhJain2018:
Thanks for the fix @thanos-pakou!

PR #2316, opened 2019-05-18
Front-end Unit Tests: Challenge controller

Sanji515:
PR contains the front-end tests of the `Challenge Controller`


Sanji515:
@RishabhJain2018 @galipremsagar Please review

galipremsagar:
@Sanji515 , did you introduce Changes to Challenge controller on purpose ? Asking this to confirm if it was introduced due to some reason?

Sanji515:
> Update: looks like you did refactoring?

yes, because if some functions are defined below the call so in unit tests it is giving me error that it is not defined to that call

Apart from that did you change anything?

no, only initialized one thing `vm.team= {}`

Sanji515:
@galipremsagar I've resolved the above errors, please review

RishabhJain2018:
@Sanji515 Please resolve the conflicts here.

Issue #2324, opened 2019-05-19
Wrong input field when updating existing participant teams 


nataliavramidou:
**Observed behavior**
When someone tries to edit an existing participant team' s name or url, the value passes also at the input fields (name, url) of the create new participant team form.
![issue](https://user-images.githubusercontent.com/15125284/57982730-0c132900-7a52-11e9-80e5-b6ada36ad912.png)

**Excpected behavior**
When editing existing participant team' s name or url, the only input fields that should show the change are the fields name and url fields of the mdDialog Box.

RishabhJain2018:
Closing this for now.

PR #2326, opened 2019-05-19
Fix wrong input fieldwhen updating an existing participant team' s name or url

nataliavramidou:
Fix:  #2324 
I changed the input field variables of the mdDialog so that when someone tries to edit the existing name or url of a participant team, the values won't pass automatically in the input fields of the form that creates new participant team


RishabhJain2018:
Good Catch @nataliavramidou. Thanks for fixing it :)

RishabhJain2018:
@nataliavramidou Can you please fix the Travis Build?

PR #2335, opened 2019-05-21
Coverage reports on front-end unit tests

Sanji515:


Sanji515:
@galipremsagar @RishabhJain2018 Please review

galipremsagar:
@Sanji515 , It says we have no coverage report?? I think we need to fix that aswell?

Sanji515:
> @Sanji515 , It says we have no coverage report?? I think we need to fix that aswell?

It's saying because it don't have any previous data to match with in master branch, so when it will be merged then it will show diff coverage report for all other PR's

galipremsagar:
> > @Sanji515 , It says we have no coverage report?? I think we need to fix that aswell?
> 
> It's saying because it don't have any previous data to match with in master branch, so when it will be merged then it will show diff coverage report for all other PR's

Cool 👍

galipremsagar:
@Sanji515, can you update your branch so that we can merge it. This looks good to me.

galipremsagar:
looks good. Please merge this @RishabhJain2018

PR #2338, opened 2019-05-23
Fix #2070: Automate the process of sending Emails to challenge hosts

### LINKED ISSUE #2070: Automate the process of sending Emails to challenge hosts
## Current Scenario

Whenever a user creates a challenge on EvalAI, then EvalAI admin has to send the email manually saying
```
Hi,

We noticed that you tried to create a challenge on EvalAI recently and you could not create it due to some reason. Please let us know if you need help in creating the challenge configuration for your challenge. We would be more than happy to help in creating the challenge.

Thanks,
EvalAI Team

```

## Deliverables

We would like to automate this process of sending the emails. We would like to use Send Grid API for sending the emails.

So, ideally, the flow should be - 
- [ ] A user creates a challenge on EvalAI
- [ ] An Email is sent to the user on the Email Id registered on EvalAI with the above content.

Please let me know if you have doubts about it.

Sanji515:
Fixes #2070 

Currently I've created one custom HTML templates with `SendGrid` where we can dynamically change the required variables. Of course we can use another template from the SendGrid if we want to. Here's the screenshot of the template:

sendgrid-template

**Variables to change dynamically according to the template:**

- `subject`: - Subject of the email
- `username`: - Username of the challenge host
- `message`: - Content of the email body
- `is_button`: - Boolean variable whether we want a button at bottom of the template
- `button_text`: - name for the same button
- `button_url`: - href url for the same button

**Rest variables are for the footer of email template**

- `aboutus_url`: - evalai about us page url
-  `contactus_url`: - evalai contact page url
 - `ourteam_url`: - evalai our team page url
 - `get_involved_url`: - evalai get involved page url
 - `privacy_policy_url`: - evalai privacy policy page url
 - `cloudcv_url`: - cloudcv home page page url
 - `evalai_url`: - evalai evalai home page url


Sanji515:
@RishabhJain2018 @galipremsagar Please review

apps/base/utils.py
```diff
@@ -103,7 +103,7 @@ def send_email(
     """
     try:
         sg = sendgrid.SendGridAPIClient(
-            apikey=os.environ.get("SENDGRID_API_KEY")
+            apikey=(os.environ.get("SENDGRID_API_KEY"))
```
galipremsagar: apikey now expects a tuple?

Sanji515: no no actually it doesn't matter even if we don't put those parenthesis, if you want then I can remove them

.gitignore
```diff
@@ -64,3 +64,4 @@ ssl/
 out/
 .idea_modules/
 
+sendgrid.env
```
galipremsagar: I don't see this file anywhere? Remove it? or any specific reason you put it?

apps/base/utils.py
galipremsagar:
```suggestion
            apikey=os.environ.get("SENDGRID_API_KEY")
```
Okay, let's not make it a tuple. Lets keep it as is.

.gitignore
Sanji515: @galipremsagar  actually when I was setting the Environment then I exported it to sendgrid.env and then ran command `source ./sendgrid.env`

galipremsagar:
```suggestion
```
Ah, okay understood. Then we can remove this as we don't have it in our codebase.

apps/base/utils.py
Sanji515: @galipremsagar done

.gitignore
Sanji515: @galipremsagar done

PR #2339, opened 2019-05-24
Add slug field for each challenge to have different URL for each challenge.

Sanji515:
- Used `slug` field of challenge model which is created by slugifying the `challenge-title` and `challenge-id` for having different url for each challenge

- As most of the API's are called based on the `challenge-pk`, therefore in `challengeCtrl` and `featuredChallengeCtrl` extracted the `id` from the slug url and set it back to `vm.challengeId` variable as it was before


frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -10,7 +10,8 @@
 
     function ChallengeCtrl(utilities, loaderService, $scope, $state, $http, $stateParams, $rootScope, Upload, $interval, $mdDialog, moment) {
         var vm = this;
-        vm.challengeId = $stateParams.challengeId;
+        var slugUrl = ($stateParams.challengeSlug).replace(/\'/g, '').split(/(\d+)/).filter(Boolean);
+        vm.challengeId = slugUrl[slugUrl.length - 1];
```
galipremsagar:
```suggestion
        vm.challengeId = slugUrl.split("-").pop();
```

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -10,7 +10,8 @@
 
     function ChallengeCtrl(utilities, loaderService, $scope, $state, $http, $stateParams, $rootScope, Upload, $interval, $mdDialog, moment) {
         var vm = this;
-        vm.challengeId = $stateParams.challengeId;
+        var slugUrl = ($stateParams.challengeSlug).replace(/\'/g, '').split(/(\d+)/).filter(Boolean);
```
galipremsagar:
```suggestion
        var slugUrl = $stateParams.challengeSlug;
```

frontend/src/js/controllers/featuredChallengeCtrl.js
```diff
@@ -10,7 +10,8 @@
 
     function FeaturedChallengeCtrl(utilities, loaderService, $scope, $state, $http, $stateParams, moment) {
         var vm = this;
-        vm.challengeId = $stateParams.challengeId;
+        var slugUrl = ($stateParams.challengeSlug).replace(/\'/g, '').split(/(\d+)/).filter(Boolean);
```
galipremsagar:
```suggestion
        var slugUrl = $stateParams.challengeSlug;
```

frontend/src/js/controllers/featuredChallengeCtrl.js
```diff
@@ -10,7 +10,8 @@
...
+        vm.challengeId = slugUrl[slugUrl.length - 1];
```
galipremsagar:
```suggestion
        vm.challengeId = slugUrl.split("-").pop();
```

Sanji515: done

frontend/src/js/controllers/featuredChallengeCtrl.js
Sanji515: done

frontend/src/js/controllers/challengeCtrl.js
Sanji515: done

frontend/src/js/controllers/challengeCtrl.js
Sanji515: done

RishabhJain2018:
@Sanji515 I think we should also take into consideration that we have data in the database which doesn't have the slug value set. Hence removing [this](https://github.com/Cloud-CV/EvalAI/pull/2339/files#diff-14a845646ed48757b5c7798b7ccdddb6L13) line will not let any challenge open on production.
Moreover, we also need to modify the places from where a challenge is opened so that they redirect to the slug link.

cc: @galipremsagar

PR #2344, opened 2019-05-26
 Fix #2210: Add a feature to change the status of a submission by challenge host

### LINKED ISSUE #2210: Add a feature to change the status of a submission by challenge host
## Current Scenario
If a challenge host wants to cancel the submission status of submission, then they can't do it. They have to approach EvalAI admin in order to do it.

## Expected Scenario
The challenge host should be able to change the submission status using UI.

Sanji515:
Fix #2210 

- After clicking on edit status button changed default showing over dropdown from static `status` to dynamic `<current submission status>`

- Added tick icon over the selected or current submission status on dropdown to easily understand the current checked submission status

- Removed the added API `change_submission_status`, instead used the already existing API `update_submission`

here's the screen-cast of the working

Screencast from Sunday 26 May 2019 11_44_26  IST


apps/jobs/views.py
```diff
@@ -945,7 +945,16 @@ def update_submission(request, challenge_pk):
         submission_pk = request.data.get("submission")
         submission_status = request.data.get("submission_status", "").lower()
         submission = get_submission_model(submission_pk)
-        if submission_status not in [Submission.RUNNING]:
+        if 'change_submission_status' in request.data:
```
RishabhJain2018: Please add tests for this code.

PR #2350, opened 2019-05-30
Unit tests for remote_submission_worker.

KhalidRmb:
12 test cases for `remote_submission_worker`.


KhalidRmb:
@Ram81 @RishabhJain2018 Please take a look.

Ram81:
@KhalidRmb can you check why the travis build is failing

KhalidRmb:
> @KhalidRmb can you organise the tests in multiple classes like we did it here https://github.com/Cloud-CV/EvalAI/blob/master/tests/unit/challenges/test_views.py
Sure, a class for each type of method. I'll do that.

KhalidRmb:
@Ram81 Please review.

tests/unit/remoteworker/test_remote_worker.py
```diff
@@ -0,0 +1,126 @@
...
+    def test_make_request_get(self, mock_make_request):
+        url = "/test/url"
+        headers = {"Authorization": "Token test_token"}
+        make_request(url, "GET")
+        mock_make_request.get.assert_called_with(url=url, headers=headers)
+
+    def test_make_request_put(self, mock_make_request):
+        url = "/test/url"
+        headers = {"Authorization": "Token test_token"}
+        data = {"test": "data"}
```
Ram81: @KhalidRmb you can move the redundant variables in a common base class

tests/unit/remoteworker/test_remote_worker.py
```diff
@@ -0,0 +1,126 @@
...
+
+@mock.patch("scripts.workers.remote_submission_worker.return_url_per_environment")
+@mock.patch("scripts.workers.remote_submission_worker.make_request")
+class APICallsTestClass(TestCase):
+
+    @mock.patch("scripts.workers.remote_submission_worker.QUEUE_NAME", "evalai_submission_queue")
+    def test_get_message_from_sqs_queue(self, mock_make_request, mock_url):
+        url = "/api/jobs/challenge/queues/evalai_submission_queue/"
+        get_message_from_sqs_queue()
+        mock_url.assert_called_with(url)
```
Ram81: make `url` as a class variable otherwise there is no use of classes plus there is lot of redundant code

tests/unit/remoteworker/test_remote_worker.py
KhalidRmb: @Ram81 Sure. Will make the changes.

tests/unit/remoteworker/test_remote_worker.py
KhalidRmb: Done!

tests/unit/remoteworker/test_remote_worker.py
```diff
@@ -0,0 +1,116 @@
...
+        mock_make_request.patch.assert_called_with(url=self.url, headers=self.headers, data=self.data)
+
+
+@mock.patch("scripts.workers.remote_submission_worker.QUEUE_NAME", "evalai_submission_queue")
+@mock.patch("scripts.workers.remote_submission_worker.return_url_per_environment")
+@mock.patch("scripts.workers.remote_submission_worker.make_request")
+class APICallsTestClass(BaseTestClass):
+
+    def test_get_message_from_sqs_queue(self, mock_make_request, mock_url):
+        self.url = "/api/jobs/challenge/queues/evalai_submission_queue/"
```
Ram81: @KhalidRmb why do we still have `urls` hardcoded? can't we just define `get` methods in base class for fetching the required url?

KhalidRmb: I'm not sure what you mean. I'm hard coding them because a different url is used for each test case.

Ram81: So what I want is `get` methods in base class to fetch the required URL, hardcoding them in each test method is not a good practice. Methods in base class can be reused for other methods. Did you get my point?

Ram81: @RishabhJain2018 your thoughts!

Ram81:
@RishabhJain2018 other than the get methods part rest of the PR LGTM 👍

Ram81:
@KhalidRmb please add the get methods for URLs and we are good to go for this PR

KhalidRmb:
> @KhalidRmb please add the get methods for URLs and we are good to go for this PR

Done.

tests/unit/remoteworker/test_remote_worker.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def test_get_message_from_sqs_queue(self, mock_make_request, mock_url):
+        url = self.get_message_from_sqs_queue_url("evalai_submission_queue")
+        get_message_from_sqs_queue()
+        mock_url.assert_called_with(url)
+        url = mock_url(url)
+        mock_make_request.assert_called_with(url, "GET")
+
+    def test_delete_message_from_sqs_queue(self, mock_make_request, mock_url):
+        test_receipt_handle = "MbZj6wDWli+JvwwJaBV+3dcjk2YW2vA3+STFFljTM8tJJg6HRG6PYSasuWXPJB+Cw"
```
RishabhJain2018: @KhalidRmb What is the expiry date of this receipt handle? I think it will change after some time and the test will start failing. Did you already check that it is constant every time?

KhalidRmb: @RishabhJain2018 It works since I'm mocking out the make_request function. Since I'm unit testing it separately anyhow.

On second thought, I think it would be better if I add a test case using mock_ecs alone. Will do it 👍

PR #2354, opened 2019-06-04
Remove redundant try/except bock from submission worker.

KhalidRmb:
This `try/except` block is unnecessary as the one above it makes sure of  the existence of the `dataset_split`.


KhalidRmb:
@Ram81 @RishabhJain2018 PTAL.

Ram81:
@RishabhJain2018  can you cross check this once

PR #2358, opened 2019-06-08
Add feature to ban the participants for the inappropriate behaviour

Sanji515:
- Added the field `banned_email_ids` in Challenge model.

- Restricted the user to participate and make submissions for the challenge.

- And also removed the leaderboard entry for the particular participant team.


apps/jobs/views.py
```diff
@@ -206,6 +206,14 @@ def challenge_submission(request, challenge_id, challenge_phase_id):
             }
             return Response(response_data, status=status.HTTP_403_FORBIDDEN)
 
+        all_participants_email = participant_team.get_all_participants_email()
+        for i in all_participants_email:
```
RishabhJain2018: `i` -> `participant_email`

apps/jobs/views.py
```diff
@@ -206,6 +206,14 @@ def challenge_submission(request, challenge_id, challenge_phase_id):
...
+            if i in challenge.banned_email_ids:
```
RishabhJain2018: `i` -> `participant_email`

apps/jobs/views.py
```diff
@@ -206,6 +206,14 @@ def challenge_submission(request, challenge_id, challenge_phase_id):
...
+                response_data = {
+                    "error": "Sorry, You cannot make submissions in this challenge due to the inappropriate behaviour"
```
RishabhJain2018: Please use the above error message.

apps/jobs/views.py
Sanji515: @RishabhJain2018 done

apps/jobs/views.py
Sanji515: done

apps/jobs/views.py
Sanji515: done

PR #2365, opened 2019-07-03
Changed the configuration of stale.yml

Sanji515:

- Changed the configuration of `stale.yml` because issues are created to get resolved and as they are not resolved so they should not be closed by the stale bot


Sanji515:
@deshraj @RishabhJain2018 Please review

PR #2368, opened 2019-07-04
Add ordering by pk in team page

RishabhJain2018:


deshraj:
I don't think the ordering should be based on pk. Can you please tell me the reasoning behind doing this ?

RishabhJain2018:
I added it so as to render the names on the team page in a particular order.

PR #2369, opened 2019-07-04
Add position in the team model

RishabhJain2018:


apps/web/models.py
```diff
@@ -48,10 +48,12 @@ class Team(models.Model):
...
     team_type = models.CharField(choices=TEAM_TYPE_OPTIONS, max_length=50)
+    position = models.PositiveIntegerField(default=0)
 
     def __str__(self):
         return self.name
 
     class Meta:
         app_label = "web"
         db_table = "teams"
+        unique_together = (("position", "team_type"),)
```
deshraj: This will fail since you have setup default as 0. You will have to skip this check or write a script that assigns values to all the existing entries in the database.

RishabhJain2018: yeah..this will conflict with the existing data in the database.

PR #2370, opened 2019-07-04
Add allowed hosts for the surge demo

RishabhJain2018:
This PR adds `.surge.sh` in allowed hosts for EvalAI-ngx demos.

### DIFF
--- a/settings/staging.py
+++ b/settings/staging.py
@@ -1,6 +1,6 @@
 from .prod import *  # noqa: ignore=F405
 
-ALLOWED_HOSTS = ["staging-evalai.cloudcv.org"]
+ALLOWED_HOSTS = ["staging-evalai.cloudcv.org", ".surge.sh"]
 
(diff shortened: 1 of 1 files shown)


settings/staging.py
```diff
@@ -1,6 +1,6 @@
 from .prod import *  # noqa: ignore=F405
 
-ALLOWED_HOSTS = ["staging-evalai.cloudcv.org"]
+ALLOWED_HOSTS = ["staging-evalai.cloudcv.org", ".surge.sh"]
```
deshraj: Adding `.surge.sh` means that anyone who has the EvalAI frontend running on surge through their account will be able to tie their frontend with EvalAI staging backend which is non-ideal. The staging server is meant to replicate the behavior of the production server (without any external interference). Instead, I think we should pair surge with localhost:8000 instead of EvalAI staging backend.

RishabhJain2018: I agree. Thanks for catching this.

PR #2376, opened 2019-07-07
RL Challenge Submission API

vkartik97:
## Deliverables
- ~`scripts/tools/rl_kube_warmup.sh`: Script that can configure the environment before running the worker~
- ~`scripts/workers/worker_util.py`: Util that can be used by `rl-worker` to interact with api~
- For Normal Submissions, only 
```
"challenge_pk": challenge_id,
"phase_pk": challenge_phase_id,
"submission_pk": submission.id
```
will be ingested into the challenge specific queues.
Proposed changes that is expected being tested and soon be pushed:
1. For RL Submissions, 
```
"challenge_pk": challenge_id,
"phase_pk": challenge_phase_id,
"submission_pk": submission.id,
"repository_uri": repository_uri
```
~2. A new field in `phase` model, evaluator_uri to get images from ECR for Evaluation~


scripts/workers/worker_util.py
```diff
@@ -0,0 +1,104 @@
+import logging
+import requests
+
+logger = logging.getLogger(__name__)
+
+
+URLS = {
+    "get_message_from_sqs_queue": "/api/jobs/challenge/queues/{}/",
```
rishabhbhardwaj: Can you make the API resource plural i.e "challenges" instead of "challenge"' ?

RishabhJain2018:
@vkartik97 Please add the PR description in detail.

scripts/tools/rl_kube_warmup.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
+# Usage:
+# bash rl_worker_kube_warmup.sh [FLAGS]
```
rishabhbhardwaj: @vkartik97 Can you add an example usage with the flags? 
Also, give a short description of what this script does.

scripts/tools/rl_kube_warmup.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
```
vkartik97: Deleted this file as it was part of other PR

RishabhJain2018: Perfect!

scripts/workers/worker_util.py
```diff
@@ -0,0 +1,104 @@
+import logging
```
vkartik97: Deleted this file as it was part of other PR

RishabhJain2018: Perfect!

apps/jobs/sender.py
```diff
@@ -55,27 +55,20 @@ def get_or_create_sqs_queue(queue_name):
...
 
 
-def publish_submission_message(challenge_pk, phase_pk, submission_pk):
+def publish_submission_message(message):
     """
     Args:
-        challenge_pk: Challenge Id
-        phase_pk: Challenge Phase Id
-        submission_pk: Submission Id
+        message
```
RishabhJain2018: Can you please write its type and what all keys it contains.

apps/jobs/views.py
```diff
@@ -276,14 +276,31 @@ def challenge_submission(request, challenge_id, challenge_phase_id):
                 "request": request,
             },
         )
+        message = dict()
```
RishabhJain2018: Please make this corresponding change in the admin action for re-running a submission.

apps/jobs/sender.py
vkartik97: Added
```py
        message: A Dict with following keys
            - "challenge_pk": int
            - "phase_pk": int
            - "submission_pk": int
```

apps/jobs/views.py
```diff
@@ -276,14 +276,31 @@ def challenge_submission(request, challenge_id, challenge_phase_id):
...
+        if challenge.is_docker_based:
```
RishabhJain2018: 1. same change here: https://github.com/Cloud-CV/EvalAI/blob/master/apps/jobs/tasks.py#L76
2. same change here: https://github.com/Cloud-CV/EvalAI/blob/master/apps/jobs/views.py#L1030

apps/jobs/sender.py
```diff
@@ -55,27 +55,23 @@ def get_or_create_sqs_queue(queue_name):
...
+def publish_submission_message(message):
     """
     Args:
-        challenge_pk: Challenge Id
-        phase_pk: Challenge Phase Id
-        submission_pk: Submission Id
+        message: A Dict with following keys
+            - "challenge_pk": int
+            - "phase_pk": int
+            - "submission_pk": int
```
RishabhJain2018: Please add a comment that if the challenge is docker based then it will contain the `submitted image URI`.

apps/jobs/views.py
vkartik97: I changes are reflected here: https://github.com/Cloud-CV/EvalAI/pull/2376/files#diff-eee3bef376ba38e2ff3289821573aa45R72-R76
Please correct me if I misunderstood this review

apps/jobs/sender.py
vkartik97: Thanks for catching this 🙌

apps/jobs/views.py
RishabhJain2018: Ohh..my bad. I didn't see the braces. Can you please create a variable there and pass it in the function.

apps/jobs/sender.py
vkartik97: Done

apps/jobs/sender.py
```diff
@@ -55,27 +55,24 @@ def get_or_create_sqs_queue(queue_name):
...
+            - "submitted_image_uri": string, Only for Docker Based Challenge submissions
```
RishabhJain2018: str, (only available when the challenge is a code upload challenge)

apps/jobs/views.py
vkartik97: Hi,
Do you mean like this(from https://github.com/Cloud-CV/EvalAI/pull/2376/files#diff-42c752186f5f6b2ed96b92afa242a6f3R279-R302), pointed the changes using comments:
```py
######### CHANGE 1 <--
 message = {                                                
                "challenge_pk": challenge_id,
                "phase_pk": challenge_phase_id
        }
        if challenge.is_docker_based:
            try:
                file_content = json.loads(
                    request.FILES['input_file'].read()
                )
                message["submitted_image_uri"] = file_content["submitted_image_uri"]
            except:
                response_data = {
                    "error": "Error in deserializing submitted_image_uri from submission file"
                }
                return Response(
                    response_data, status=status.HTTP_400_BAD_REQUEST
                )
        if serializer.is_valid():
            serializer.save()
            response_data = serializer.data
            submission = serializer.instance
            ###########  CHANGE 2 <--
            message["submission_pk"]: submission.id
            # publish message in the queue
            publish_submission_message(
                challenge_id, challenge_phase_id, submission.id
            )
            publish_submission_message(message)
```

apps/jobs/sender.py
vkartik97: Done

apps/jobs/views.py
RishabhJain2018: No No, the change here is fine but in apps/jobs/admin.py file we will re-run the code upload challenge submission from admin then it will break because it won't have `submitted_image_uri` as a key.
So, I was suggesting to create a message dict there and then append the keys to it based upon the challenge type instead of directly passing the dict into the function.

I hope this is clear, right? Feel free to ask if you've any doubt.

apps/jobs/views.py
```diff
@@ -1027,7 +1044,25 @@ def re_run_submission(request, submission_pk):
...
+    }
+
+    if submission.challenge_phase.challenge.is_docker_based:
+        try:
+            response = requests.get(submission.input_file)
+        except Exception as e:
+            response_data = {
+                "error": "Failed to get submission input file with error: {0}".format(e)
+            }
+            return Response(response_data, status=status.HTTP_500_INTERNAL_SERVER_ERROR)
```
RishabhJain2018: The error code here should be 404 NOT FOUND.

vkartik97: If response for `re_run_submission` is `404`, it might mean something the browser/client was not couldn't endpoint. I kept here 500 endpoint is further using http calls, which might return 40x errors.
Please let me know if it makes sense.

RishabhJain2018: I see..sure!

PR #2392, opened 2019-07-17
Adding Environment URL field in Phase model

vkartik97:
Added `environment_url` field in `ChallengePhase` model, that is the repository url for environment that will be used by RL Worker.
The worker can get the `environment_url` using `api/challenges/phase/environment/<phase-slug>`.


vkartik97:
@RishabhJain2018 @rishabhbhardwaj @deshraj @fristonio Please review

yashdusing:
Hey @vkartik97 ! I read your blog and was wondering if you tried [this](https://developer.nvidia.com/kubernetes-gpu) ?
Also, you mentioned of a problem regarding GPU orchestration. Could you please elaborate ?

vkartik97:
@yashdusing 
>Hey @vkartik97 ! I read your blog and was wondering if you tried this ?

Yes, this is already enabled in [EC2 P2](https://aws.amazon.com/ec2/instance-types/p2/) but I wanted to achieve sharing of GPU resources among the environment containers in multiple pods.

> Also, you mentioned of a problem regarding GPU orchestration. Could you please elaborate ?

GPU Orchestration on P2 is working as expected, the problem is just the sharing extension. I tried it on 1 node minikube, which didn’t  give expected behaviour.

yashdusing:
@vkartik97 I was wondering if [this](https://ray.readthedocs.io/en/latest/resources.html) can be used on top of k8s ? It has fractional gpu support and also runs on top of k8s.

vkartik97:
@yashdusing Really thanks for researching on the topic.
I did try ray-project at the end of Phase 2, and I ran into similar this issue: [[ray] ray misuse gpu in docker container](https://github.com/ray-project/ray/issues/5245), while trying running tensorflow training on `nividia-docker`, one of the dependency for Kubernetes GPU Scheduling. I believe that the issue is due to nvidia-docker, but I need to spike more to come to any conclusion.
Again, thanks for working on this :)

scripts/seed.py
```diff
@@ -175,11 +175,15 @@ def create_challenge_phases(challenge, number_of_phases=1):
     challenge_phases = []
```
RishabhJain2018: Can you please open up a new PR with the changes in this file?

vkartik97: Done

PR #2393, opened 2019-07-17
Backend: Add feature to get and post the user email to notify them

Sanji515:

- [x] Added `Subscribers` model and `SubscribeSerializer` serializer with the `email` field

- [x] Updated the `admin.py` to show up on the admin panel

- [x] Added `GET` and `POST` request to retrieve and post the emails of user


apps/web/admin.py
```diff
@@ -2,7 +2,7 @@
 
 from base.admin import ImportExportTimeStampedAdmin
 
-from .models import Contact, Team
+from .models import Contact, Team, Subscribers
```
RishabhJain2018: Please import it alphabetically.

Sanji515: @RishabhJain2018 done here and at other places also

PR #2395, opened 2019-07-18
Add feature to update the leaderboard data when re-running the submission

Sanji515:
**Issue:**

When re-running the submission then instead of updating the existing leaderboard data, it creates a new entry into the `LeaderboardData` table

**Fix:**

- Added a check for checking if the submission already exists. If it exists then update the exisitng leaderboard data.


RishabhJain2018:
Closing this due to #2749

PR #2404, opened 2019-07-21
Features to deploy worker on Fargate.

KhalidRmb:
Adds the following features:
- Start
- Stop
- Scale
- Restart
- Delete
- Automatic restart on updating Evaluation script or Annotation files.


KhalidRmb:
@RishabhJain2018 @Ram81

RishabhJain2018:
@KhalidRmb Please add the environment variables as discussed on the call.

apps/challenges/admin.py
```diff
@@ -1,7 +1,17 @@
-from django.contrib import admin
+from django import forms
+from django.contrib import admin, messages
+from django.contrib.admin.helpers import ActionForm
 
 from base.admin import ImportExportTimeStampedAdmin
 
+from .aws_utils import (
+    start_workers,
```
deshraj: Can we please order these imports in lexicographical order?

apps/challenges/admin.py
```diff
@@ -15,6 +25,11 @@
 )
 
 
+class UpdateNumOfWorkersForm(ActionForm):
+    label = " Number of Workers. (Enter a whole number while scaling. Otherwise, ignore.)"
```
deshraj: Please replace with:

> label = "Number of workers. (Enter a whole number while scaling. Otherwise, ignore.)"

apps/challenges/admin.py
```diff
@@ -44,6 +61,71 @@ class ChallengeAdmin(ImportExportTimeStampedAdmin):
         "end_date",
     )
     search_fields = ("title", "creator__team_name", "slug")
+    actions = ["start_selected_workers", "stop_selected_workers", "scale_selected_workers", "restart_selected_workers", "delete_selected_workers"]
+    action_form = UpdateNumOfWorkersForm
+
+    def start_selected_workers(self, request, queryset):
+        response = start_workers(queryset)
+        count = response["count"]
+        message = response["message"]
```
deshraj: Can we please club the lines 69 and 70 by writing something like this:

> message, count = response["message"], response["count"]

Please make this change for other methods as well.

apps/challenges/aws_utils.py
```diff
@@ -0,0 +1,542 @@
...
+    "AWS_REGION": os.environ.get("AWS_DEFAULT_REGION", "us-east-1"),
+}
+
+TASK_ROLE_ARN = os.environ.get("TASK_ROLE_ARN", ""),
+TASK_EXECUTION_ROLE_ARN = "arn:aws:iam::{}:role/evalaiTaskExecutionRole".format(aws_keys["AWS_ACCOUNT_ID"])
+
+DJANGO_SETTINGS_DICT = {
+    "DJANGO_SERVER": os.environ.get("DJANGO_SERVER", "localhost"),
+    "DEBUG": settings.DEBUG,
+    "EMAIL_HOST": os.environ.get("EMAIL_HOST", "https://email_host"),
```
deshraj: 1. For the variables that are available in `django.conf.settings`, let's try to import them using `settings.VARIABLE_NAME` instead of `os.environ.get(VARIABLE_NAME)`. 
2. Also, let's try to not put default here since it will break stuff later on. We should raise an exception and send message to sentry if any of the following variables are missing.

apps/challenges/aws_utils.py
```diff
@@ -0,0 +1,542 @@
...
+        "FARGATE"
+    ],
+    "cpu": "1024",
+    "memory": "2048",
+}}
+"""
+
+service_definition = """
+{{
+    "cluster":"gsoc2019",
```
deshraj: let's change the clustername to something else like `evalai-prod-cluster`

apps/challenges/aws_utils.py
```diff
@@ -0,0 +1,542 @@
...
+    "schedulingStrategy":"REPLICA",
+    "deploymentController":{{
+        "type": "ECS"
+    }},
+}}
+"""
+
+update_service_args = """
+{{
+    "cluster":"gsoc2019",
```
deshraj: Let's change the cluster name here as well.

apps/challenges/aws_utils.py
```diff
@@ -0,0 +1,542 @@
...
+    "serviceName":"{service_name}",
+    "taskDefinition":"{task_def_arn}",
+    "desiredCount":1,
+    "clientToken":"{client_token}",
+    "launchType":"FARGATE",
+    "platformVersion":"LATEST",
+    "networkConfiguration":{{
+        "awsvpcConfiguration": {{
+            "subnets": [
+                "subnet-e260d5be",
```
deshraj: Doesn't feel right to hard code the subnet-id here. If we want other people to use this code and run workers on their fargate cluster, it will be really painful for them. 

Ideally, all of this should come from a fargate config file or environment variables. We don't have to do this for this version but would be good to open an issue for this so that we keep an eye on this. (cc: @RishabhJain2018 @Ram81 )

apps/challenges/aws_utils.py
```diff
@@ -0,0 +1,542 @@
...
+    "service":"{service_name}",
+    "desiredCount":num_of_tasks,
+    "taskDefinition":"{task_def_arn}",
+    "forceNewDeployment":{forceNewDeployment}
+}}
+"""
+
+delete_service_args = """
+{{
+    "cluster": "gsoc2019",
```
deshraj: Cluster name change here as well please.

apps/challenges/aws_utils.py
```diff
@@ -0,0 +1,542 @@
...
+        if not challenge.task_def_arn:
+            try:
+                response = client.register_task_definition(**definition)
+                if (response["ResponseMetadata"]["HTTPStatusCode"] == HTTPStatus.OK):
+                    task_def_arn = response["taskDefinition"]["taskDefinitionArn"]
+                    challenge.task_def_arn = task_def_arn
+                    challenge.save()
+                return response
+            except ClientError as e:
+                logger.info(e)
```
deshraj: It should be `logger.exception()` here, no? 

Reference: https://docs.python.org/3/library/logging.html#logging.Logger.exception

apps/challenges/aws_utils.py
```diff
@@ -0,0 +1,542 @@
...
+        definition = service_definition.format(service_name=service_name, task_def_arn=task_def_arn, client_token=client_token)
+        definition = eval(definition)
+        try:
+            response = client.create_service(**definition)
+            if response["ResponseMetadata"]["HTTPStatusCode"] == HTTPStatus.OK:
+                challenge.workers = 1
+                challenge.save()
+            return response
+        except ClientError as e:
+            return e.response
```
deshraj: Let's log the exception here and then return. See above comment.

apps/challenges/aws_utils.py
```diff
@@ -0,0 +1,542 @@
...
+    kwargs = eval(kwargs)
+
+    try:
+        response = client.update_service(**kwargs)
+        if response["ResponseMetadata"]["HTTPStatusCode"] == HTTPStatus.OK:
+            challenge.workers = num_of_tasks
+            challenge.save()
+        return response
+    except ClientError as e:
+        return e.response
```
deshraj: Let's log exception here as well.

apps/challenges/aws_utils.py
```diff
@@ -0,0 +1,542 @@
...
+        response = client.delete_service(**kwargs)
+        if response["ResponseMetadata"]["HTTPStatusCode"] == HTTPStatus.OK:
+            challenge.workers = None
+            challenge.save()
+            client.deregister_task_definition(taskDefinition=challenge.task_def_arn)
+            challenge.task_def_arn = ""
+            challenge.save()
+        return response
+    except ClientError as e:
+            return e.response
```
deshraj: Let's log exception here.

apps/challenges/aws_utils.py
```diff
@@ -0,0 +1,542 @@
...
+
+
+def service_manager(client, challenge, num_of_tasks=None, forceNewDeployment=False):
```
deshraj: Let's try to use underscrore convention only here while listing arguments. Replace `forceNewDeployment` -> `force_new_deployment`

RishabhJain2018:
Thanks, @deshraj for the review.
@KhalidRmb Please incorporate it and then I'll also review it.

apps/challenges/admin.py
```diff
@@ -44,6 +61,71 @@ class ChallengeAdmin(ImportExportTimeStampedAdmin):
...
+
+        if(count == queryset.count()):
```
Ram81: missing space after `if`

apps/challenges/admin.py
```diff
@@ -44,6 +61,71 @@ class ChallengeAdmin(ImportExportTimeStampedAdmin):
...
+            messages.success(request, "{} challenge workers were succesfully started.".format(count))
+            messages.error(request, message)
+    start_selected_workers.short_description = "Start all selected challenge workers."
+
+    def stop_selected_workers(self, request, queryset):
+        response = stop_workers(queryset)
+        count = response["count"]
+        message = response["message"]
+
+        if(count == queryset.count()):
```
Ram81: same here

apps/challenges/aws_utils.py
```diff
@@ -0,0 +1,542 @@
...
+    Returns:
+    dict: The response returned by the create_service method from boto3. If unsuccesful, returns an error dictionary
+    """
+
+    queue_name = challenge.queue
+    service_name = "{}_service".format(queue_name)
+    if challenge.workers is None:  # Verify if the challenge is new (i.e, service not yet created.).
+        if challenge.task_def_arn is "":
+            response = register_task_def_by_challenge_pk(client, queue_name, challenge)
+            if (response["ResponseMetadata"]["HTTPStatusCode"] != HTTPStatus.OK):
```
Ram81: @KhalidRmb it should be `(` everywhere or no where, please make this consistent over whole codebase

apps/challenges/admin.py
```diff
@@ -15,6 +25,11 @@
 )
 
 
+class UpdateNumOfWorkersForm(ActionForm):
+    label = " Number of Workers. (Enter a whole number while scaling. Otherwise, ignore.)"
+    num_of_tasks = forms.IntegerField(initial=-1, label=label, required=False)
```
Ram81: @KhalidRmb the checks for negative values are handled right? did you check whether all corner cases are covered?

KhalidRmb: Only, whole numbers are allowed.

KhalidRmb:
> Overall the PR looks good just some minor changes, nice work @KhalidRmb . Just one more doubt
> 
> 1. In case starting multiple workers does not happen for all workers, let's say only 3 were launched out of 4 (which was expected) what'll happen to other 3 and will we retry for 4th one ?

The 3 will be launched succesfully and the database will be uploaded accordingly. However, the error for the first failure will be reported on the admin page & will be logged. (All workers will be launched unitl the first error is met.)

apps/challenges/aws_utils.py
KhalidRmb: I'm mainly using them when there is more than one condition to verify, for readablity. I'll make all if statements bracketed in that case.

apps/challenges/aws_utils.py
KhalidRmb: But in some cases settings file doesn't have the analogous variable for dev environment. For them, I'll stick to taking them directly as env variables with a default value, so that tests don't break.

KhalidRmb:
@RishabhJain2018 @Ram81 @deshraj Have changed stuff based on your reviews, and also improved on the features to give bettor error messages on admin dashboard.

apps/challenges/models.py
```diff
@@ -119,6 +121,12 @@ def __init__(self, *args, **kwargs):
     cli_version = models.CharField(
         max_length=20, verbose_name="evalai-cli version", null=True, blank=True
     )
+    # The number of active workers on Fargate of the challenge.
+    workers = models.IntegerField(
+        null=True, blank=True, default=None
+    )
+    # The task definition ARN for the challenge, used for updating and creating service.
+    task_def_arn = models.CharField(blank=True, max_length=2048, default="")
```
RishabhJain2018: Please add `null=True`.

docker/prod/docker_staging.env
```diff
@@ -33,3 +33,13 @@ RDS_PASSWORD=x
...
 SENTRY_URL=x
 SENDGRID_API_KEY=x
+
+CLUSTER=cluster
+CPU=0
+EXECUTION_ROLE_ARN=x
+MEMORY=0
+SUBNET_1=x
+SUBNET_2=x
+SUBNET_SECURITY_GROUP=x
```
Ram81: @RishabhJain2018 @KhalidRmb what about the values here? will this be changed at the time of deployment?

RishabhJain2018: yes, this will be modified at the time of deployment.

RishabhJain2018:
@KhalidRmb Can you please resolve the conflicts here?

PR #2405, opened 2019-07-21
Fix Update Submission Status Endpoint

vkartik97:
Fixed `update_submission_status` endpoint


RishabhJain2018:
Thanks, @vkartik97 for fixing this.

deshraj:
That was a really good catch. Thanks @vkartik97 :)

PR #1926, opened 2018-11-25
Add backend to make submissions using URL

RishabhJain2018:

PR #2409, opened 2019-07-26
 Backend: Add feature to make submissions using URL

Sanji515:
Based upon https://github.com/Cloud-CV/EvalAI/pull/1926

- Configure celery to work with the redis as broker

- Added separate `Dockerfile` for the celery in dev and prod and updated the `docker-compose` files

- Updated the `challenge_submission` API to work when user give URL instead of submissions file

- Added the task for downloading and send the submission for evaluation


apps/jobs/views.py
```diff
@@ -234,32 +239,48 @@ def challenge_submission(request, challenge_id, challenge_phase_id):
             >= challenge_phase.max_concurrent_submissions_allowed
         ):
             message = "You have {} submissions that are being processed. \
-                       Please wait for them to finish and then try again."
+                Please wait for them to finish and then try again."
```
RishabhJain2018: Please remove this change.

apps/jobs/views.py
```diff
@@ -234,32 +239,48 @@ def challenge_submission(request, challenge_id, challenge_phase_id):
...
-            },
-        )
-        if serializer.is_valid():
-            serializer.save()
-            response_data = serializer.data
-            submission = serializer.instance
-            # publish message in the queue
-            publish_submission_message(
-                challenge_id, challenge_phase_id, submission.id
+        if 'file_url' in request.data:
```
RishabhJain2018: Please use the condition from here: https://github.com/Cloud-CV/EvalAI/pull/1926/files#diff-42c752186f5f6b2ed96b92afa242a6f3R166

It will be shorter and crisp.

docker-compose-production.yml
```diff
@@ -1,6 +1,11 @@
 version: "3"
 services:
 
+  redis:
```
RishabhJain2018: Why do we need Redis container?

RishabhJain2018: I think we should use https://docs.celeryproject.org/en/latest/getting-started/brokers/sqs.html
I've added it in this PR: https://github.com/Cloud-CV/EvalAI/pull/1926/

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -396,8 +398,13 @@
                     parameters.url = 'jobs/challenge/' + vm.challengeId + '/challenge_phase/' + vm.phaseId + '/submission/';
                     parameters.method = 'POST';
                     var formData = new FormData();
+                    if (vm.isSubmissionUsingUrl) {
+                        formData.append("file_url", vm.fileUrl);
+                    } else {
+                        formData.append("input_file", vm.input_file);
+                    }
+                    console.log(vm.fileUrl);
```
RishabhJain2018: Please remove this console.

apps/jobs/tasks.py
```diff
@@ -0,0 +1,88 @@
...
+    request_method,
+    challenge_id,
+    challenge_phase_id
+):
+    """
+    Download submission file from url and send it for the evaluation
+    """
+    user = User.objects.get(pk=user_pk)
+    challenge = Challenge.objects.get(pk=challenge_id)
+    challenge_phase = ChallengePhase.objects.get(
```
RishabhJain2018: We don't need these extra queries here. Please see this function: https://github.com/Cloud-CV/EvalAI/pull/1926/files#diff-284171ca9611333c9e201afc979c757aR23

apps/jobs/tasks.py
```diff
@@ -0,0 +1,91 @@
...
+    user_pk,
+    request_method,
+    challenge_id,
+    challenge_phase_id
+):
+    """
+    Download submission file from url and send it for the evaluation
+    """
+    user = User.objects.get(pk=user_pk)
+    challenge = Challenge.objects.get(pk=challenge_id)
```
RishabhJain2018: I think we can get the challenge id using the `challenge_phase`. So, there won't be any need for this query.

RishabhJain2018:
@Sanji515 Can you please fix the Travis build here.

apps/jobs/utils.py
```diff
@@ -133,3 +135,36 @@ def get_remaining_submission_for_a_phase(
...
+
+def get_file_from_url(url):
+    """ Get file object from a url """
+
+    BASE_TEMP_DIR = tempfile.mkdtemp()
+    file_name = url.split("/")[-1]
+    file_path = os.path.join(BASE_TEMP_DIR, file_name)
+    file_obj = {}
+    headers = {'user-agent': 'Wget/1.16 (linux-gnu)'}
+    r = requests.get(url, stream=True, headers=headers)
```
RishabhJain2018: Please change variable `r` to `response`.

Sanji515: Done

apps/jobs/tasks.py
Sanji515: But we need this query in order to get the `challenge_phase` object for that particular challenge.

PR #2413, opened 2019-07-29
Tests for aws_utils.

KhalidRmb:
- 24 test cases covering the aws_utils features.
- Migration file for adding in `task_def_arn`,  `null-true` parameter.


KhalidRmb:
@Ram81 @RishabhJain2018 PTAL.

apps/challenges/migrations/0058_add_null_true_for_task_def_arn.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
RishabhJain2018, line 1: Hey @KhalidRmb, Why do we need this file in this PR?

PR #2423, opened 2019-08-09
Add feature for Private Annotation.

KhalidRmb:
Adds feature to pass annotation files as environment variables (as a zip file) to the remote worker container.


KhalidRmb:
@Ram81 @RishabhJain2018 PTAL.

PR #2428, opened 2019-08-15
Add feature to export field affiliation in the csv for each of the participant under view all submissions

Sanji515:

- Added new function in `ChallengeSubmissionManagementSerializer` to get all the participants affiliation.

- Updated tests and also the API for download submission data API with the field affiliation.

### DIFF
--- a/apps/jobs/serializers.py
+++ b/apps/jobs/serializers.py
@@ -90,6 +90,7 @@ class ChallengeSubmissionManagementSerializer(serializers.ModelSerializer):
     challenge_phase = serializers.SerializerMethodField()
     created_by = serializers.SerializerMethodField()
     participant_team_members_email_ids = serializers.SerializerMethodField()
+    participant_team_members_affiliations = serializers.SerializerMethodField()
     created_at = serializers.SerializerMethodField()
     participant_team_members = serializers.SerializerMethodField()
 
@@ -112,6 +113,7 @@ class ChallengeSubmissionManagementSerializer(serializers.ModelSerializer):
             "submission_result_file",
             "submission_metadata_file",
             "participant_team_members_email_ids",
+            "participant_team_members_affiliations",
             "created_at",
             "method_name",
             "participant_team_members",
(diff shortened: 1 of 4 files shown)


apps/jobs/serializers.py
```diff
@@ -163,6 +165,27 @@ def get_participant_team_members(self, obj):
...
+            return "Participant team does not exist"
+
+        participant_ids = Participant.objects.filter(
+            team=participant_team
+        ).values_list("user_id", flat=True)
+        users = User.objects.filter(id__in=participant_ids)
+        participant_team_members_affiliation = list()
+        for user in users:
+            if user.profile.affiliation == "":
+                affiliation = "None"
```
galipremsagar: Instead of having a special case handling we can straight away append empty strings to `participant_team_members_affiliation`. It will result in a valid csv like this:
```csv
id,Team Name,Team Members Affiliation,Challenge Phase
2,sdf,,Linda Phase
1,West Elizabeth Participant Team,aff name,Linda Phase
```

Which should be good. or Is there any specific reason other than this to Inject "None" string inplace of when affiliation field's are empty. 
If we don't have to do the above special handling, You can simplify the code too to 
```python
return [user.profile.affiliation for user in users]
```

Sanji515: @galipremsagar Actually I did this way cause if there are more than one participants in a team then it will be easy to know which participant's affiliation is `None` and whose affiliation is provided.

galipremsagar: You mean for filtering in google docs & excel? If so then lets get rid of Explicit `"None"` string insertion. We can filter empty fields in both of them like this:
Screenshot 2019-08-17 at 4 32 50 PM

Sanji515: okay, got your point @galipremsagar , thanks :+1: Made the changes, please review

tests/unit/challenges/test_views.py
```diff
@@ -3085,6 +3085,7 @@ def test_get_all_submissions_when_user_is_host_of_challenge(self):
                     "submission_result_file": None,
                     "submission_metadata_file": None,
                     "participant_team_members_email_ids": ["user6@test.com"],
+                    "participant_team_members_affiliations": ["None"],
```
galipremsagar: Test will fail here

Sanji515: Thanks :smile_cat: Fixed! @galipremsagar

PR #2433, opened 2019-08-23
Add feature to seed the database with Submission and Leaderboard entries.

KhalidRmb:
On starting up the containers, the seed script will now seed the database with 
- Submission       
- Leaderboard data.


scripts/seed.py
```diff
@@ -205,6 +209,18 @@ def create_leaderboard():
     return leaderboard
 
 
+def create_leaderboard_data(challenge_phase_split, submission):
+    result = {'score': 0}
```
RishabhJain2018: Please generate a random number here.

Issue #2443, opened 2019-09-18
Add and update existing tests for challenge submission API


Sanji515:
Reference: #2409 

- Some tests were removed in the above mentioned PR but we need those tests with some updation when we upload a file for challenge submission.

- Also need to add tests when we make submission using URL.

coldk91ght:
@Sanji515 Can I take up this issue?

Sanji515:
Sure @coldk91ght

coldk91ght:
@Sanji515 Ohkay. On it. 👍

coldk91ght:
@Sanji515 Shall I make a separate class for submission post tests?  Also, can you please lead me to the input format for url uploads?

r4zorgeek:
@Sanji515 Can I take this issue?

coldk91ght:
I'm sorry for the delay. A close relative of mine had passed away and so I couldn't focus on the task. I have put some new tests. I need some help with the successful url tests and to make a sample input file url for testing. These are the updated tests for now: https://github.com/coldk91ght/EvalAI/commit/38480d99cedd377194fcf8d83e89add0b77750ff

pushkalkatara:
Hi @nikochiko These tests might help you out -  [test_submission_worker.py](https://github.com/Cloud-CV/EvalAI/blob/ca0351dda61be800b2ee50a2b9d4f41b913e8ea9/tests/integration/worker/test_submission_worker.py#L231)

RishabhJain2018:
Thanks, @pushkalkatara for the pointer.

RishabhJain2018:
@nikochiko Please open up the PR for it.

nikochiko:
@pushkalkatara Thanks for the help.
Opening up PR now. Will add changes to it.

pushkalkatara:
Hi @nikochiko . I think these two tests are also required which were removed - [status, file not submitted](https://github.com/Cloud-CV/EvalAI/pull/2409/files#diff-c6268b43fc26de9e08ed355975be069bL424) and [wrong form encoding](https://github.com/Cloud-CV/EvalAI/pull/2409/files#diff-c6268b43fc26de9e08ed355975be069bL445). @Sanji515 Can you comment on this?

Sanji515:
[...]

PR #2470, opened 2019-11-05
Updated tests. New tests for uploaded files

nikochiko:
New tests for testing banned team, and number of concurrent submissions. Also added new test for invalid file url submission.

As per https://github.com/Cloud-CV/EvalAI/issues/2443

### DIFF
--- a/apps/jobs/utils.py
+++ b/apps/jobs/utils.py
@@ -140,7 +140,7 @@ def get_remaining_submission_for_a_phase(
 
 def is_url_possible(url):
     """ Uses urllib.urlparse to determine whether a string qualifies as a valid URL string """
-    o = urlparse(string)
+    o = urlparse(url)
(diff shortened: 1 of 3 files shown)


nikochiko:
Had to try out lot of edits, but finally got the tests right. Used `mock` for testing submission with URL. - The [`is_url_valid` function](https://github.com/Cloud-CV/EvalAI/blob/ca0351dda61be800b2ee50a2b9d4f41b913e8ea9/apps/jobs/utils.py#L140) from tasks.utils however, only checks that the URL does not return a `HttpError`. It raises `urllib` errors when the URL string is not an actual URL. Do we need to edits `views.py` and add tests for that here?

nikochiko:
Moved to https://github.com/Cloud-CV/EvalAI/pull/2489

PR #2472, opened 2019-11-10
Fix #2471 update installation

### LINKED ISSUE #2471: update installation instruction in readme
When I was setting up EvalAI on my system then for Building and running the Docker containers I had to give the command with admin priviledges(in both windows & ubuntu). 
So ``` sudo  docker-compose up --build``` for ubuntu and running cmd with **Run as Administration** should be there.

Thanks!

Saloni-Tayal:
As I tried running the project on my system(windows & linux). I found that installation worked when I used the command with admin priviledge. So change ```docker-compose up --build``` to  ```sudo docker-compose up --build```  .

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,10 +56,10 @@ The steps are:
     git clone https://github.com/Cloud-CV/EvalAI.git evalai && cd evalai
     ```
 
-3. Build and run the Docker containers. This might take a while.
+3. Build and run the Docker containers with admin privileges. This might take a while.
 
     ```
-    docker-compose up --build
+     docker-compose up --build
     ```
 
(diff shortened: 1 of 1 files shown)


Saloni-Tayal:
@RishabhJain2018  Can you please review this?

nikochiko:
@Saloni-Tayal , I think it would be better to break that instruction into two parts, one for Windows and one for Linux, as you mentioned in your issue:#2471.
Also, you can reference your issue in the PR-description, so that both are linked.

Saloni-Tayal:
@nikochiko  Thanks for the advise. 
Issue #2471 is fixed here #2477

PR #2477, opened 2019-11-22
Update README.md

dubesar:
Fixes the issue for running the command to run EvalAI on the Ubuntu and Windows System.

This fixes the issue - #2472

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,8 +57,15 @@ The steps are:
     ```
 
 3. Build and run the Docker containers. This might take a while.
+    
+    ***For Ubuntu :***
 
+    ```shell 
+    sudo docker-compose up --build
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -57,8 +57,15 @@ The steps are:
     ```
 
 3. Build and run the Docker containers. This might take a while.
+
```
RishabhJain2018: Hi @dubesar ,
I don't think we need these changes as you can  add docker user in root user group. For more details please read here: https://docs.docker.com/install/linux/linux-postinstall/

RishabhJain2018:
As per above comment, I'm closing this, please feel free to reopen it in case something is still not resolved.

PR #2483, opened 2019-11-24
Fix issue of not showing leader-board entries after seeding the database

Sanji515:

- Update challenge phase seeding by setting the field `is_submission_public` to `True` so that submission can be go into the `public` state.
- Removed `json.dumps` from the result as it add the line break `\n` in the `result` field of `LeaderboardData` table.
- Removed `is_flagged=True` from Submission object creation.


RishabhJain2018:
Thanks @Sanji515 for catching and fixing it 🎉

PR #2497, opened 2019-12-03
[GCI] Added test for load_challenge_and_return_max_submissions

nikochiko:
* Added test for submission worker [`load_challenge_and_return_max_submissions`](https://github.com/Cloud-CV/EvalAI/blob/master/scripts/workers/submission_worker.py#L648)
* According to GCI task: 'Add test for load_challenge_and_return_max_submissions function for submission worker'

### DIFF
--- a/tests/unit/worker/test_submission_worker.py
+++ b/tests/unit/worker/test_submission_worker.py
@@ -12,7 +12,7 @@ from django.contrib.auth.models import User
 from django.core.files.uploadedfile import SimpleUploadedFile
 from django.utils import timezone
 
-from rest_framework.test import APIClient, APITestCase
+from rest_framework.test import APITestCase
 
 from challenges.models import Challenge
 from hosts.models import ChallengeHostTeam
@@ -27,7 +27,6 @@ from scripts.workers.submission_worker import (
 
 class BaseAPITestClass(APITestCase):
     def setUp(self):
-        self.client = APIClient(enforce_csrf_checks=True)
         self.BASE_TEMP_DIR = tempfile.mkdtemp()
         self.temp_directory = join(self.BASE_TEMP_DIR, "temp_dir")
         self.url = "/test/url"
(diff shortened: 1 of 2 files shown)


nikochiko:
@Ram81 , fixed the Travis build failure. Please review.

tests/unit/worker/test_submission_worker.py
```diff
@@ -30,6 +39,32 @@ def setUp(self):
...
+            description="Description for test challenge",
+            terms_and_conditions="Terms and conditions for test challenge",
+            submission_guidelines="Submission guidelines for test challenge",
+            creator=self.challenge_host_team,
+            start_date=timezone.now() - timedelta(days=2),
+            end_date=timezone.now() + timedelta(days=1),
+            published=False,
+            enable_forum=True,
+            anonymous_leaderboard=False,
+            max_concurrent_submission_evaluation=200000,
```
RishabhJain2018: Please reduce this number to something reasonable.

tests/unit/worker/test_submission_worker.py
```diff
@@ -45,6 +80,16 @@ def test_return_file_url_per_environment(self):
...
 
+    @mock.patch("scripts.workers.submission_worker.load_challenge")
+    @mock.patch("scripts.workers.submission_worker.Challenge.objects.get")
+    def test_load_challenge_and_return_max_submissions(self, mocked_get_challenge, mocked_load_challenge):
+        mocked_get_challenge.return_value = self.challenge
+        q_params = {"pk": 1}
+        response = load_challenge_and_return_max_submissions(q_params)
+        mocked_get_challenge.assert_called_with(pk=1)
+        mocked_load_challenge.assert_called_with(self.challenge)
+        self.assertEqual(response, (200000, self.challenge))
```
RishabhJain2018: Can you please use `self.challenge. max_concurrent_submission_evaluation` instead of hardcoding the number.

tests/unit/worker/test_submission_worker.py
```diff
@@ -45,6 +80,16 @@ def test_return_file_url_per_environment(self):
         returned_url = return_file_url_per_environment(self.url)
         self.assertEqual(returned_url, "http://testserver/test/url")
 
+    @mock.patch("scripts.workers.submission_worker.load_challenge")
+    @mock.patch("scripts.workers.submission_worker.Challenge.objects.get")
+    def test_load_challenge_and_return_max_submissions(self, mocked_get_challenge, mocked_load_challenge):
+        mocked_get_challenge.return_value = self.challenge
+        q_params = {"pk": 1}
+        response = load_challenge_and_return_max_submissions(q_params)
+        mocked_get_challenge.assert_called_with(pk=1)
```
RishabhJain2018: Please change 1 with `self.challenge.pk`

tests/unit/worker/test_submission_worker.py
nikochiko: The default is 100000, so I thought to keep it 200000. Will 100 do?

RishabhJain2018: I think 100 should be fine.

nikochiko: Done. Please see changes.

nikochiko:
@RishabhJain2018 , please look at the changes. Trying to connect to the database caused an error, so the `pk` field had to be put in manually. Does this look good?

nikochiko:
@RishabhJain2018 , Can we make the test cases as  `rest_framework.test.APITestCase` rather than `unittest.TestCase` to make access to the database possible? This is the case with [integration tests](https://github.com/Cloud-CV/EvalAI/blob/master/tests/integration/worker/test_submission_worker.py#L17). That way, we can avoid having to set the `pk` field manually and make tests more reliable.

tests/unit/worker/test_submission_worker.py
```diff
@@ -45,6 +81,22 @@ def test_return_file_url_per_environment(self):
...
+        self.assertEqual(response, (self.challenge.max_concurrent_submission_evaluation, self.challenge))
+
+    def test_load_challenge_and_return_max_submissions_when_challenge_does_not_exist(self):
+        non_existing_challenge_pk = self.challenge.pk + 1
+        with mock.patch("scripts.workers.submission_worker.logger.exception") as mock_logger:
+            try:
+                load_challenge_and_return_max_submissions({"pk": non_existing_challenge_pk})
+            except Challenge.DoesNotExist:
+                pass
+            mock_logger.assert_called_with("Challenge with pk {} doesn't exist".format(non_existing_challenge_pk))
```
RishabhJain2018: Should this condition be in the except part?

nikochiko: But then the test would be successful if the `Challenge.DoesNotExist` is not raised. So I placed it outside the `try...except` block.

nikochiko: Will make some updates to add a `with self.assertRaises` instead.

PR #2501, opened 2019-12-09
Add tests for download_and_extract_file

takitsuse:
## The corresponding GCI task:
https://codein.withgoogle.com/dashboard/task-instances/6134019935174656/

## Brief explanation of my changes: 
2 tests for [`download_and_extract_file`](https://github.com/Cloud-CV/EvalAI/blob/master/scripts/workers/remote_submission_worker.py#L99)
- a test that confirms when no exception raised, the file on `download_location` is correctly written.
- a test that confirms when an exception is raised, an error message is written by a logger.


tests/integration/worker/test_remote_submission_worker.py
```diff
@@ -0,0 +1,54 @@
+import responses
```
Ram81: @takitsuse can you reorder the imports alphabetically

tests/integration/worker/test_remote_submission_worker.py
```diff
@@ -0,0 +1,54 @@
...
+        self.download_filename = "file"
+        self.download_path = "{}{}".format(self.download_dir, self.download_filename)
+
+        try:
+            os.makedirs(self.download_dir)
+        except OSError:
+            pass
+
+
+class DownloadAndExtractFileWithProperURL(BaseTestClass):
```
Ram81: @takitsuse can you just create one class named `TestDownloadAndExtractFile` and two separate methods to test both positive and negative cases.

yashdusing:
I guess there exists 2 instances for the function. One in remote worker and the other in the normal submission worker. I think you have to write test cases for both

takitsuse:
I think I might be asked to add tests only for the function in remote submission worker specifically, which is said in the task description.

tests/integration/worker/test_remote_submission_worker.py
```diff
@@ -0,0 +1,44 @@
...
+            responses.GET,
+            self.url,
+            body=self.body,
+            status=200
+        )
+
+        try:
+            os.makedirs(self.download_dir)
+        except OSError:
+            pass
```
Ram81: @takitsuse can you add a log statement here, in case directory creation fails we won't know what exactly the issue is

tests/integration/worker/test_remote_submission_worker.py
```diff
@@ -0,0 +1,44 @@
...
+
+    @responses.activate
+    def test_download_and_extract_file_when_proper_url_is_given(self):
+        remote_submission_worker.download_and_extract_file(self.url, self.download_path)
+
+        with open(self.download_path, "r") as f:
+            assert f.read() == self.body
+
+    @mock.patch("scripts.workers.remote_submission_worker.logger.error")
+    def test_download_and_extract_file_when_improper_url_is_given(self, mock_logger):
```
Ram81: Rename this function to `test_download_and_extract_file_with_incorrect_url`

tests/integration/worker/test_remote_submission_worker.py
```diff
@@ -0,0 +1,44 @@
...
+            status=200
+        )
+
+        try:
+            os.makedirs(self.download_dir)
+        except OSError:
+            pass
+
+    @responses.activate
+    def test_download_and_extract_file_when_proper_url_is_given(self):
```
Ram81: Rename this to `test_download_and_extract_file_with_correct_url`

tests/integration/worker/test_remote_submission_worker.py
takitsuse: Sure! I am planning to mock the logger and exec `logger.error()` in except statement and exec `mock_logger.assert_called()` in the test function.
Does this sounds okay to you?

takitsuse: Hmm...I looked around other testing codes in EvalAI and I found no examples of logging statement is being used for creating directory even though `os.makedirs()` I found ocassionally.
Do I really have to add logging statement for creating directory?

Ram81: @takitsuse instead of writing your own `mkdir` can you use this https://github.com/Cloud-CV/EvalAI/blob/master/scripts/workers/remote_submission_worker.py#L146. In case this fails you can your mock_logger to log the error

scripts/workers/submission_worker.py
```diff
@@ -167,7 +167,14 @@ def create_dir(directory):
         Creates a directory if it does not exists
     """
     if not os.path.exists(directory):
-        os.makedirs(directory)
+        try:
```
Ram81: Can we revert this, as this is not part of the task. You can submit a separate Pr for this

takitsuse: Of course! After I revert this, I will be using `assert os.path.exists()` to detect the failure of file creation, since I think we can't detect it with mock_logger without being logged.

tests/unit/worker/test_remote_submission_worker.py
```diff
@@ -0,0 +1,47 @@
...
+        )
+
+        remote_submission_worker.create_dir(self.download_dir)
+
+    def tearDown(self):
+        shutil.rmtree(self.download_dir)
+
+    @responses.activate
+    def test_download_and_extract_file_with_correct_url(self):
+        self.assertTrue(os.path.exists(self.download_dir))
```
takitsuse: Here the testing code checks if the directory is created successfully, and if not, the test fails with the message indicating that the directory creation went wrong.

PR #2502, opened 2019-12-10
[GCI] Add test for extract_challenge_data

nikochiko:
* As per the GCI task, added tests for [`submission_worker.extract_challenge_data`](https://github.com/Cloud-CV/EvalAI/blob/master/scripts/workers/submission_worker.py#L197)

CL:
* Added required class variables, formatted them properly.
* Added helper method `get_url_for_test_environment`
* Refactored code into classes
* Added following tests:
  * To see that the challenge_module is being correct and imported correctly, and other components of the function are being executed without any issues.
  * To see that the `evaluation_script` for the challenge has correctly been added to `EVALUATION_SCRIPTS` dict.
  * To see that the name of the annotation file has correctly been added to `PHASE_ANNOTATION_FILE_NAME_MAP` dict.

The test for the case when the method should throw an error is already covered in the integration tests here: https://github.com/Cloud-CV/EvalAI/blob/master/tests/integration/worker/test_submission_worker.py#L239


nikochiko:
Flake8 shows it is a `SyntaxError`, but I think my syntax should be valid according to documentation: https://docs.python.org/3/reference/compound_stmts.html#the-with-statement

nikochiko:
Fixed the `with` block syntax errors.
The other error was being caused due to a bug in `mock.patch.dict` where the code in the example of the documentation fails: 3rd example in https://docs.python.org/3/library/unittest.mock.html#patch-dict --> https://colab.research.google.com/gist/nikochiko/6870c696c859c6cde00e1bfef0665ad8/test-mock-documentation.ipynb
Will post an issue there.

tests/unit/worker/test_submission_worker.py
```diff
@@ -27,27 +31,50 @@
...
+        )
+        self.CHALLENGE_IMPORT_STRING = "challenge_data.challenge_{challenge_id}"
+
         self.temp_directory = join(self.BASE_TEMP_DIR, "temp_dir")
+
+        self.testserver = "http://testserver"
         self.url = "/test/url"
-        self.input_file = open(join(self.BASE_TEMP_DIR, 'dummy_input.txt'), "w+")
+
+        self.input_file = open(
```
Ram81: @nikochiko can we revert this to original code, making it multiline makes it hard to read.

nikochiko: Yep. :+1: Added the change

nikochiko:
@Ram81 The code for this test was becoming too long and unorganized. As per the docstring of `submission_worker.extract_challenge_data`, the method should add corresponding data to `EVALUATION_SCRIPTS` and `PHASE_ANNOTATION_FILE_NAME_MAP`. I separated the test into three components:
1. For testing the flow, and if arguments are being correctly passed. 
2. For testing if correct data is being added to `EVALUATION_SCRIPTS`
3. For testing if correct data is being added to `PHASE_ANNOTATION_FILE_NAME_MAP`.
If any error occurs, these tests will help us to pinpoint the problem easily.

There were lot of variables being used commonly in these three tests. I separated the specifics of these tests into one class and similarly refactored for other tests. 
Please comment on whether this refactoring was a good idea here.

tests/unit/worker/test_submission_worker.py
```diff
@@ -79,8 +139,84 @@ def test_create_dir_as_python_package(self):
...
+            self.challenge_phase.test_annotation.url
+        )
+        self.challenge_data_directory = self.CHALLENGE_DATA_DIR.format(challenge_id=self.challenge.id)
+        self.challenge_import_string = self.CHALLENGE_IMPORT_STRING.format(challenge_id=self.challenge.id)
+        self.challenge_zip_file = join(self.challenge_data_directory, "challenge_{}.zip".format(self.challenge.id))
+        self.evaluation_script_url = self.get_url_for_test_environment(self.challenge.evaluation_script.url)
+
+    @mock.patch("scripts.workers.submission_worker.download_and_extract_file")
+    @mock.patch("scripts.workers.submission_worker.download_and_extract_zip_file")
+    def test_extract_challenge_data(self,
```
Ram81: @nikochiko can you also add tests for cases when exception is thrown

nikochiko: @Ram81 That test case is already covered in the integration tests: https://github.com/Cloud-CV/EvalAI/blob/master/tests/integration/worker/test_submission_worker.py#L239
Do we need it here as well?

nikochiko: Anyways, having all tests in one place makes more sense. Also coveralls says this is not covered: https://coveralls.io/builds/27483663/source?filename=scripts/workers/submission_worker.py#L257

nikochiko:
@Ram81 Added the test.

RishabhJain2018:
@nikochiko Can you please resolve the conflicts here?

nikochiko:
@RishabhJain2018 Done! 😄

PR #2508, opened 2019-12-13
Tests for download_and_extract_zip_file

grave-panda:
Added tests for download_and_extract_zip_file function for submission worker

### DIFF
--- a/tests/unit/worker/test_submission_worker.py
+++ b/tests/unit/worker/test_submission_worker.py
@@ -1,3 +1,5 @@
+import hashlib
+
 import boto3
 import mock
 import os
@@ -19,10 +21,12 @@ from hosts.models import ChallengeHostTeam
 from scripts.workers.submission_worker import (
     create_dir,
     create_dir_as_python_package,
+    download_and_extract_zip_file,
     load_challenge_and_return_max_submissions,
     return_file_url_per_environment,
     get_or_create_sqs_queue
 )
+from unit.hosts.test_utils import BaseTestClass
 
 
 class BaseAPITestClass(APITestCase):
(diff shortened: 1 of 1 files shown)


tests/unit/worker/test_submission_worker.py
```diff
@@ -109,3 +113,43 @@ def test_get_or_create_sqs_queue_for_non_existing_queue(self):
...
+        create_dir(self.base_dir)
+
+    def tearDown(self):
+        shutil.rmtree(self.base_dir)
+
+    def test_download_and_extract_file_with_correct_url(self):
+        assert os.path.exists(self.base_dir)
+
+        download_and_extract_zip_file(self.url, self.download_path, self.extract_path)
+        # with open(self.download_path, "r") as f:
```
Ram81: @grave-panda can you remove the unnecessary code

tests/unit/worker/test_submission_worker.py
```diff
@@ -109,3 +113,43 @@ def test_get_or_create_sqs_queue_for_non_existing_queue(self):
...
+        #     assert f.read() == self.body
+        assert os.path.exists("{}/zip_10MB".format(self.extract_path))
```
Ram81: @grave-panda can we rename the file `zip_10MB` to some appropriate name

tests/unit/worker/test_submission_worker.py
```diff
@@ -109,3 +113,43 @@ def test_get_or_create_sqs_queue_for_non_existing_queue(self):
...
+
+        with open("{}/zip_10MB/file_example_ODS_5000.ods".format(self.extract_path)) as file_to_check:
```
Ram81: same here

tests/unit/worker/test_submission_worker.py
```diff
@@ -109,3 +113,43 @@ def test_get_or_create_sqs_queue_for_non_existing_queue(self):
...
+class TestDownloadAndExtractFile(APITestCase):
+    def setUp(self):
+        super(TestDownloadAndExtractFile, self).setUp()
+
+        self.base_dir = "{}/JdyW0ZCFLr".format(tempfile.gettempdir())
+        self.download_filename = "VwYOaVWkdw"
+        self.download_path = "{}/{}".format(self.base_dir, self.download_filename)
+        self.extract_dirname = "gMTblaJFbM"
+        self.extract_path = "{}/{}".format(self.base_dir, self.extract_dirname)
+        self.correct_md5 = "86629294bdc8256500ca37a9afecbd79"
```
Ram81: @grave-panda can we change this to be computed from the file being used in the test. Hardcoding it will be a problem if we change the zip file later on

pranavgade20: Well I just download a random zip file from the internet(I googled sample zip file download and went with the first link: https://file-examples.com/index.php/text-files-and-archives-download/). So, unless you know about a CDN, or can put a sample zip file somewhere in this project, I cannot change this. Also, if I understand this correctly, testing is checking the code against arbitrary inputs, so this is the expected output. In this case, if I do not hardcore the hash, some bug in future versions of hashlib might cause this test to behave incorrectly.

tests/unit/worker/test_submission_worker.py
pranavgade20: Yeah sorry about that, I forgot to cleanup

tests/unit/worker/test_submission_worker.py
pranavgade20: Unfortunately, no. Since this is not hosted by me, I don't think this can be changed.

tests/unit/worker/test_submission_worker.py
grave-panda: Well I just download a random zip file from the internet(I googled sample zip file download and went with the first link: https://file-examples.com/index.php/text-files-and-archives-download/). So, unless you know about a CDN, or can put a sample zip file somewhere in this project, I cannot change this. Also, if I understand this correctly, testing is checking the code against arbitrary inputs, so this is the expected output. In this case, if I do not hardcore the hash, some bug in future versions of hashlib might cause this test to behave incorrectly.

tests/unit/worker/test_submission_worker.py
grave-panda: Since I do not host this sample zip, I cannot change the name, unless you can suggest any other sample file that is hosted online.

tests/unit/worker/test_submission_worker.py
grave-panda: Sorry about that, I forgot to clean up

pushkalkatara:
Hi @grave-panda The build is failing. Please resolve the build.

PR #2509, opened 2019-12-14
Add test for create_dir_as_python_package

takitsuse:
## Clarification:
- Add a test to see `test_create_dir_as_python_package(directory)` creates new directory in the path `directory` and new empty file `__init__.py` in the path.

### DIFF
--- a/tests/unit/remoteworker/test_remote_worker.py
+++ b/tests/unit/remoteworker/test_remote_worker.py
@@ -1,8 +1,14 @@
 import mock
+import os
+import shutil
+import tempfile
+
+from os.path import join
 
 from unittest import TestCase
 
 from scripts.workers.remote_submission_worker import (
+    create_dir_as_python_package,
     make_request,
     get_message_from_sqs_queue,
     delete_message_from_sqs_queue,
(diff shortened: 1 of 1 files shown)


tests/unit/remoteworker/test_remote_worker.py
```diff
@@ -141,3 +147,20 @@ def test_return_url_per_environment(self):
...
+        self.temp_directory = join(self.BASE_TEMP_DIR, "temp_dir")
+
+    def test_create_dir_as_python_package(self):
+        create_dir_as_python_package(self.temp_directory)
+        self.assertTrue(os.path.isfile(join(self.temp_directory, "__init__.py")))
+
+        with open(join(self.temp_directory, "__init__.py")) as f:
+            self.assertEqual(f.read(), "")
+
+        shutil.rmtree(self.temp_directory)
```
Ram81: @takitsuse can you add a check to test whether folder is deleted or not.

takitsuse: File deletion is not what `create_dir_as_python_package` does, so I would say we don't have to test it...

Ram81: Yeah I understood that, but next time when this test runs and remove dir fails second run can give a false positive as the folder already exists

takitsuse: Now I see what you meant. will change that.

PR #2528, opened 2019-12-21
Remove verbosity in challengeCtrl.js and fix a typo

takitsuse:
This PR removes verbosity when defining `vm.phaseStartDate` and `vm.phaseEndDate`.


Sanji515:
Can we please correct this type error here(https://github.com/Cloud-CV/EvalAI/blob/master/apps/hosts/models.py#L13) too in this PR? :smile:

PR #2540, opened 2019-12-25
Add functionality to update the maximum submissions per month

takitsuse:
- [The corresponding GCI task](https://codein.withgoogle.com/dashboard/task-instances/5787553504952320/)
- Changes
  - Add functionality to update the maximum submissions per month from UI


yashdusing:
@takitsuse these are just the frontend changes. You need to make backend API changes for the functionality to be added as well

yashdusing:
also the update functionality should operate from the admin panel i.e. the admin should be able to update the max submissions per month field

takitsuse:
@yashdusing Yes, I am aware of it. However, we are already able to update the value from admin panel and from frontend UI, so I think we don't have to add new API.

takitsuse:
And why we can already update the value without adding an API is that there is already an API for this. See [here](https://github.com/Cloud-CV/EvalAI/blob/master/apps/challenges/views.py#L515-L583) and [here](https://github.com/Cloud-CV/EvalAI/blob/master/apps/challenges/serializers.py#L63-L92).

takitsuse:
@Ram81 @Sanji515 @RishabhJain2018 
Could you also take a look at this? Thanks.

RishabhJain2018:
@Sanji515 Can you please test this once on your local system?

PR #2541, opened 2019-12-25
Code Coverage

Suryansh5545:
Added the Code coverage file which contains the detail about file/code which is not tested and is decreasing overall code coverage.
If test are created the coverage will go form 72% to 90% or above


Suryansh5545:
@RishabhJain2018 The Pr has served its purpose and no longer need to be open

Issue #1705, opened 2018-06-16
Add support for editing the challenge start and end date


RishabhJain2018:
**Current Scenario:**
Currently, a challenge host can't edit the challenge start date and end date. This can only be done by EvalAI admin.

**Deliverables:**
Add functionality to edit the challenge start date and end date on frontend.
Add a check to make sure that challenge end date is always after challenge start date.

Use this(https://indrimuska.github.io/angular-moment-picker/) package for adding date-time support.

pujanm:
I am interested in solving this issue how can I go about the same?

pujanm:
Can you help me out.
I am really interested to contribute to CloudCV.

RishabhJain2018:
Hi, @pujanm I am sorry that I missed this message. Thanks for your interest in CloudCV.

For the above issue, you can check how we have implemented the challenge details edit section. We will be following the same approach as that. Please let me know if you have further doubts. I'll be happy to help.

pujanm:
I was trying to edit the challenge overview it is giving this `/api/challenges/challenge_host_team/undefined/challenge/3` there is an undefined in place of `challengeHostId`. Is there any reason or I might have done some mistake while making this challenge?

RishabhJain2018:
I think @pujanm is still working on this issue.
>/api/challenges/challenge_host_team/undefined/challenge/3 there is an undefined in place of challengeHostId

I am not sure why it is showing this behaviour to you. Can you please try to debug it?

pujanm:
Yes.

Ayukha:
@RishabhJain2018 I would like to work on it.

RishabhJain2018:
Sure!

RishabhJain2018:
Since @Ayukha isn't active these days, would you like to work on this @nikhilmudholkar?

Ayukha:
@RishabhJain2018  I am working on it. And I'll open a PR soon.

nikhilmudholkar:
Yaa i would like to work on it but it seems that someone is already working on this issue

RishabhJain2018:
@Ayukha Can you please let us know ETA on this? This is now becoming a priority thing for us

Sanji515:
@RishabhJain2018 If @Ayukha isn't active these days, then I can work on this issue.

Ayukha:
@RishabhJain2018  I'll try to fix it by tonight. Actually, I'm struggling in using the package suggested.

PR #2552, opened 2019-12-28
Restrict challenge start and end date edit option to challenge host only

nsjcorps:
Minor fixes to #1705 .

I have added the functionality to restrict the date edit option so it will only show for the challenge host.

I used an `ng-if` conditional to ensure that this change works.


nsjcorps:
@Sanji515 @RishabhJain2018 @Ram81 Kindly Review

PR #2556, opened 2019-12-29
[GCI] : Why code coverage is at 73% while expected is 90%<

amanex007:
Making a codecoverage.md file to describe the reasons because of which the code coverage is lower than expected. Code extracts have been explained in brief as required. All files with with less than 90% coverage have been mentioned here. Some files with less than 40% coverage have not been discussed as a need is felt to test the full code they contain.


KhalidRmb:
You can do better. Here's an example for what I meant when I said, mention the **test cases**:
Also, there's no point to put the actual code in the file itself since anyone can simply check the coverage on codecov. Just add a link to the uncovered line wherever relevant.

_**apps/jobs/sender.py (76%)**_ 
**To-do:**

1. Test case for passing an empty queue_name in function 'get_or_create_sqs_queue'
2. Test case for passing a queue_name that would cause an exception other than 'NonExistentQueue' in function 'get_or_create_sqs_queue'
3.  Test case for passing an invalid challenge_id in function 'publish_submission_message'

You essentially have to minimize the time any person spends to recognize what tests are missing. And make it readable, and easy to navigate.

amanex007:
I have made changes as expected, please take a look.

Codecoverage.md
```diff
@@ -0,0 +1,3087 @@
...
+
+<details>
+    <summary>View Code</summary> 
+    <br>
+    Turn data into a hash and an encoded string, suitable for use with decode_data.
+    The inverse of encode_data.
+    Function to send email.
+    
+```
+def encode_data(data):
```
pushkalkatara: We can remove the code as links to code are added with the missing tests.

pushkalkatara:
@amanex007 The code still exists in the coverage.md, please make changes as suggested by @KhalidRmb

amanex007:
Please take a look @pushkalkatara

pushkalkatara:
LGTM :+1: What do you think @KhalidRmb

KhalidRmb:
Hey @amanex007 From what I see you've just described the function. How does that offer any value? Any person can do that by just seeing the code. \

But, listing out which test cases are missing is better. Refer the example I gave. **Please read about unit testing and try to understand what a test case means**. Every combination of input will give a different result and run different parts of the code. And your task is to identify such test cases.

amanex007:
I understand that @KhalidRmb , i will try doing what you asked for, but can you offer me sometime and maybe a bit help. This is my first time at analyzing codes or github, so its really hard for me without proper help.

amanex007:
Will this tutorial help me ?
https://www.toptal.com/qa/how-to-write-testable-code-and-why-it-matters

KhalidRmb:
Have a look at these as well: https://www.guru99.com/unit-testing-guide.html
http://softwaretestingfundamentals.com/unit-testing/

amanex007:
Could any mentor can please review the changes that i have made for just the first file, so that i can proceed like that onto the next too

KhalidRmb:
Hey @amanex007 Sorry for the delay...I was expecting something like this:
For ex, let's take the get_or_create_sqs_queue_object method:
1. Testcase when the settings.DEBUG & TEST variables are both False.
2. Testcase when the queue with name "queue_name" doesn't exist. 
3. Testcase when the ENV variables for getting the sqs object are invalid. (Identifying this testcase is crucial, since there's no code to handle this within the function. And it would signal any other person to go and fix that.)

Since, for each of my points, a different part of the method is executed right? And that's our goal. And maybe along the way identify cases which the tests are missing entirely as well!

amanex007:
Hi Khalid, i really want to complete this task, but i feel that maybe i am not eligible enough to meet up to the expectations, i have already used up my 8 days, i believe i should move on to a task that i can do.

amanex007:
What do you suggest?

PR #2563, opened 2019-12-31
Fix #bugnum:2553 Main & Sim Header Fixed For Scaling in all type of devices

### LINKED ISSUE #2553: Header Buttons Not behaving as expected on Mobile view
- the expected behaviour:-

The 3 Buttons Challenge,Docs & Forum should have either collapse with header or align in more proper way so user dont have problem viewing and the log out button should stick to the header 
- the observed behaviour:-

All 4 Buttons mentioned above follow the view if user scroll down it would have been correct if aligned properly . But they will interrupt when user might try to click something else 
- steps to reproduce the problem:-

Login and go to any page like dashboard in smartphone 
- which operating system and/or browser you're using, if relevant
Android 8.0, Chrome

Screenshot_20191229-051254_Chrome

Suryansh5545:
This Fix #2553 
Sir I have made those 3 buttons(all challenge, forum, docs) disappear after 665px. i am using 665px because if they don't disappear after that then log out button will not behave as expected. and when they disappear. two new options for forum and docs appear in the  side nav bar. also i found that the main header disappears just after 1000px or around it which is quiet weird . so i make it disappear like sim header .

### **Without Login(main-header):
At 348px**
Selection_009

------------------------------------------------------------------------------------------------------------------------------------

At 620px
Selection_010

------------------------------------------------------------------------------------------------------------------------------------

At 666px
Selection_011
**All Elements will appear before 666px, previously all elements were disappearing after or around a 1000px .**

### **After Login(Sim-header):
At 348px**
Selection_012
--------------------------------------------------------------------------------------------------------------------------------------
At 620px
Selection_013
--------------------------------------------------------------------------------------------------------------------------------------
At 666px
Selection_014
**All Elements will appear before 666px.**

The log out Button after 250px doesn't behave properly. but i didn't considered it as most small display device after bigger screen than 250px


Suryansh5545:
@Sanji515 Please take a look

Suryansh5545:
@RishabhJain2018 Sir, Please take a look at this

Sanji515:
Also on the landing page after login it doesn't look same as you have fixed for before login. @Suryansh5545 Please take a look at this.

Suryansh5545:
@Sanji515 Sir I fixed the font weight and fixed the difference

PR #2569, opened 2020-01-02
Fix coverage for jobs/utils

nikochiko:
As a Google-Code In task

#### Objective:
To increase coverage of `jobs/utils.py` to > 80 %

#### Reason for low coverage:
1. Uncovered code
2. Missing `test_` prefix in names of some tests in `test_views.py`

#### Changes proposed:
1. Add missing tests for `is_url_valid` and `get_file_from_url`
2. Fix the names by adding `test_` prefix


nikochiko:
The tests are failing because the responses library only mocks requests from the `requests` library. Here we use `urllib.request`. However, it is better to use the `requests` library as recommended in the urllib documentation as well (refer: https://docs.python.org/3.6/library/urllib.request.html#module-urllib.request). 
That would also enable the tests to use the `responses` library and make better tests. Shall I make the switch to requests from urllib in the utils.py file? 
(Also mention if it should be done in another PR and maybe throughout the repo!?)
@RishabhJain2018 @vkartik97 @Ram81

nikochiko:
The coverage for jobs/utils is now 100% (see [codecov](https://codecov.io/gh/Cloud-CV/EvalAI/pull/2569/tree?path=apps%2Fjobs) and [coveralls](https://coveralls.io/builds/27877904/source?filename=apps/jobs/utils.py)).

Note: the tests use https://httpstat.us for obtaining response with desired status codes for testing.

vkartik97:
>  In the utils.py file, both requests and urllib are used. It would be better if we only use the requests library as we don't need two libraries for the same purpose.

Can you make an issue for this?

nikochiko:
@vkartik97 Sure! Here it is: https://github.com/Cloud-CV/EvalAI/issues/2570

PR #2078, opened 2019-02-06
Fix #2067: Make sure all users in team are verified before participation.

KhalidRmb:
Fixes #2067 - Made necessary changes to participants/views.py file to make sure that the invitee's email is verified before being added to a team and add an appropriate warning.

PR #2573, opened 2020-01-03
Fix#2067: Disable team to participate if user has unverified email

### LINKED ISSUE #2067: Disable user to participate in a challenge when one of the user is unverified
### Observed Behaviour
When a verified user adds an unverified user to his team and then tries to participate in a challenge, he is successfully able to submit his entry to the challenge.
Please see the [screencast](https://drive.google.com/file/d/0B1bLU2zTKVNmRUhRSkVpSGRheEFQSmVHWTdUMnhaWnNXd2I4/view?usp=sharing) for reference.

### Expected Behaviour
If the user hasn’t verified and is invited to a team, then he/she cannot participate in the challenge. They have to first verify their email address.

@deshraj @RishabhJain2018

nikochiko:
Fix #2067 

Changes proposed:
* Throw a HTTP `406 NOT ACCEPTABLE` error when one or more of team members have unverified emails
* List the members with unverified emails so that they can be informed

PR used for reference: https://github.com/Cloud-CV/EvalAI/pull/2078

Edit: Added unit tests for the mentioned changes

@gautamjajoo @KhalidRmb @Ram81 Please have a look.

### DIFF
--- a/apps/challenges/views.py
+++ b/apps/challenges/views.py
@@ -312,8 +312,8 @@ def add_participant_team_to_challenge(
             continue
 
     if unverified_emails:
-        message = "All team members need to have verified emails in order to participate. \
-        Users with the following email addresses have unverified emails:\n{}".format(
+        message = "Sorry, all team members must be verified through Email before participating. \
+        The following users have unverified email address:\n{}".format(
             "\n".join(unverified_emails)
         )
         response_data = {"error": message}
--- a/apps/challenges/views.py
+++ b/apps/challenges/views.py
@@ -269,19 +269,6 @@ def add_participant_team_to_challenge(
                     response_data, status=status.HTTP_406_NOT_ACCEPTABLE
                 )
 
-    # Check if all participants have verified emails
-    unverified_users = []
-    for user in participant_team_user_ids:
-        if EmailAddress.objects.filter(user=user, verified=False).exists():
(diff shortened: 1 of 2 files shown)


apps/challenges/views.py
```diff
@@ -302,6 +302,23 @@ def add_participant_team_to_challenge(
...
+    unverified_emails = []
+    for user in participant_team_user_ids:
+        try:
+            unverified_user = EmailAddress.objects.get(user=user, verified=False)
+            unverified_emails.append(unverified_user.email)
+        except EmailAddress.DoesNotExist:
+            continue
+
+    if unverified_emails:
+        message = "All team members need to have verified emails in order to participate. \
```
Ram81: @nikochiko rest of the changes LGTM but can we rephrase the message. It seems quite difficult to follow.

nikochiko: @Ram81 Sure! How about, simply 'The following users have unverified emails:'?

nikochiko: Or this is @KhalidRmb 's message : "Sorry, all team members must be verified through Email before participating.", 
then we can change to "Sorry, all team members must be verified through Email before participating. The following users have unverified emails: ..."

nikochiko: Done!

PR #2575, opened 2020-01-03
Fix returning response in worker_utils for RL worker

RishabhJain2018:


scripts/workers/worker_util.py
```diff
@@ -55,7 +55,7 @@ def delete_message_from_sqs_queue(self, receipt_handle):
         url = self.return_url_per_environment(url)
         data = {"receipt_handle": receipt_handle}
         response = self.make_request(url, "POST", data)  # noqa
```
nikochiko: @RishabhJain2018 Sorry for the late response. I think the `# noqa` here can be removed as well.
Shall I create a new PR for it or is it okay if I merge it into the tests PR (https://github.com/Cloud-CV/EvalAI/pull/2576)?

PR #2576, opened 2020-01-03
Fix #2557: Add tests for worker_util.py for RL workers

### LINKED ISSUE #2557: Add tests for new RL Worker files
#### Scenario:
Recently, new files for RL workers have been added:
1. [rl_submission_worker.py](https://github.com/Cloud-CV/EvalAI/blob/master/scripts/workers/rl_submission_worker.py)
2. [worker_util.py](https://github.com/Cloud-CV/EvalAI/blob/master/scripts/workers/worker_util.py)

#### Solution
We will need new tests for these.

> warning note: the concerned part is still under development. Therefore, tests should be added for utility functions first (which are guaranteed to be used), and then those for the newer part.

nikochiko:
Add tests for worker_util.py in RL workers

### DIFF
--- a/scripts/workers/worker_util.py
+++ b/scripts/workers/worker_util.py
@@ -6,7 +6,7 @@ logger = logging.getLogger(__name__)
 
 URLS = {
     "get_message_from_sqs_queue": "/api/jobs/challenge/queues/{}/",
-    "delete_message_from_sqs_queue": "/api/jobs/queues/{}/receipt/{}/",
+    "delete_message_from_sqs_queue": "/api/jobs/queues/{}/",
     "get_submission_by_pk": "/api/jobs/submission/{}",
     "get_challenge_phases_by_challenge_pk": "/api/challenges/{}/phases/",
     "get_challenge_by_queue_name": "/api/challenges/challenge/queues/{}/",
@@ -51,11 +51,10 @@ class EvalAI_Interface:
         return response
 
     def delete_message_from_sqs_queue(self, receipt_handle):
-        url = URLS.get("delete_message_from_sqs_queue").format(
-            self.QUEUE_NAME, receipt_handle
-        )
+        url = URLS.get("delete_message_from_sqs_queue").format(self.QUEUE_NAME)
         url = self.return_url_per_environment(url)
-        response = self.make_request(url, "GET")  # noqa
+        data = {"receipt_handle": receipt_handle}
+        response = self.make_request(url, "POST", data)  # noqa
         return response.status_code
 
     def get_submission_by_pk(self, submission_pk):
--- a/scripts/workers/worker_util.py
+++ b/scripts/workers/worker_util.py
@@ -55,7 +55,7 @@ class EvalAI_Interface:
         url = self.return_url_per_environment(url)
         data = {"receipt_handle": receipt_handle}
         response = self.make_request(url, "POST", data)  # noqa
-        return response.status_code
+        return response
 
     def get_submission_by_pk(self, submission_pk):
         url = URLS.get("get_submission_by_pk").format(submission_pk)
--- a/scripts/workers/worker_util.py
+++ b/scripts/workers/worker_util.py
@@ -54,8 +54,8 @@ class EvalAI_Interface:
         url = URLS.get("delete_message_from_sqs_queue").format(self.QUEUE_NAME)
         url = self.return_url_per_environment(url)
         data = {"receipt_handle": receipt_handle}
-        response = self.make_request(url, "POST", data)
-        return response
+        response = self.make_request(url, "POST", data)  # noqa
+        return response.status_code
 
     def get_submission_by_pk(self, submission_pk):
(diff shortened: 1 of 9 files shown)


tests/unit/rl_worker/test_worker_util.py
```diff
@@ -0,0 +1,174 @@
...
+                      body=requests.exceptions.RequestException(self.example_error_description))
+
+        with self.assertRaises(requests.exceptions.RequestException):
+            interface.make_request(url, "GET")
+
+        mock_logger.assert_called_with("The worker is not able to establish connection with EvalAI")
+
+    def test_get_message_from_sqs_queue(self):
+        url = "{}{}".format(evalai_api_server, self.get_message_from_sqs_queue_url(queue_name))
+        with mock.patch("__main__.EvalAI_Interface.make_request", return_value=self.success_response) as mock_make_request:
```
nikochiko: @vkartik97 @RishabhJain2018 I am currently stuck on this issue since a few days.
The tests for `make_request` are working correctly, so I wanted to mock make_request in other tests, but this always throws an error. I tried using several different ways to patch like `patch("__main__.interface.make_request")` and simply `patch(interface.make_request)` which should work according to docs but the travis throws a `ModuleNotFoundError` with the message `__main__ is not a  package`.  
Is there a workaround for this or should we go with using the `responses` library to write tests?

nikochiko:
@RishabhJain2018 @vkartik97 I have added the tests, PTAL.

scripts/workers/worker_util.py
```diff
@@ -54,8 +54,8 @@ def delete_message_from_sqs_queue(self, receipt_handle):
         url = URLS.get("delete_message_from_sqs_queue").format(self.QUEUE_NAME)
         url = self.return_url_per_environment(url)
         data = {"receipt_handle": receipt_handle}
-        response = self.make_request(url, "POST", data)  # noqa
-        return response.status_code
+        response = self.make_request(url, "POST", data)
+        return response
```
vkartik97: Please remove this change, as it is being handled in https://github.com/Cloud-CV/EvalAI/pull/2575

scripts/workers/worker_util.py
```diff
@@ -85,7 +85,7 @@ def get_challenge_phase_by_pk(self, challenge_pk, challenge_phase_pk):
         response = self.make_request(url, "GET")
         return response
 
-    def update_submission_data(self, data, challenge_pk, submission_pk):
+    def update_submission_data(self, data, challenge_pk):
```
vkartik97: Can you make a separate PR for this change as it is not part of `Add tests for worker_util.py`

scripts/workers/worker_util.py
nikochiko: @vkartik97 :+1: I will also remove the test for `delete_message_for_sqs_queue`. It will cause an error without the correction so I think we can add it after the fix in #2575 is applied.

vkartik97: Sure 💯

scripts/workers/worker_util.py
nikochiko: Great! @takitsuse had opened an PR already: https://github.com/Cloud-CV/EvalAI/pull/2611

nikochiko: Done here! :+1:

scripts/workers/worker_util.py
nikochiko: @vkartik97 Done! 👍 Also verified that the Travis CI is passing.

nikochiko:
@vkartik97 As the PR (https://github.com/Cloud-CV/EvalAI/pull/2575) with corrections in delete_message_from_sqs_queue has been merged, I have brought back the test for it :+1:.
cc: @RishabhJain2018

tests/unit/rl_worker/test_worker_util.py
```diff
@@ -0,0 +1,172 @@
...
+    def test_update_submission_data_partially(self):
+        url = "{}{}".format(EVALAI_API_SERVER, self.update_submission_status_url(self.challenge_pk))
+        data = {"test_field": "new_value"}
+        response = self.interface.update_submission_status(data, self.challenge_pk)
+
+        self.mock_make_request.assert_called_with(url, "PATCH", data=data)
+        self.assertEqual(response, self.success_response)
+
+    def test_delete_message_from_sqs_queue(self):
+        test_receipt_handle = "MbZj6wDWli+JvwwJaBV+3dcjk2YW2vA3+STFFljTM8tJJg6HRG6PYSasuWXPJB+Cw"
```
vkartik97: Is this random string? If yes, how about using replacing hard coded string with a randomize function for alphanumeric and symbols?

nikochiko: Nice idea! I will create a helper a helper for that.

nikochiko: Done! 👍

PR #2577, opened 2020-01-04
Fix #2241 Allow unverified user to view the challenge

### LINKED ISSUE #2241: Allow unverified user to view the challenge
### Observed Behaviour
If an unverified user tries to view a challenge page, he is not able to do so and is told to verify the email first.

### Expected Behaviour
The unverified user should be allowed to view the challenge page and if he visits the `Participate` tab , an error stating to verify email address should be shown.

@RishabhJain2018 @deshraj

Suryansh5545:
I tried to use a different method but it didn't work so i have taken help from this previously made pr by thanos-pakou #2325 . In his code the user was directed toward the permission denied page . so instead of this the user is now displayed the message in the participate view only so he is not redirected. The problem that i am unable to solve is to how can i show the error messgae . At this time only error is showed .
Selection_021


KhalidRmb:
Hi @Suryansh5545 . So, you're saying that the challenge page is being shown now even for an unverified user. But you're facing some issue customizing the message on the Participate page. Is that right?

Suryansh5545:
Yes sir i was asking how can i show the error message . By the way there is a way to show message by copying the whole permission-denied page but it wouldn't look good

Suryansh5545:
test_video

KhalidRmb:
@Sanji515 Would you please have a look here? I'm not sure about the front-end part of this...Thanks.

Suryansh5545:
@KhalidRmb Sir How can i solve the message problem?

Suryansh5545:
I wasn't able to find anyway of showing the message so i have tried to insert it statically with the template also the default message unable to create or select a team doesn't make any sense when we are already showing a error below so i removed and placed the dynamic message instead of it

Suryansh5545:
Selection_026
Selection_027

Suryansh5545:
@Sanji515 Sir, I have added the response in the api you mentioned but i am unable to call it by using 
`parameters.url = 'participant_teams/challenges/' + vm.challengeId + '/user';`
.Also previously i tried calling the error message stored in error.details but the problem was that it was not displaying the error message .because of which i need to use static error message. So How can i resolve it

Sanji515:
> @Sanji515 Sir, I have added the response in the api you mentioned but i am unable to call it by using
> `parameters.url = 'participant_teams/challenges/' + vm.challengeId + '/user';`
> .Also previously i tried calling the error message stored in error.details but the problem was that it was not displaying the error message .because of which i need to use static error message. So How can i resolve it

@Suryansh5545 You don't need to call this API separately, just update this API to check if the email is verified and return the response and this API gets called when the challenge page gets loaded.

Suryansh5545:
Selection_029
I tried this logic but it doesn't seems to work.

Suryansh5545:
`@api_view(["GET"])
      @throttle_classes([UserRateThrottle])
      def get_teams_and_corresponding_challenges_for_a_participant(
       request, challenge_pk
      ):
    """
    Returns list of teams and corresponding challenges for a participant
    """
    if (HasVerifiedEmail):
        return Response(status=status.HTTP_202_OK)
    else:
        response_data = {
            "error": "Please verify your email first!"
        }
        return Response(response_data, status=status.HTTP_202_UNAUTHORIZED)
     # first get list of all the participants and teams related to the user
     participant_objs = Participant.objects.filter(
        user=request.user
      ).prefetch_related("team")

    is_challenge_host = is_user_a_host_of_challenge(
        user=request.user, challenge_pk=challenge_pk
    )

    challenge_participated_teams = []
    for participant_obj in participant_objs:
        participant_team = participant_obj.team

        challenges = Challenge.objects.filter(
            participant_teams=participant_team
        )

        if challenges.count():
            for challenge in challenges:
                challenge_participated_teams.append(
                    ChallengeParticipantTeam(challenge, participant_team)
                )
        else:
            challenge = None
            challenge_participated_teams.append(
                ChallengeParticipantTeam(challenge, participant_team)
            )
    serializer = ChallengeParticipantTeamListSerializer(
        ChallengeParticipantTeamList(challenge_participated_teams)
    )
    response_data = serializer.data
    response_data["is_challenge_host"] = is_challenge_host
    return Response(response_data, status=status.HTTP_200_OK)`

i don't know much about how it is working so i need a bit of guidance on how can i make it work

Sanji515:
Hey @Suryansh5545 You cannot call `HasVerifiedEmail` directly as its a permission class, try checking it first by writing the same code which is written inside the function named `has_permission` of `HasVerifiedEmail` class and then return the response.

Suryansh5545:
@Sanji515 Sir I have implemented the condition to check weather the email is verified or not . All the components work but the console is having some errors because of calling ChallengeCtrl two times 
Selection_030

In this the style error is coming because it belongs to participate.html file and main challenge file have no clue of this.
SO how can i resolve?

Suryansh5545:
Also now the api is implemented . how can we make the error message dynamic? storing it in utilities only work if we are sending the user to permission-denied page.

Suryansh5545:
And if we are to show different message as per the situation then we should have different icons to represent it .

Sanji515:
Hey @Suryansh5545 With the current code, it will show the message of `email verification` on navigating to challenge page, I think you need to update this API [here](https://github.com/Cloud-CV/EvalAI/blob/master/apps/challenges/views.py#L1776) in such a way that it should show email verification error when we try to start challenge not on the reload of challenge page.

Suryansh5545:
@Sanji515  Sir do i have to edit get_teams_and_corresponding_challenges_for_a_participant Api or star_challenge. Also i think star_challenge is called when star button is clicked so how can it check for it on start I am confused sir??

Sanji515:
> do i have to edit get_teams_and_corresponding_challenges_for_a_participant Api or star_challenge

I think you need to modify both API.

> Also i think star_challenge is called when star button is clicked

@Suryansh5545 No, it gets called on the load of page you can check [here](Also i think star_challenge is called when star button is clicked) as it have fetch the star count and whether the current user starred or not. That's why you also need to modify this API as well.

Suryansh5545:
@Sanji515 Now i understand why we need both API to check and also now it shows a nice red error when i click on star button Thank you Sir . But now just one thing remains. How can i use the permissionDenied template with dynamic message as currently we are using a static hardcoded html

Sanji515:
You can store another variable on localstorage and can modify the template a/c to it to render the same template under the participate tab.

Suryansh5545:
@Sanji515 I have figured out that the API were indeed already perfect and it was my mistake which costly this much of time. i have removed all the changes from the API and also have made the message dynamic.

Suryansh5545:
@Sanji515 By the way i was thinking that if we know what error message will probably be shown to us. we can change the image of the error . At this time only a email icon is shown. What do you say SIr?

Suryansh5545:
@Ram81 Sir i have fixed the coverage problem.

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -340,7 +341,11 @@
                             vm.displayDockerSubmissionInstructions(vm.page.is_docker_based, vm.isParticipated);
                             utilities.hideLoader();
                         },
-                        onError: function() {
+                        onError: function(response) {
+                            var error = response.data;                            
+                            $rootScope.notify("error", error.detail);
+                            vm.emailVerification = !true;
```
Sanji515: @Suryansh5545 Can we please directly set it to false here?

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -2103,4 +2105,4 @@
         
     }
 
-})();
+})();
```
Sanji515: Please add a new line here.

frontend/src/js/controllers/challengeCtrl.js
Suryansh5545: done

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -525,9 +530,6 @@
                 utilities.hideLoader();
             },
             onError: function(response) {
-                var error = response.data;
-                utilities.storeData('emailError', error.detail);
-                $state.go('web.permission-denied');
```
Sanji515: This way, you have only hidden the error, but the API still didn't get called due to which challenge phase tab is empty.

Suryansh5545: ok sir but if change it in api then how can i fix the test. like i dont know how is it being tested

Suryansh5545: can you tell me the file which is testing it?

Suryansh5545:
@Ram81 Sir i need a bit guidence on this. I don't know how is permission.py is being tested. I can't find the folder where the test is

Ram81:
@Suryansh5545 here https://github.com/Cloud-CV/EvalAI/blob/master/apps/accounts/permissions.py

Suryansh5545:
@Ram81 sir i know that but which file is testing this permission.py file?? . I can't understand that

Suryansh5545:
Sir The task timeline expired but i still want to complete it so please tell me how can i test it?

PR #2585, opened 2020-01-05
Add successful_submission_flag==True condition before saving submission_output in submission worker

daedaldan:
I added an if statement before the line that saves the submission_output to check if successful_submission_flag is True.

### DIFF
--- a/apps/challenges/serializers.py
+++ b/apps/challenges/serializers.py
@@ -113,6 +113,7 @@ class ChallengePhaseSplitSerializer(serializers.ModelSerializer):
             "challenge_phase_name",
             "dataset_split_name",
             "visibility",
+            "show_leaderboard_by_latest_submission"
         )
 
     def get_dataset_split_name(self, obj):
--- a/docs/source/architecture.md
+++ b/docs/source/architecture.md
@@ -6,7 +6,7 @@ EvalAI helps researchers, students, and data scientists to create, collaborate, 
 
 #### Django
 
-Django is the heart of the application, which powers our backend. We use Django version 1.11.18.
+Django is the heart of the application, which powers our backend. We use Django version 1.11.23.
 
 #### Django Rest Framework
 
--- a/scripts/workers/submission_worker.py
+++ b/scripts/workers/submission_worker.py
@@ -466,7 +466,7 @@ def run_submission(
                 # get split_code_name that is the key of the result
                 split_code_name = list(split_result.keys())[0]
 
-                # Check if the challenge_phase_split exists for the challenge_phaseand dataset_split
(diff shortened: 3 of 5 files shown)


pushkalkatara:
@daniel22373 Can you write a test for it.

daniel22373:
Just one test for the function I edited?

daniel22373:
Also, do I need to worry about this:

" No coverage uploaded for pull request base (master@bf36a9d). Click here to learn what that means (https://docs.codecov.io/docs/error-reference#section-missing-base-commit).
The diff coverage is 50%."
Thank you.

daniel22373:
> @daniel22373 Can you write a test for it.

I'm having some troubles writing the test, which I described in the task page on GCI. Could you help me?

KhalidRmb:
So, One of the ways is to run a test with forcing an exception inside the try catch blocks in the method, which after being caught sets the submission_successful flag to False.

daniel22373:
Hello, I wrote the test by mocking the challenge_phase_split.dataset_split attribute and adding an Exception side effect. After running the run_submission function, there is an assertEquals for the mocked output and the original output, which I saved, to test whether the if successful_submission_flag==True statement worked as intended.

However, when running the tests on my own machine, three of the submission_worker unit tests failed because of a ModuleNotFoundError (image attached), and I'm not sure why.
ModuleNotFoundError

Could you help me understand what's going on so I can figure out how to fix it?

daniel22373:
> Hello, I wrote the test by mocking the challenge_phase_split.dataset_split attribute and adding an Exception side effect. After running the run_submission function, there is an assertEquals for the mocked output and the original output, which I saved, to test whether the if successful_submission_flag==True statement worked as intended.
> 
> However, when running the tests on my own machine, three of the submission_worker unit tests failed because of a ModuleNotFoundError (image attached), and I'm not sure why.
> 
> Could you help me understand what's going on so I can figure out how to fix it?

Nevermind, I figured out what the issue was myself.

KhalidRmb:
Hey, a few pointers:
1. Why are you mocking the run_submission method? You are trying to test it right...it doesn't make sense to replace it with a mocked version. 
2. Also, the order of your parameters in the  mock function is not correct. Refer the documentation.
3. After reading the coverall report, I think we should be testing for a successful run_submission scenario. 

Thanks for the work! I think we can have a separate task for that, and you can pick it up from there if you want. You can remove the test you've added from this PR. 

@Ram81 @RishabhJain2018 We could approve this for now.

daniel22373:
Hi @KhalidRmb , thanks for the heads up, I appreciate the feedback! I updated the pull request and removed the test. I'll be glad to work on it again after this.

KhalidRmb:
Looks good. @ram81 @rishabhjain2018

Issue #2589, opened 2020-01-06
Add Collapsible to Challenge list


Suryansh5545:
Currently When user visits the evalai challenge list page . He is bombarded with all the challenges the on going, up coming and past ones . This make it hard for user to navigate and search. Currently there are 48! challenges on evalai challenge list .

Expected Solution:
Add a Collapsible so the user can open and close the list he wants . like if he want to look for current challenge then we will open on going tab .etc

Can i work on this

RishabhJain2018:
This is fixed in frontend v2.

Issue #2588, opened 2020-01-06
Scroll up Button not showing up 


Suryansh5545:
- **The expected behaviour**
In the challenge list the scroll up button should show up after going down a bit the page.  
- **The observed behaviour**
The scroll up button doesn't show up at all even after reaching at the bottom 
- steps to reproduce the problem
1.go to evalai.cloudcv.org/web/challenges/list
2. Scroll down 

Can i work on it?

RishabhJain2018:
@Kajol-Kumari Can you please check if this is fixed in frontend v2.

Kajol-Kumari:
> @Kajol-Kumari Can you please check if this is fixed in frontend v2.

Yes @RishabhJain2018 It's working perfectly fine for frontend_v2.

devanshraj3000:
Though the scroll up button is visible in the page but it's width is quite less and a lot of space is getting wasted in left and the right side, so we can increase the width of scroll bar a little to make it user friendly.
Should I do this?

shikhar168:
i would like to work on this issue

PR #2590, opened 2020-01-06
Fix:#2588 #2589 Fixed the Scroll Up Button && Added a Collapsible to the Challenge List

Suryansh5545:
#2588 #2589
**Minor Change:**
I have Fixed the scroll button.
**Major Change:**
Made All three challenge tab list into a collapsible so user can select which one he wants to look to.

Reason:- When we visit evalai challenge list it is not good to bombard the user with so many challenges. At this time there are 48! Challenges(*Including Past and Upcoming one).  Which makes it harder to navigate and search for the challenge.
Selection_028


Suryansh5545:
@Sanji515 Sir Can you please take a look and tell me if it is good ?

PR #2593, opened 2020-01-06
Check feasibility of changing worker from Celery to Dramtiq

nikochiko:
PR for codebase changes. 
This branch uses dramatiq and dramatiq_sqs to replace Celery.
The spike report (in another branch): https://github.com/nikochiko/EvalAI/blob/SpikeDramatiq/evalai/spike_dramatiq.md
 
Note: django_dramatiq is not being used here.


nikochiko:
Changing the endpoint_url from `http://127.0.0.1:9324` to `http://sqs:9324` seems to have solved the problem.
Thanks @yashdusing for pointing it out. 
Setting up dev env worked on my local env. To share the logs in a readable format, I have tried to do the same on a Travis VM: https://travis-ci.org/nikochiko/Test-In-Travis/builds/633625612

Note: The travis build here fails because the change in endpoint url works for docker containers (as on local env) whereas here we run elasticmq via bash, (https://github.com/Cloud-CV/EvalAI/blob/master/.travis.yml#L36)

yashdusing:
LGTM :+1: 
Just make the changes so that sqs:9324 is used in docker containers while 127.0.0.1:9324 is used in travis (bash)

nikochiko:
@yashdusing Done! :+1:

docker/dev/docker.env
```diff
@@ -1,10 +1,10 @@
 AWS_ACCESS_KEY_ID=x
 AWS_SECRET_ACCESS_KEY=x
-AWS_DEFAULT_REGION=us-east-1
+AWS_DEFAULT_REGION=elasticmq
 AWS_ACCOUNT_ID=x
+AWS_ENDPOINT_URL=http://sqs:9324
```
yashdusing: @nikochiko could you change this to AWS SQS endpoint url ?

yashdusing:
Please fix the travis as well

nikochiko:
@yashdusing Done here! 😄 
BTW, has a decision been made on whether we will be using dramatiq in production?

requirements/dev.txt
```diff
@@ -19,6 +21,6 @@ sphinx==1.7.2
 sphinx-autobuild==0.7.1
 sphinx-markdown-tables==0.0.9
 sphinx-rtd-theme==0.3.0
+prometheus-client==0.7.1
```
yashdusing: @nikochiko do we need this ?

nikochiko: @yashdusing Yes, it is a dependency of dramatiq_sqs

nikochiko: Reference:  https://github.com/Bogdanp/dramatiq/blob/master/setup.py#L42-L44

yashdusing:
@nikochiko not yet. As you mentioned, celery has better benchmarks for production but this can be used for development branch

requirements/dev.txt
yashdusing: @nikochiko could you also check the coverage ?

nikochiko:
@yashdusing Okay :+1:

nikochiko:
So this PR cannot be put into the production branch either as we changed `app.task` to `dramatiq.actor`.
Should we add conditional statements that check the environment before using an API (celery or dramtiq)?

nikochiko:
Capture
I don't see a file whose coverage has gone down. I think it is because dramatiq_conf.py has fewer lines than celery.py so the percentage of covered lines in evalai directory has gone down.
I think this can be ignored.

PR #2597, opened 2020-01-08
Add citation bibentry on homepage under "Cite our work"

matsujika:
**The corresponding GCI task**: https://codein.withgoogle.com/dashboard/task-instances/5953309093920768/

**Changes**:
- Add a citation bibentry copied from [reference](https://github.com/Cloud-CV/EvalAI#citing-evalai).

**Screencast**:
Screenshot from 2020-01-08 18-17-09

**CC**: @vkartik97 @Ram81 @RishabhJain2018

### DIFF
--- a/frontend/src/views/web/landing.html
+++ b/frontend/src/views/web/landing.html
@@ -281,6 +281,23 @@
                         <img src="dist/images/evalai-paper.jpg" alt="" width="100%" class="thumb">
                     </a>
                 </div>
+                <div class="row">
+                    <div class="col s12 m12 l12">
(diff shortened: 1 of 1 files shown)


pushkalkatara:
Maybe we should add the arvix link too [https://arxiv.org/abs/1902.03570](https://arxiv.org/abs/1902.03570) . Easy for someone to navigate and use any type of citation format.

takitsuse:
@pushkalkatara I think we should create another PR for your suggestion. What I am requested to add is not arxiv link but citation bibentry, and I think UNIX philosophy "make each program do one thing well." applies here.

PR #2604, opened 2020-01-10
AWS utils: Add unit tests for restart_workers

nikochiko:
<b>Changes proposed in this PR:</b>
* Add tests for aws_utils.restart_workers
* Mock the intermediate function service_manager

Reference PR: https://github.com/Cloud-CV/EvalAI/pull/2413

### DIFF
--- a/tests/unit/challenges/test_aws_utils.py
+++ b/tests/unit/challenges/test_aws_utils.py
@@ -196,8 +196,3 @@ class TestRestartWorkers(BaseAdminCallsClass):
         response = aws_utils.restart_workers(queryset)
         self.assertEqual(response, expected_response)
         self.assertEqual(list(c.workers for c in queryset), expected_num_of_workers)
-
-    def test_create_service_by_challenge_pk(self):
-        response = aws_utils.create_service_by_challenge_pk(self.ecs_client, self.challenge, self.client_token)
-        self.assertEqual(response["ResponseMetadata"]["HTTPStatusCode"], HTTPStatus.OK)
-        self.assertEqual(self.challenge.workers, 1)
--- a/tests/unit/challenges/test_aws_utils.py
+++ b/tests/unit/challenges/test_aws_utils.py
@@ -198,6 +198,7 @@ class TestRestartWorkers(BaseAdminCallsClass):
         self.assertEqual(list(c.workers for c in queryset), expected_num_of_workers)
 
     def test_create_service_by_challenge_pk(self):
+        self.ecs_client.create_cluster(clusterName=aws_utils.COMMON_SETTINGS_DICT["CLUSTER"])
         response = aws_utils.create_service_by_challenge_pk(self.ecs_client, self.challenge, self.client_token)
         self.assertEqual(response["ResponseMetadata"]["HTTPStatusCode"], HTTPStatus.OK)
         self.assertEqual(self.challenge.workers, 1)
--- a/tests/unit/challenges/test_aws_utils.py
+++ b/tests/unit/challenges/test_aws_utils.py
@@ -4,7 +4,6 @@ import os
 
 from moto import mock_ecs
 from datetime import timedelta
-from django.core.files.uploadedfile import SimpleUploadedFile
 from django.utils import timezone
 
 from allauth.account.models import EmailAddress
@@ -182,8 +181,8 @@ class TestRestartWorkers(BaseAdminCallsClass):
     @mock.patch("challenges.aws_utils.service_manager")
     def test_restart_workers_when_service_manager_fails_for_second_challenge(self, mock_sm):
         self.set_challenge_workers(self.challenge, 1)
-        self.set_challenge_workers(self.challenge, 1)
-        self.set_challenge_workers(self.challenge, 1)
+        self.set_challenge_workers(self.challenge2, 1)
+        self.set_challenge_workers(self.challenge3, 1)
 
         mock_sm.side_effect = [self.response_OK, self.response_BAD_REQUEST, self.response_OK]
 
--- a/tests/unit/challenges/test_aws_utils.py
+++ b/tests/unit/challenges/test_aws_utils.py
@@ -1,0 +1,210 @@
+import mock
+import boto3
+import os
+
+from unittest import TestCase
+
+from moto import mock_ecs
+from datetime import timedelta
+from django.core.files.uploadedfile import SimpleUploadedFile
+from django.utils import timezone
+
+from allauth.account.models import EmailAddress
+from django.contrib.auth.models import User
+from http import HTTPStatus
+
+from rest_framework.test import APITestCase, APIClient
+
+from hosts.models import ChallengeHost, ChallengeHostTeam
+
+import challenges.aws_utils as aws_utils
+from challenges.models import Challenge, ChallengePhase
+
+
+class BaseTestClass(APITestCase):
+    def setUp(self):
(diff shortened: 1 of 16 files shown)


tests/unit/challenges/test_aws_utils.py
```diff
@@ -0,0 +1,199 @@
+import mock
+import boto3
```
Ram81: @nikochiko can you reorder the imports lexicographically

tests/unit/challenges/test_aws_utils.py
```diff
@@ -0,0 +1,199 @@
...
+
+        self.challenge3 = Challenge.objects.create(
+            title="Test Challenge 3",
+            creator=self.challenge_host_team,
+            queue="test_queue_3",
+            start_date=timezone.now() - timedelta(days=2),
+            end_date=timezone.now() + timedelta(days=1),
+        )
+        self.ecs_client.create_cluster(clusterName="cluster")
+        self.client_token = "abc123"
```
Ram81: @nikochiko can we use a randomly generated string here

tests/unit/challenges/test_aws_utils.py
```diff
@@ -0,0 +1,199 @@
...
+
+        self.set_challenge_workers(self.challenge, 1)
+        self.set_challenge_workers(self.challenge2, 1)
+        self.set_challenge_workers(self.challenge3, 1)
+
+        pklist = [self.challenge.pk, self.challenge2.pk, self.challenge3.pk]
+        queryset = super(TestRestartWorkers, self).queryset(pklist)
+
+        expected_count = 3
+        expected_num_of_workers = [1, 1, 1]
```
Ram81: Can this `1` be declared as a class variable so that we can use the variable name where needed

tests/unit/challenges/test_aws_utils.py
```diff
@@ -0,0 +1,199 @@
...
+
+        self.set_challenge_workers(self.challenge, 0)
+        self.set_challenge_workers(self.challenge2, 1)
+        self.set_challenge_workers(self.challenge3, 1)
+
+        pklist = [self.challenge.pk, self.challenge2.pk, self.challenge3.pk]
+        queryset = super(TestRestartWorkers, self).queryset(pklist)
+
+        expected_count = 2
+        expected_num_of_workers = [0, 1, 1]
```
Ram81: Same here

tests/unit/challenges/test_aws_utils.py
```diff
@@ -0,0 +1,199 @@
...
+        self.set_challenge_workers(self.challenge, 1)
+        self.set_challenge_workers(self.challenge2, 1)
+        self.set_challenge_workers(self.challenge3, 1)
+
+        mock_sm.side_effect = [self.response_OK, self.response_BAD_REQUEST, self.response_OK]
+
+        pklist = [self.challenge.pk, self.challenge2.pk, self.challenge3.pk]
+        queryset = super(TestRestartWorkers, self).queryset(pklist)
+
+        expected_count = 2
```
Ram81: Same here

tests/unit/challenges/test_aws_utils.py
nikochiko: @Ram81 I think it is better if we keep it this way. As it states the number of workers and could be any number (0, 1, 2, 3, etc.) the number of variables would be too much to handle. 
What do you think?

tests/unit/challenges/test_aws_utils.py
nikochiko: Done :+1:

tests/unit/challenges/test_aws_utils.py
nikochiko: Done :+1:

tests/unit/challenges/test_aws_utils.py
```diff
@@ -0,0 +1,199 @@
+import boto3
+import mock
+import os
+
+from moto import mock_ecs
+from datetime import timedelta
```
Ram81: @nikochiko same here for imports they should be lexicographically ordered based on package name

tests/unit/challenges/test_aws_utils.py
Ram81: @nikochiko but the number is based on the number of challenges right? so can't we use that to compute the number of workers?

Ram81: Also as I can see in the code above you are setting the number of workers for every challenge. So I think we can just create 1 variable with a value that can be used in both setting the number of worker and using that in `expected_num_of_workers`. Let me know if this makes sense

nikochiko: @Ram81 Okay, I think we can create a variable for the first case. (so something like `self.one_active_worker`?  Let me know if I have the right idea).
For the second case though, I want to clarify that it is the `expected_count` and not `expected_num_of_workers`. Both are different in that the expected count describes the number of workers successfully restarted during the process. So that can be different from the num_of_workers. For example in this case: https://github.com/Cloud-CV/EvalAI/pull/2604/files#diff-cc74cf6292b78dc74ea17124c792900eR192-R193

tests/unit/challenges/test_aws_utils.py
nikochiko: Yes, I just copy-pasted that from the reference PR, so I didn't pay much attention. I will fix it right away.

tests/unit/challenges/test_aws_utils.py
nikochiko: > but the number is based on the number of challenges right? so can't we use that to compute the number of workers?

I'm sorry I didn't get what you meant by computing the number of workers. Can you please elaborate?

nikochiko: If you are referring to computing the `expected_count`, I want to clarify that it is only the number of workers successfully restarted and not the number of active workers. It can be different for example in this case: https://github.com/Cloud-CV/EvalAI/pull/2604/files#diff-cc74cf6292b78dc74ea17124c792900eR192-R193

tests/unit/challenges/test_aws_utils.py
```diff
@@ -0,0 +1,201 @@
...
+from moto import mock_ecs
+
+from allauth.account.models import EmailAddress
+from django.contrib.auth.models import User
+from django.utils import timezone
+
+from rest_framework.test import APITestCase, APIClient
+
+import challenges.aws_utils as aws_utils
+from challenges.models import Challenge
```
Ram81: @nikochiko please reorder all imports lexicographically. Another thing all `from` imports should come at last and all the direct `import` statements should be at start. Your lines 5-7 are correct but all from imports should be considered together to order lexicographically

nikochiko: Done! 👍 Please see if it is okay now.

yashdusing:
LGTM :+1:

KhalidRmb:
@yashdusing Could you comment here what the error was with the deploymentController? I can't find your comment. Thanks!

yashdusing:
@KhalidRmb I thought it had to do with braces and unpacking. But, after inspecting the logs, I inferred that botocore was being reinstalled to a lower version after awscli was installed

KhalidRmb:
Oh...Thanks for the catch! @yashdusing

nikochiko:
@yashdusing Thanks for the fix, nice catch! 👌🎉

Issue #2574, opened 2020-01-03
Improper error message


hkmatsumoto:
**The observed behavior**:
When I input Japanese words into team name, the error message "This field is required" is shown.
![Screenshot from 2020-01-04 02-06-17](https://user-images.githubusercontent.com/57856193/71737568-5e953e80-2e97-11ea-87f9-22d276e3d0f1.png)

**The expected behavior**:
When unexpected letter (not limited to Japanese letter) is in input, the error message should be something like "Invalid letter is contained".

drepram:
Can you try this out in EvalAI-ngx too?

takitsuse:
Sure :smile_cat:

Tanuj22:
@takitsuse Are you solving this or can I?

takitsuse:
@Tanuj22 I'm not working on this, so go ahead :+1:

Tanuj22:
I was looking into this issue. Does creating a regex to allow only english letter sound good?

takitsuse:
Sounds good to me. What do you say @Sanji515?

deshraj:
We are only planning to support only alphanumeric characters for the team name going forward. 

Here are the updated rules:
- Team name can be contain alphanumeric characters with spaces, `-`, and `_`
- Show proper error message if not correct team name is entered

\cc @Kajol-Kumari

RishabhJain2018:
Closing this as we already have this.

PR #2605, opened 2020-01-10
 Fix#2574 Fixed Improper error message in team name

Tanuj22:
#2574

**Changes**

Screenshot from 2020-01-10 22-41-39

All non-english characters input show "Invalid letter is contained" error message.

### DIFF
--- a/apps/challenges/serializers.py
+++ b/apps/challenges/serializers.py
@@ -113,6 +113,7 @@ class ChallengePhaseSplitSerializer(serializers.ModelSerializer):
             "challenge_phase_name",
             "dataset_split_name",
             "visibility",
+            "show_leaderboard_by_latest_submission"
         )
 
     def get_dataset_split_name(self, obj):
--- a/apps/hosts/models.py
+++ b/apps/hosts/models.py
@@ -4,6 +4,7 @@ from django.db import models
 from django.contrib.auth.models import User
 
 from base.models import TimeStampedModel
+from django.core.validators import RegexValidator
 
 # from challenges.models import (Challenge, )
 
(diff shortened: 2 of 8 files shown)


Suryansh5545:
@Tanuj22 I am not a mentor . but i will like to suggest that we should do the check in the [API](https://github.com/Cloud-CV/EvalAI/blob/master/apps/hosts/views.py#L32) and then return the response which can be then shown by the onError function . What do you say?

Tanuj22:
@Suryansh5545 Seems to be the better option! Thanks!

Suryansh5545:
@Tanuj22 The error message for improper error message is not displayed. I think you have implemented it in a wrong file. When the create team button is clicked this [API](https://github.com/Cloud-CV/EvalAI/blob/master/apps/hosts/views.py#L213) `create_challenge_host_team` is called. you have to implement it in that and then return the response with error message. also as this is the fix for improper team name can you please add this functionality to participate tab. For this you have to edit this [API](https://github.com/Cloud-CV/EvalAI/blob/master/apps/participants/views.py#L47).

Tanuj22:
@Suryansh5545 I have changes in the model `ChallengeHostTeam`  to accept only english characters which is better than making changes in the API I think. As for adding it in the participate tab I can do that.

Suryansh5545:
@Tanuj22 The error message for invalid character is still the same please fix it. And please update the branch

Tanuj22:
@Suryansh5545 I have checked it again now. Could you recheck?  I am attaching a GIF for the same.

ezgif-7-e645f4a2b945

Suryansh5545:
@Tanuj22 It is working now . :+1:

Suryansh5545:
Now if you can please add this functionality to participate.

Tanuj22:
@Suryansh5545  Done

Suryansh5545:
@Tanuj22 LGTM:+1:


@Sanji515 @Ram81 @RishabhJain2018 @vkartik97 Please take a look

PR #2606, opened 2020-01-10
Add Tests for worker_util.py

nsjcorps:
I have added tests for the worker_util which is used by the RL-worker to communicate with EvalAI Server to evaluate Docker Based Challenge submissions

### DIFF
--- a/tests/integration/worker/test_worker_util.py
+++ b/tests/integration/worker/test_worker_util.py
@@ -17,6 +17,7 @@ class BaseAPITestClass(APITestCase):
     def setUp(self):
         self.evalai_interface = EvalAI_Interface(AUTH_TOKEN, EVALAI_API_SERVER, QUEUE_NAME)
         self.test_url = "/test-url"
+        self.data = {"test": "data"}
 
 
 class TestGetRequestHeaders(BaseAPITestClass):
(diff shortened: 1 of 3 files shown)


nsjcorps:
I have finished writing tests for most of the file. However, I do not know what the response of the EvalAI server is supposed to be. Eg. I do not know what the get_message_from_sqs_queue function returns from the server. How can I find out what they actually are? @utsav @vkartik97 @RishabhJain2018 @Ram81

vkartik97:
> I do not know what the get_message_from_sqs_queue function returns from the server.

You can try using `Postman` to get responses of all the endpoints used in EvalAI Interface.

nsjcorps:
@vkartik97 As mentioned before sir, I can't use docker. So I won't be able to start the server. Is there anyway I can get the expected response sir?

vkartik97:
You can have a look at view which is endpoints for : /api/jobs/challenge/queues/{QUEUE_NAME}

nsjcorps:
I did. But it doesn't show me what the json would look like @vkartik97

nsjcorps:
It only says that it returns some data from the database @vkartik97

nsjcorps:
I don know what the data from the database looks like @vkartik97

nsjcorps:
Never mind sir. I have found the way to test the function. Thank you very much  @vkartik97

nsjcorps:
@vkartik97 I have written the test cases after a load of commits ;)...Please take a look

tests/integration/worker/test_worker_util.py
```diff
@@ -0,0 +1,127 @@
...
+    def test_return_url_per_environment(self):
+        expected = self.url_per_environment
+        result = self.evalai_interface.return_url_per_environment(self.test_url)
+        self.assertEqual(expected, result)
+
+
+@mock.patch("scripts.workers.worker_util.EvalAI_Interface.return_url_per_environment")
+@mock.patch("scripts.workers.worker_util.EvalAI_Interface.make_request")
+class APICallsTestClass(BaseAPITestClass):
+
```
pushkalkatara: please remove this extra space.

tests/integration/worker/test_worker_util.py
```diff
@@ -0,0 +1,127 @@
+import mock
+
+from rest_framework.test import APITestCase
+from scripts.workers.worker_util import EvalAI_Interface, URLS
+
+
+AUTH_TOKEN = "mock_auth_token123456"
```
pushkalkatara: we can define these variables in setUp itself.

tests/integration/worker/test_worker_util.py
nsjcorps: done

pushkalkatara:
@vkartik97 Looks good to me. We can merge this and approve the GCI task.

nsjcorps:
Thank you @pushkalkatara

tests/integration/worker/test_worker_util.py
nsjcorps: Done!! @pushkalkatara

PR #2611, opened 2020-01-12
Fix #2609: Remove unused argument in update_submission_data

### LINKED ISSUE #2609: Unused argument in update_submission_data
**Current**:
`update_submission_data` function is implemented [here](https://github.com/Cloud-CV/EvalAI/blob/master/scripts/workers/worker_util.py#L88) and [there](https://github.com/Cloud-CV/EvalAI/blob/master/scripts/workers/remote_submission_worker.py#L447), but none of them use `submission_pk` argument in their codes.

**Proposal**:
I think we should remove the unused argument.

P.S: I'd like to work on this if we get a consensus on removing it.

matsujika:
Fix #2609

### DIFF
--- a/scripts/workers/remote_submission_worker.py
+++ b/scripts/workers/remote_submission_worker.py
@@ -444,7 +444,7 @@ def get_challenge_phase_by_pk(challenge_pk, challenge_phase_pk):
     return response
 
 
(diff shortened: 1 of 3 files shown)


RishabhJain2018:
@matsujika Can you please resolve the conflicts here?

PR #2614, opened 2020-01-13
Stop evaluating booleans

drepram:
This is one of the fix of the problem that has occurred on EvalAI-ngx, particularly https://github.com/Cloud-CV/EvalAI-ngx/issues/249.

The changes I've proposed removed the evaluation though I've not been able to resolve the angular-devkit issue.

Here are the things that I've read and have tried:

https://github.com/akveo/ngx-admin/issues/2138
https://stackoverflow.com/questions/56393158/errors-data-path-buildersapp-shell-should-have-required-property-class
https://stackoverflow.com/questions/56542518/schema-validation-failed-with-the-following-errors-data-path-buildersapp-sh/56624167
https://stackoverflow.com/questions/56506473/ionic-problem-ngschema-validation-failed-with-the-following-errorsngdata-p/56537342#56537342
https://github.com/bpmn-io/bpmn-js-example-angular/issues/15

Perhaps someone else can give me some insight on how I can resolve this issue?


scripts/workers/submission_worker.py
```diff
@@ -696,7 +696,7 @@ def main():
         q_params["pk"] = challenge_pk
 
     if settings.DEBUG or settings.TEST:
-        if eval(LIMIT_CONCURRENT_SUBMISSION_PROCESSING):
+        if LIMIT_CONCURRENT_SUBMISSION_PROCESSING:
```
RishabhJain2018: @drepram Why do we need to remove this?

PR #2617, opened 2020-01-14
Add update_submission_status to URLS

matsujika:
**Changes**:
- Add `update_submission_status` to `URLS` and make a use of it.


scripts/workers/worker_util.py
```diff
@@ -12,6 +12,7 @@
     "get_challenge_by_queue_name": "/api/challenges/challenge/queues/{}/",
     "get_challenge_phase_by_pk": "/api/challenges/challenge/{}/challenge_phase/{}",
     "update_submission_data": "/api/jobs/challenge/{}/update_submission/",
+    "update_submission_status": "/api/jobs/challenge/{}/update_submission/",
```
vkartik97: Hey, I missed this.
Can you remove this line as it is same as `update_submission_data`?

scripts/workers/worker_util.py
```diff
@@ -92,7 +93,7 @@ def update_submission_data(self, data, challenge_pk, submission_pk):
         return response
 
     def update_submission_status(self, data, challenge_pk):
-        url = "/api/jobs/challenge/{}/update_submission/".format(challenge_pk)
+        url = URLS.get("update_submission_status").format(challenge_pk)
```
vkartik97: `update_submission_status` -> `update_submission_data`

scripts/workers/worker_util.py
takitsuse: @vkartik97 Okay

scripts/workers/worker_util.py
takitsuse: How about deleting `update_submission_status` and changing `update_submission_data` to `update_submission`?

takitsuse: @vkartik97 Since `update_submission_data` will be used by both `*_data` and `*_status`, I think the change will be reasonable.

RishabhJain2018:
@matsujika Can you please resolve the conflicts here?

PR #2627, opened 2020-01-18
Update awscli, botocore, boto3 and moto in Travis

nikochiko:
Reference: https://github.com/Cloud-CV/EvalAI/pull/2604
Solve the issue of botocore requirement getting overwritten during travis.

#### Changes proposed:
* Update awscli, botocore, boto3 and moto to their latest versions


nikochiko:
#### Results for the first attempt:
Updating all to their latest versions is breaking the build. I think this is probably due to updating botocore. 
#### Traceback from the Travis logs:
```
self = <botocore.auth.SigV4Auth object at 0x7f754d7b5710>
request = <botocore.awsrequest.AWSRequest object at 0x7f754d7b54e0>
    def add_auth(self, request):
        if self.credentials is None:
>           raise NoCredentialsError
E           botocore.exceptions.NoCredentialsError: Unable to locate credentials
```
#### Error occurs while:
1. Trying to create queues (specifically while testing `get_or_create_sqs_queue`)

#### Travis logs:
https://travis-ci.org/Cloud-CV/EvalAI/builds/638741307

nikochiko:
### Second update:
Only mention awscli in .travis.yml and install awscli's version of botocore

#### Results:
Build is passing. I added some tests for aws_utils to confirm that the versions are compatible with each other.

#### Notes:
botocore requirement got overwritten by awscli.
Travis throws this warning:
```
boto3 1.9.88 has requirement botocore<1.13.0,>=1.12.88, but you'll have botocore 1.14.5 which is incompatible.
boto3 1.9.88 has requirement s3transfer<0.2.0,>=0.1.10, but you'll have s3transfer 0.3.1 which is incompatible.
aws-sam-translator 1.20.1 has requirement jsonschema~=3.0, but you'll have jsonschema 2.6.0 which is incompatible.
```

yashdusing:
@nikochiko instead of doing this. Could you try installing awscli and coveralls first.
Then install requirements/dev.txt
and in requirements/dev.txt remove boto3, botocore, moto versions

nikochiko:
@yashdusing Yes! Doing the same.
Directly making the changes from stackoverflow (https://stackoverflow.com/questions/51911075/how-to-check-awscli-and-compatible-botocore-package-is-installed) is causing pip install to use the boto3 cached during installing from requirements/dev.txt

.travis.yml
```diff
@@ -42,7 +42,7 @@ services:
 - xvfb
 install:
 - pip install -r requirements/dev.txt
-- pip install awscli==1.16.57 coveralls
+- pip install --upgrade awscli boto3 botocore moto
```
yashdusing: update boto3, botocore, moto in requirements/dev.txt as well

nikochiko: Is it okay if the changes are made in requirements/common.txt? 
https://github.com/Cloud-CV/EvalAI/blob/master/requirements/common.txt#L2-L3

yashdusing: yes. We need changes in both

requirements/common.txt
```diff
@@ -10,13 +10,13 @@ django-import-export==0.5.1
...
 django-cors-headers==1.3.1
-django-rest-auth[with_social]==0.9.2 
+django-rest-auth[with_social]==0.9.2
 django-ses==0.8.5
 docker-compose==1.21.0
 drfdocs==0.0.11
 drf-yasg==1.11.0
 kubernetes==10.0.1
-moto==1.3.8
+moto
```
yashdusing: moto should be in dev

yashdusing: nvm, since it was in common you can keep it here

nikochiko: Okay!

nikochiko:
I have updated the required deps: awscli, boto3, botocore and moto
I have also verified that they are working by writing some sample aws_utils tests. I think the changes in the PR have been tested well and the fix is working. :+1:

I have removed the additional tests now.

nikochiko:
As pointed out by @yashdusing the reason the travis build was failing was because the tests in submission_worker were creating an actual boto3 object and not one mocked with moto. This was solved by moving the mock_sqs decorator to base class.

yashdusing:
LGTM 👍

yashdusing:
@Ram81 @vkartik97 @Sanji515 @RishabhJain2018 please review 😄

PR #2629, opened 2020-01-19
Integrate celery worker for prod and dramatiq worker for dev env

nikochiko:
#### Changes proposed:
* Define a new wrapper for tasks
* Create common method for the value returned by this wrapper to push tasks to queue (`send` in dramatiq and `delay` in celery)

Reference: #2593


nikochiko:
I have built this PR on top of PR#2593.
See this commit to see the changes specific to this PR: https://github.com/Cloud-CV/EvalAI/pull/2629/commits/a42fb548a88a680a71a8b1143a445251af3df72f

Only a few linting changes have been added after this.

docker/dev/celery/container-start.sh
```diff
@@ -1,3 +0,0 @@
-#!/bin/sh
-cd /code && \
-celery -A evalai worker --loglevel=INFO
```
yashdusing: @nikochiko we don't have to delete this

evalai/dramatiq.py
```diff
@@ -0,0 +1,21 @@
...
+    namespace="dramatiq_sqs_tests",
+    middleware=[
+        Prometheus(),
+        AgeLimit(),
+        TimeLimit(),
+        Callbacks(),
+        Pipelines(),
+        Retries(min_backoff=1000, max_backoff=900000, max_retries=96),
+    ],
+    endpoint_url=os.environ.get("AWS_SQS_ENDPOINT_URL", "http://localhost:9324"),
```
yashdusing: change to "sqs:9324"

docker/dev/celery/container-start.sh
nikochiko: @yashdusing This is in `docker/dev`. I don't think we will need celery in dev env?

yashdusing: Nevermind this comment. I thought the dockerfile and container script was shared between the prod & dev.

evalai/dramatiq.py
nikochiko: @yashdusing The issue here is that we define sqs://9324 in docker.env file for dev env, so the locak dev env would work but Travis uses ElasticMQ on localhost:9324 only. So if we change it to sqs:9324, then the local dev env would work but travis build would fail.

nikochiko: What do you think about this?
```python
if settings.DEBUG:
    sqs_endpoint = os.environ.get("AWS_SQS_ENDPOINT_URL", "sqs:9324")
else:
    sqs_endpoint = os,environ.get("AWS_SQS_ENDPOINT_URL", "localhost:9324")
```

yashdusing: yes, this sounds better. You may also add a comment to describe 'for travis', 'for local'

nikochiko:
Ah, the Travis build isn't starting automatically here :/
@yashdusing Thanks for this pointer: https://travis-ci.community/t/known-issue-travis-ci-reports-expected-waiting-for-status-to-be-reported-on-the-github-status-api-but-the-status-never-arrives/1154

nikochiko:
Moving to here: https://github.com/Cloud-CV/EvalAI/pull/2630

PR #2630, opened 2020-01-19
Integrate celery worker for prod and dramatiq worker for dev env

nikochiko:
#### Changes proposed:
 * Define a new wrapper for tasks
 * Create common method for the value returned by this wrapper to push tasks to queue (send in dramatiq and delay in celery)

Reference: #2593


yashdusing:
LGTM :+1:

evalai/worker_config.py
```diff
@@ -0,0 +1,18 @@
+from django.conf import settings
+
+
+def task_wrapper(fn, *args, **kwargs):
+    def inner(fn):
+        if settings.DEBUG:
+            import dramatiq  # noqa
+            from evalai.dramatiq_conf import broker  # noqa
```
vkartik97: Can you explain the reason of why is this not on top and here?

nikochiko: @vkartik97 The dramatiq import isn't required in the other case in prod env. 
Also, we don't have dramatiq in `requirements/prod.txt` so having it at the top would cause an error in prod and test envs.

vkartik97: Can this moved before `def task_wrapper` ?

nikochiko: Oh right! That will make the code much cleaner! 😄 
Thanks for the suggestion that had slipped my mind.

nikochiko: Done! :+1:

Issue #2666, opened 2020-02-13
Unclear Naming to variables


Kajol-Kumari:
Current Behaviour:
Many Variable names are unclear and creates confusion regarding their purpose.

Expected Behaviour:
Variable names should clearly describe its purpose.

References:
1. https://github.com/Cloud-CV/EvalAI/blob/master/frontend/tests/controllers-test/challengeHostTeamsCtrl.test.js#L418
Here, the test is written for delete function for hostTeam where Variable name "ParticipantId" is not appropriate and hence leading to the confusion.

2. https://github.com/Cloud-CV/EvalAI/blob/master/frontend/tests/controllers-test/challengeHostTeamsCtrl.test.js#L441
Here, the controller function is written for delete function for hostTeam where Variable name "ParticipantId" is not appropriate and hence leading to the confusion.

3. https://github.com/Cloud-CV/EvalAI/blob/master/apps/hosts/urls.py#L27
According to the name the url has got, it should have been for getting details of the hostTeam but instead it's doing all of the get,post,patch and delete operations on the challenge_host_team

RishabhJain2018:
Hi @Kajol-Kumari,
Thanks for opening the issue and the corresponding PR. We would be happy to incorporate the suggestion. Can you please write in the description here why the variable names are incorrect and are creating confusion?

PR #2667, opened 2020-02-13
Improving_code_readability

Kajol-Kumari:
Fix: #2666 

Variable have understandable name that clearly describes it's functionality.


apps/hosts/views.py
```diff
@@ -152,7 +152,7 @@ def challenge_host_list(request, challenge_host_team_pk):
 @throttle_classes([UserRateThrottle])
 @permission_classes((permissions.IsAuthenticated, HasVerifiedEmail))
 @authentication_classes((ExpiringTokenAuthentication,))
-def challenge_host_detail(request, challenge_host_team_pk, pk):
+def challenge_host_delete(request, challenge_host_team_pk, challenge_host_pk):
```
RishabhJain2018: I think we should not change the name of this function since it takes all the requests `GET`, `PUT`, `PATCH` and `DELETE`.

apps/hosts/views.py
```diff
@@ -162,7 +162,7 @@ def challenge_host_detail(request, challenge_host_team_pk, pk):
         return Response(response_data, status=status.HTTP_406_NOT_ACCEPTABLE)
 
     try:
-        challenge_host = ChallengeHost.objects.get(pk=pk)
+        challenge_host = ChallengeHost.objects.get(pk=challenge_host_pk)
```
RishabhJain2018: Can we please use the `get_model_object` function here to fetch the object. For reference, please see - https://github.com/Cloud-CV/EvalAI/blob/master/apps/challenges/utils.py#L25

Kajol-Kumari: @RishabhJain2018 Sir, when i used get_model_object, one of the test failed as the error now is thrown is in different format i.e it's attaching id also along with other details like code_error. Any suggestions regarding, how can i deal with it sir as i am unable to fix it.

nikochiko: @Kajol-Kumari You can edit the tests accordingly. 
For example, see [here](https://github.com/Cloud-CV/EvalAI/blob/master/tests/unit/challenges/test_views.py#L359-L369).
As an improvement on the mentioned code, you can store the invalid pk as `self.invalid_challenge_pk` and use it in `kwargs` and `expected`.
cc: @RishabhJain2018

Kajol-Kumari: @nikochiko the problem is i am unable to get that ErrorDetail string as when i am trying to access it, it is showing undefined "detail", undefined "ErrorDetail"

apps/hosts/views.py
```diff
@@ -162,7 +167,7 @@ def challenge_host_detail(request, challenge_host_team_pk, pk):
         return Response(response_data, status=status.HTTP_406_NOT_ACCEPTABLE)
 
     try:
-        challenge_host = ChallengeHost.objects.get(pk=pk)
+        challenge_host = get_challenge_host_model(pk)
     except ChallengeHost.DoesNotExist:
         response_data = {"error": "ChallengeHost does not exist"}
         return Response(response_data, status=status.HTTP_406_NOT_ACCEPTABLE)
```
nikochiko: @Kajol-Kumari You can replace the whole `try except` block with `challenge_host = get_challenge_host_model(pk)`.
See here: https://github.com/Cloud-CV/EvalAI/blob/1b1318d458c9dff4d2a0d936015e57215989ca21/apps/base/utils.py#L61-L64, the `get_model_object` already throws a `NotFound` (404). 
You need to change the code from `406` to `404` in tests as well.

apps/hosts/views.py
nikochiko: Please see the new comment below.

apps/hosts/views.py
Kajol-Kumari: @nikochiko when i replaced that try case with similar to this, I got "NotFound" undefined and still getting the same error in test_views , though i had changed the code. Btw the lines you are referring here is the definition of get_model_object and as far as i could understand it's output is being stored in "get_challenge_host_model" created by me in views.py.PLease correct me if i am missing something.

nikochiko: @Kajol-Kumari No, sorry if I wasn't clear. I didn't want you to use that code here.
What I meant was that, instead of the whole `try except` block:
```python
    try:
        challenge_host = get_challenge_host_model(pk)
    except ChallengeHost.DoesNotExist:
        response_data = {"error": "ChallengeHost does not exist"}
```
you can just do:
```python
challenge_host = get_challenge_host_model(pk)
```
I mentioned the code to show that when the function `get_model_object` is called, it itself gives a 404 error (by raising `rest_framework.NotFound`), and the except block is not reached.

Kajol-Kumari: My mistake, i understood something else. Thanks a lot for clarifying it @nikochiko

Kajol-Kumari:
@RishabhJain2018 Sir, made the required changes. Please take a look and if further changes are required, please let me know.

Kajol-Kumari:
@RishabhJain2018 please take a look at this PR.

PR #2673, opened 2020-02-14
Fixes Argoai image's incorrect directory 

yashdusing:
ArgoAI's logo is not loaded because of an incorrect path for the image. This PR fixes the issue


yashdusing:
@RishabhJain2018 sorry for the trouble. I didn't realise the mistake until I was too late.

PR #2674, opened 2020-02-15
Fix: Dockerised rl-worker

Ayukha:
- Added docker file for rl worker in dev and prod
- Updated docker-compose for dev, staging and prod environment


Ayukha:

PR #2684, opened 2020-02-18
Fix #2670: Add API for counting SQS messages in queue with unit tests

### LINKED ISSUE #2670: Create an API for getting the count of pending messages on in SQS Queue
**Description:**
Currently, we have the APIs for fetching & deleting the messages from SQS Queue but the challenge host doesn't know the pending messages in the AWS SQS queue.

**Deliverables**:
- [ ] Add an API for returning the count of the pending messages in SQS Queue.
- [ ] Add relevant tests for the API.

**References:**
- https://stackoverflow.com/questions/11901273/how-to-get-messages-receive-count-in-amazon-sqs-using-boto-library-in-python

yashdusing:
### Description
This PR adds the functionality to count pending SQS messages currently in queue.

### Deliverables
- [x] Add implementation for API used to count SQS messages
- [x] Add unit tests for SQS messages API (fetch and count) 
- [x] Add implementation function in remote submission function with tests cases for the same


yashdusing:

apps/base/utils.py
```diff
@@ -156,7 +156,7 @@ def get_boto3_client(resource, aws_keys):
 
 def get_or_create_sqs_queue_object(queue_name):
     if settings.DEBUG or settings.TEST:
-        queue_name = "evalai_submission_queue"
+        # queue_name = "evalai_submission_queue"
```
RishabhJain2018: Why do we need to comment this line?

apps/jobs/sender.py
```diff
@@ -29,7 +29,7 @@ def get_or_create_sqs_queue(queue_name):
             aws_access_key_id=os.environ.get("AWS_ACCESS_KEY_ID", "x"),
         )
         # Use default queue name in dev and test environment
-        queue_name = "evalai_submission_queue"
+        # queue_name = "evalai_submission_queue"
```
RishabhJain2018: Same as above

apps/jobs/urls.py
```diff
@@ -55,6 +55,11 @@
         views.get_submission_message_from_queue,
         name="get_submission_message_from_queue",
     ),
+    url(
+        r"^count/queues/(?P<queue_name>[\w-]+)/$",
+        views.get_total_messages_in_queue,
```
RishabhJain2018: `get_total_messages_in_queue` -> `get_message_count_from_queue`

apps/jobs/urls.py
```diff
@@ -55,6 +55,11 @@
...
+        name="get_total_messages_in_queue",
```
RishabhJain2018: `get_total_messages_in_queue` -> `get_message_count_from_queue`

apps/jobs/urls.py
```diff
@@ -55,6 +55,11 @@
         views.get_submission_message_from_queue,
         name="get_submission_message_from_queue",
     ),
+    url(
+        r"^count/queues/(?P<queue_name>[\w-]+)/$",
```
RishabhJain2018: `count/queues/(?P<queue_name>[\w-]+)/$` -> `messages/(?P<queue_name>[\w-]+)/count$`

apps/jobs/views.py
```diff
@@ -1361,6 +1361,79 @@ def delete_submission_message_from_queue(request, queue_name):
...
 
 
+@swagger_auto_schema(
+    methods=["get"],
+    manual_parameters=[
+        openapi.Parameter(
+            name="queue_name",
+            in_=openapi.IN_PATH,
+            type=openapi.TYPE_STRING,
+            description="Queue Name",
```
RishabhJain2018: Get count of the pending message in a queue.

apps/jobs/views.py
```diff
@@ -1361,6 +1361,79 @@ def delete_submission_message_from_queue(request, queue_name):
...
+        status.HTTP_400_BAD_REQUEST: openapi.Response(
+            "{'error': 'Error message goes here'}"
+        ),
+    },
+)
+@api_view(["GET"])
+@throttle_classes([UserRateThrottle])
+@permission_classes((permissions.IsAuthenticated, HasVerifiedEmail))
+@authentication_classes((ExpiringTokenAuthentication,))
+def get_total_messages_in_queue(request, queue_name):
```
RishabhJain2018: `get_total_messages_in_queue` -> `get_message_count_from_queue`

apps/jobs/views.py
```diff
@@ -1361,6 +1361,79 @@ def delete_submission_message_from_queue(request, queue_name):
...
+    """
+    API to fetch total number of messages in AWS SQS queue.
```
RishabhJain2018: `API to get the total number of pending messages in AWS SQS queue.`

apps/jobs/views.py
```diff
@@ -1361,6 +1361,79 @@ def delete_submission_message_from_queue(request, queue_name):
...
+
+    - Arguments:
+        ``queue_name``: AWS SQS queue name
+
+    - Returns:
+        ``total messages``: The total messages a key-value pair in a dict
+    """
+    try:
+        challenge = Challenge.objects.get(queue=queue_name)  # noqa
```
RishabhJain2018: Please `get_challenge_model` function here.

apps/base/utils.py
yashdusing: On dev end, if we use same queue name for testing, both the submission worker and unit tests would monitor & poll the same queue for messages. So, this would cause problems because the results would be unexpected.

RishabhJain2018: I don't think this will create problems (unless forced) as while testing you will explicitly send the messages into the queue and then test it.
One case I can think of is that when you run tests and at the same time create submissions from UI and it will create conflicts. Other than this do you have any other cases in your mind?

yashdusing: If the submission worker is working in the background on the same queue, it will poll the message regardless if it’s from the UI or from the unit tests so I think it could fail. I’ll try uncommenting though since it affects only the dev environment

apps/jobs/views.py
yashdusing: This is the description for the parameters list. Do you mean that I should change the description to the API function ?

apps/base/utils.py
RishabhJain2018: This is what I wrote above, no?

yashdusing: Yes. But, since there are two consumers (the unit test & submission queue) for one producer (the unit test), the message can be picked up by any of the 2 causing unexpected results. This will happen only on dev env, since that's the only place we're overriding the queue name

apps/jobs/views.py
yashdusing: `get_challenge_model` uses pk whereas this function takes in queue name as argument. If you want I can change this to pk and get queue name indirectly

yashdusing:
@RishabhJain2018 Please review now

apps/jobs/views.py
RishabhJain2018: Oh..my bad. Thanks for correcting me.

RishabhJain2018:
Closing this since it is no longer needed.

Issue #2564, opened 2020-01-01
Insecure requirement: Django version insecure


nikochiko:
#### Current:
The Django version currently in use for EvalAI (1.11.23) is now being marked as insecure by `requires.io`. While 1.11.23 was a security release and is much safer than its previous versions, it can still be a major trouble in the future as Django has ended support for all 1.x versions.
![django-insecure](https://user-images.githubusercontent.com/37668193/71638325-ac852900-2c82-11ea-9f08-99a03c8265d4.PNG)

#### Proceeding:
The Django version should be updated to 3.0.x so that we won't have to make major changes urgently in the API when a new vulnerability is discovered in Django. 
However, this can be a big change, many dependencies will have to be updated to support the new version. Along with API updates for Django, the API for these dependencies will also have to be updated. 
I think the best way to proceed would be to:
1. Research and enumerate the required changes
2. Formulate a plan (e.g. which dependencies will have to be updated first and which should be kept for later)
3. Add modular changes in steps according to the plan (reference: #2505)

adrijshikhar:
@nikochiko is this issue open?
i would like to work on this.

nikochiko:
Yes, the issue is open!

But heads up, the maintainers haven't shared their opinions yet. And as it will be a big decision to bump django this is only a ["spike"](https://www.quora.com/What-is-a-spike-in-software-engineering) task for now.

RishabhJain2018:
@adrijshikhar Would you still like to work on it?

devanshbatra04:
Can I take this up?

Suryansh5545:
@devanshbatra04 I don't think @adrijshikhar or anyone else is working on this so you can take it. But it will be a lot of work. :+1:

RishabhJain2018:
@devanshbatra04 Sure!

Suryansh5545:
@Rishabhjain2018 I think this issue is free as @devanshbatra04 is working on other issue #2686 . So can I take it?

devanshbatra04:
@Suryansh5545 please go ahead, I have my exams so won't be able to work on either issues this week.

apoorvkhare07:
Hey @nikochiko @RishabhJain2018 I think porting complete backend from Django 1.11 to 2.2 is in itself a big task and would take time. So can we break it small tasks so that multiple people would be able to work on it?

nikochiko:
@apoorvkhare07 
[...]

PR #2686, opened 2020-02-21
[WIP] Fix #1347 : Invite to host team via email

### LINKED ISSUE #1347: Add feature to invite users to host team using URL
Very similar to #1346. 

## Deliverables

Add a feature to invite new team member to a participant team using a Unique URL. Also, don't forget to write unit tests. :) 

## Steps

1. **user1** enters the email of his teammate (let's say **user2**)
2. Send an email to **user2** (if he/she is registered on EvalAI) with a unique URL
3. **user2** clicks on the URL to accept the invitation to the team.
4. *user1* gets an email saying that "**user2** has accepted your invite to team <blah>" (we can iterate over the email message later) 

### Some important points to note

- [ ] Checks performed before inviting to the team should be similar to current invitation logic

devanshbatra04:
Closes #1347 
Still a lot of work to be done - need to add code to actually send the email and all tests.

### DIFF
--- a/apps/hosts/serializers.py
+++ b/apps/hosts/serializers.py
@@ -59,8 +59,6 @@ class InviteHostToTeamSerializer(serializers.Serializer):
             self.user = context.get("request").user
 
     def validate_email(self, value):
-        if value == self.user.email:
-            raise serializers.ValidationError("A host cannot invite himself")
         try:
             User.objects.get(email=value)
         except User.DoesNotExist:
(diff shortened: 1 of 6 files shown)


Suryansh5545:
@devanshbatra04 Thx for the pr. But you should have asked the maintainers before starting to work on it . Also as I can see you are already assigned a issue #2564. And it is not allowed to work on two issues at a same time.

devanshbatra04:
> But you should have asked the maintainers before starting to work on it

I did, but didn't get a reply. Will keep in mind next time.

> it is not allowed to work on two issues at a same time.

Oh okay. Didn't know about that rule. Will keep the other issue on hold until this gets completed.

PR #2687, opened 2020-02-22
Fix #2685 Updated Python to 3.7.5

### LINKED ISSUE #2685: Test and Verify django, worker container with python 3.7.5
**Description:**

We would like to update the python dependency from 3.6.5  to 3.7.5 and it requires testing of django and worker development container with python3.7.5

**Deliverables:**
- [ ] Test the Django docker container and if it works fine, then update the dev Dockerfile with python3.7.5 as the base image.
- [ ] Test the worker docker container and if it works fine, then update the dev Dockerfile with python3.7.5 as the base image.

Ayukha:
- [x] Updated Django docker container dev Dockerfile with python3.7.5 as the base image.

- [x] Updated worker docker container dev Dockerfile with python3.7.5 as the base image.

- [x] Updated celery container dev Dockerfile with python3.7.5 as the base image
(closes #2685 )

### DIFF
--- a/docker/dev/celery/Dockerfile
+++ b/docker/dev/celery/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.6.5
+FROM python:3.7.5
 
 ENV PYTHONUNBUFFERED 1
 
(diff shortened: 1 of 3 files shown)


Ayukha:
@RishabhJain2018  I tested it by building the containers but didn't disable docker cache, I'll just do it.

Ayukha:
worker

@RishabhJain2018 If works fine after disabling the docker cache.

PR #2714, opened 2020-03-10
functional image avatar upload & UI

MarchingVoxels:
Image avatar uploading pipeline
The following is the update-profile page
image_avatar_UI
Display after update
image_ui_display


little0707:
There’s an underlying issue with the code and I can fix that

PR #2721, opened 2020-03-22
[Frontend] Update challenge page UI

deshraj:
Updated UI Screencast: https://www.loom.com/share/402f5596e9f647019baf43eb5776a169

Some of the screenshots of the updated UI:

screencapture-localhost-8888-web-challenges-challenge-page-1-evaluation-2020-03-22-04_04_41

screencapture-localhost-8888-web-challenges-challenge-page-1-leaderboard-1-2020-03-22-04_05_29

screencapture-localhost-8888-web-challenges-challenge-page-1-my-challenge-all-submission-2020-03-22-04_05_17

screencapture-localhost-8888-web-challenges-challenge-page-1-my-submission-2020-03-22-04_05_05

screencapture-localhost-8888-web-challenges-challenge-page-1-overview-2020-03-22-04_04_26

screencapture-localhost-8888-web-challenges-challenge-page-1-phases-2020-03-22-04_04_48

screencapture-localhost-8888-web-challenges-challenge-page-1-submission-2020-03-22-04_04_55

### DIFF
--- a/frontend/src/views/web/challenge/my-submission.html
+++ b/frontend/src/views/web/challenge/my-submission.html
@@ -1,163 +1,165 @@
-<section class="ev-sm-container ev-view ">
-    <div class="row">
-        <div class="col s12 m12">
-            <div class=" ev-card-panel ev-z-depth-5 ">
-                <div class="loader-container card-loader-container" id="loader" ng-show="challenge.isExistLoader" class="fade">
-                    <div class="loader">
-                        <div></div>
-                        <div></div>
-                        <div></div>
-                        <div></div>
-                        <div></div>
-                    </div>
+<section class="ev-sm-container ev-view challenge-container">
+    <div class="ev-md-container ev-card-panel ev-z-depth-5 ev-challenge-view">
+        <div class="row margin-bottom-cancel">
+            <div class="col s12">
+                <h5 class="w-300">My Submissions</h5>
+            </div>
+        </div>
+        <div class="row">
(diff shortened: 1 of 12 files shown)


frontend/src/views/web/challenge/my-submission.html
```diff
@@ -1,163 +1,165 @@
...
-            <div class=" ev-card-panel ev-z-depth-5 ">
-                <div class="loader-container card-loader-container" id="loader" ng-show="challenge.isExistLoader" class="fade">
-                    <div class="loader">
-                        <div></div>
-                        <div></div>
-                        <div></div>
-                        <div></div>
-                        <div></div>
-                    </div>
+<section class="ev-sm-container ev-view challenge-container">
```
RishabhJain2018: @deshraj I think it would be good if we should reorder the fields under `my-submissions` Tab.
I think it should be -
`#, Team, Status, Submitted File, Result File, Stdout File, Stderr File, Method Name, Method Description, Project URL, Publication URL, Execution Time, Show on Leaderboard, Baseline,  Submitted At, `

The reason is since most participants don't fill the details for method name etc, showing these fields with None value is not good. Instead, we can use the space for which participants have to do the horizontal scroll every time such as result or stderr field. What do you think?

deshraj: Let's keep it as it is for now. I have a better idea in mind to use the collapsible (https://materializecss.com/collapsible.html). Then it will look nicer.

RishabhJain2018: yep, that makes sense.

Issue #2693, opened 2020-03-01
Irrelevant dependencies installed in docker images


Ayukha:
The docker files contain dependencies related to Django which are not used by the Workers and Celery. The current size of the docker images can be reduced to a large amount if we remove the irrelevant requirements.

- [ ] Separate requirement files for worker and remote worker

- [ ]  Seperate dockerfile for remote worker

- [ ] remove irrelevant dependencies from dockerfile

@RishabhJain2018

RishabhJain2018:
Thanks @Ayukha for pointing it out. Please feel free to work on it.

RishabhJain2018:
This issue is still open :)

pystar-oss:
@RishabhJain2018  I would like to work on it

RishabhJain2018:
Sure @pystar-oss you can take this up.

adityaketkar:
@pystar-oss if you are still working on this issue? if not, @RishabhJain2018 can I take this up?

RishabhJain2018:
This is fixed, hence closing it.

PR #2725, opened 2020-03-23
[Docker] Improve docker build setup

deshraj:
Right now, we have duplicate steps in the celery and worker image that we already cover in the django. So, adding a new multistage build for celery and worker image which imports from the built django image rather than using `python:3.6.5`

### DIFF
--- a/docker-compose-production.yml
+++ b/docker-compose-production.yml
@@ -1,6 +1,5 @@
 version: "3"
 services:
-
   django:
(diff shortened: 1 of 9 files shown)


Ayukha:
@deshraj @RishabhJain2018 Using evalai_django as the base image for celery and worker is installing many irrelevant dependencies in those two containers and making the container heavy(memory). As mentioned #2693 

image

deshraj:
@Ayukha agree. That is a separate issue I think. I was mostly trying to avoid the execution of same set of commands twice in this PR. 

Please feel free to open a PR for that and I am happy to review it. :)

PR #2729, opened 2020-03-25
Integrate API for bearer token and cluster name with RL worker

RishabhJain2018:
This PR -
- [x] Integrates API for fetching AWS EKS bearer token with RL worker
- [x] Integrates API for fetching cluster details with RL worker

### DIFF
--- a/scripts/workers/rl_submission_worker.py
+++ b/scripts/workers/rl_submission_worker.py
@@ -23,7 +23,6 @@ class GracefulKiller:
 
 logger = logging.getLogger(__name__)
 config.load_kube_config()
-batch_v1 = client.BatchV1Api()
 
 AUTH_TOKEN = os.environ.get("AUTH_TOKEN", "auth_token")
 EVALAI_API_SERVER = os.environ.get(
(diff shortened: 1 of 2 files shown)


scripts/workers/rl_submission_worker.py
```diff
@@ -148,6 +147,17 @@ def process_submission_callback(body, challenge_phase, evalai):
...
 
 
+def get_api_object(cluster_name, challenge, evalai):
+    configuration = client.Configuration()
+    aws_eks_api = evalai.get_aws_eks_bearer_token(challenge.get("id"))
+    configuration.api_key["authorization"] = aws_eks_api[
+        "aws_eks_bearer_token"
+    ]
+    configuration.api_key_prefix["authorization"] = "Bearer"
+    api_instance = client.BatchV1Api(client.ApiClient(configuration))
```
yashdusing: @RishabhJain2018 if we're going to be using Bearer tokens for authentication then what is the need for the kubeconfig field in the challenge model (one of the recent PRs has it). Is it to create the CRUD API for cluster deliverable so that we can take kubeconfig file as input and just use that if challenge host uploads it?

pushkalkatara: I think for now it is to communicate between the host cluster running the worker node and the cluster running the EvalAI API. [Ref](https://jamesdefabia.github.io/docs/user-guide/sharing-clusters/)
In later implementations, can be used in cluster CRUD API for host.

PR #2730, opened 2020-03-25
add class to make anchor links appear in blue

RishabhJain2018:
This PR fixes minor issues from #2721 
- [x] Adds class on UI to make the anchor links visible in blue
cc: @deshraj


deshraj:
Oh, thats a good catch. Totally missed it. Thanks!

PR #2732, opened 2020-03-27
Fix #2691: Updated Python Version to 3.7.5 in .travis/yml

### LINKED ISSUE #2691: Update Python version to latest across all environments
As we have updated the python version in the dev environment to 3.7.5, I think we should also update the python version used by Travis for CI, currently, it uses python 3.6
https://github.com/Cloud-CV/EvalAI/blob/1b1318d458c9dff4d2a0d936015e57215989ca21/.travis.yml#L6

@RishabhJain2018

Rubix982:
Modified the Python version at line number 6 from '3.6' to '3.7.5' as per issue #2691 [Update Python version in .travis.yml #2691](https://github.com/Cloud-CV/EvalAI/edit/master/.travis.yml).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ sudo: false
 services:
 - docker
 python:
-- '3.6'
+- '3.7.5'
 addons:
   postgresql: '10'
   apt:


.travis.yml
```diff
@@ -3,7 +3,7 @@ sudo: false
 services:
 - docker
 python:
-- '3.6'
+- '3.7.5'
```
deshraj, line 6: Ideally we want to upgrade the python version across the project if we decide to bump up in travis for the sake of consistency. 

Can you please upgrade in the docker requirements and see if it works? I remember running into some issue when upgrading the docker python to 3.7.5

Rubix982: Hi! I just received a notification an hour ago because of the bot. I did not know you commented here. Let me check that out.

Rubix982: I just pulled the master branch with the latest commit, and `docker-compose --build` does not work. I am thrown this error,
```shell
Step 2/6 : RUN apt-get update &&   apt-get install -q -y default-jdk
 ---> Running in 379e2b43bde4
ERROR: Service 'worker' failed to build: failed to create endpoint elated_tharp on network bridge: failed to add the host (veth6d1a8f2) <=> sandbox (veth7e6d7eb) pair interfaces: operation not supported
```

I'm assuming there is something wrong with the package `default-jdk`, but I'm not that professional with docker so I maybe wrong. Please let me know when the master branch builds again so I can test with the new version number for Python.

GauravSarkar:
@deshraj what is the present state of this issue. I can see python version has been updated but the docker files have not been updated. So if it is still open, can I work on it.

PR #2756, opened 2020-04-19
Fixed pagination error for submissions

MarchingVoxels:
when the submissions reach > 100 then on page 2 under "My Submissions" Tab, the numbering starts again from 1 but it should be 101. This commit fixes this issue. 
Before: 
submissions_unfixed

After: 
submission_fixed


frontend/src/views/web/challenge/my-challenge-all-submission.html
```diff
@@ -94,7 +94,7 @@ <h5 class="w-300">All Submissions</h5>
                     </thead>
                     <tbody>
                         <tr ng-repeat="key in challenge.submissionResult.results" class="result-val">
-                            <td>{{$index + 1}}</td>
+                            <td>{{$index + 1 + (challenge.currentRefPage - 1) * 100}}</td> <!--target info-->
```
RishabhJain2018, line 97: Please remove `<!--target info-->` and also add the screenshot of the interface with the changes before and after.

PR #2758, opened 2020-04-24
Added submission_file_type field and migration script

MarchingVoxels:
Added a submission_file_type attribute for ChallengePhase model and the corresponding migration script.


apps/challenges/models.py
```diff
@@ -237,6 +237,10 @@ def __init__(self, *args, **kwargs):
     environment_image = models.CharField(
         max_length=2128, null=True, blank=True
     )  # Max length of repository name and tag is 2000 and 128 respectively
+    submission_file_type = models.CharField(
```
RishabhJain2018, line 240: Can you please rename it to - `allowed_submission_file_types`

apps/challenges/models.py
```diff
@@ -237,6 +237,10 @@ def __init__(self, *args, **kwargs):
...
+        max_length=200,
+        default='.json, .zip, .txt, .tsv, .gz, .csv, .h5, .npy'
```
RishabhJain2018, line 242: `default='.json, .zip, .txt, .tsv, .gz, .csv, .h5, .npy'` -> `default=".json, .zip, .txt, .tsv, .gz, .csv, .h5, .npy"`

scripts/migration/set_phase_submission_file_type.py
```diff
@@ -0,0 +1,22 @@
+# To run the file:
```
RishabhJain2018, line 1: We won't need this file as we're already setting a default value.

PR #2762, opened 2020-04-26
Fix#2761: functionality to send slack notifications

### LINKED ISSUE #2761: Add functionality to send slack notifications
**Description:**
Some of the challenge hosts are curious to know when the submissions to their challenge are coming, hence they want to post on the slack channel whenever a new submission comes in.

**Deliverables:**
- [ ] Add a nullable with blank=True URL field in the Challenge model with the name `slack_webhook_url`.
- [ ] Add a migrations file for the same.
- [ ] Use the function from base/utils for sending slack notification.
- [ ] In the function `publish_submission_message` in `jobs/sender.py`, if there is a `slack_webhook_url` in the challenge model, then call the function in the above step with this URL after sending the message in the queue.
- [ ] Add required unit tests.

Ayukha:
- [X] Model field Added

- [X] Migrations added

- [X]  function in jobs/utils.py created

- [X] Function call done

- [ ] Unit tests

closes #2761 

slack


apps/challenges/migrations/0062_challenge_slack_webhook_url.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
RishabhJain2018, line 1: Please rename this file to `0062_add_slack_webhook_url_field_in_challenge_model.py`.

apps/challenges/models.py
```diff
@@ -127,7 +127,9 @@ def __init__(self, *args, **kwargs):
     task_def_arn = models.CharField(
         null=True, blank=True, max_length=2048, default=""
     )
-
+    slack_webhook_url = models.URLField(
+        max_length=100, blank=True, null=True
```
RishabhJain2018, line 131: Can we please change this `max_length` to 200.

apps/challenges/models.py
```diff
@@ -127,7 +127,9 @@ def __init__(self, *args, **kwargs):
...
+    )
     class Meta:
```
RishabhJain2018, line 133: Please add a blank line before this class.

apps/jobs/sender.py
```diff
@@ -76,6 +77,14 @@ def publish_submission_message(message):
         )
         return
     queue_name = challenge.queue
+    slack_url = challenge.slack_webhook_url
     queue = get_or_create_sqs_queue(queue_name)
     response = queue.send_message(MessageBody=json.dumps(message))
+    # sending slack notification to the host
```
RishabhJain2018, line 83: `# sending slack notification to the host` -> `# send slack notification`

apps/jobs/utils.py
```diff
@@ -220,3 +221,21 @@ def handle_submission_rerun(submission, updated_status):
                 "submitted_image_uri"
             ]
     return message
+
+
+def notify_on_slack(slack_url, slack_data):
```
RishabhJain2018, line 226: @Ayukha I just realized that we already had this function written in base/utils.py. Can you please use that -> https://github.com/Cloud-CV/EvalAI/blob/master/apps/base/utils.py#L207

apps/jobs/sender.py
```diff
@@ -76,6 +77,14 @@ def publish_submission_message(message):
...
+    if slack_url:
+        data = {
```
RishabhJain2018, line 85: message = {
                    "text": "A *new submission* is uploaded to {}".format(challenge_name),
                    "fields": [
                        {
                            "title": "Challenge Phase",
                            "value": phase_name,
                            "short": True,
                        },
                        {
                            "title": "Participant Team Name",
                            "value": participant_team_name,
                            "short": True,
                        },
                    ],
                }

RishabhJain2018: Please change data to the message given above and get the fields `challenge_name`, `phase_name` and `participant_team_name` using the submission pk. Make sure to query the database only when the slack URL is present.

RishabhJain2018: Also, can you please add this icon: https://evalai.cloudcv.org/dist/images/evalai-logo-single.png instead of a default icon for the webhook.

apps/jobs/sender.py
```diff
@@ -72,10 +74,39 @@ def publish_submission_message(message):
...
         )
         return
     queue_name = challenge.queue
+    slack_url = challenge.slack_webhook_url
     queue = get_or_create_sqs_queue(queue_name)
     response = queue.send_message(MessageBody=json.dumps(message))
+    # sending slack notification
+    if slack_url:
+        challenge_name = challenge.title
+        phase_name = ChallengePhase.objects.get(pk=message["phase_pk"]).name
```
RishabhJain2018, line 89: @Ayukha you can get the `phase_name` from the Submission object itself. There is no need for this query.

apps/jobs/sender.py
Ayukha: img

@RishabhJain2018 Did you mean this  image? This can be changed when installing the integration on slack channel.

RishabhJain2018: yes, it can be changed when installing the integration on slack but I think it would be good to automate it from our code as it will reduce the steps to ask the challenge host to do it.

They can simply add the integration and give us the URL to post to and rest we should be able to take care of it.

Ayukha:
img

> I'm curious since we have flake8 and black setup as pre-commit hook, how are you able to push?

I didn't had pre-commit hooks setup, should we add it in the contribution guidelines?

RishabhJain2018:
> I didn't had pre-commit hooks setup, should we add it in the contribution guidelines?
I see..yes we should add it in contribution guidelines. Please add it in this issue only #2766.

apps/jobs/sender.py
```diff
@@ -72,10 +74,40 @@ def publish_submission_message(message):
...
+                message["challenge_pk"]
+            )
         )
         return
     queue_name = challenge.queue
+    slack_url = challenge.slack_webhook_url
     queue = get_or_create_sqs_queue(queue_name)
     response = queue.send_message(MessageBody=json.dumps(message))
+
+    # sending slack notification
```
RishabhJain2018, line 87: `# sending slack notification` -> `# send slack notification`

apps/jobs/sender.py
```diff
@@ -72,10 +74,40 @@ def publish_submission_message(message):
...
+    if slack_url:
+
```
RishabhJain2018, line 89: Please remove this blank line.

apps/jobs/sender.py
```diff
@@ -72,10 +74,40 @@ def publish_submission_message(message):
...
+        challenge_name = challenge.title
+        submission = Submission.objects.get(pk=message["phase_pk"])
```
RishabhJain2018, line 91: I think `phase_pk` is wrong here. It should be `message["submission_pk"]`.

apps/jobs/sender.py
```diff
@@ -72,10 +74,40 @@ def publish_submission_message(message):
...
+                },
+                {
+                    "title": "Participant Team Name",
+                    "value": participant_team_name,
+                    "short": True,
+                },
+            ],
+        }
+        send_slack_notification(slack_url, message)
+
```
RishabhJain2018, line 112: Please remove this blank line.

apps/base/utils.py
```diff
@@ -215,6 +215,8 @@ def send_slack_notification(webhook=settings.SLACK_WEB_HOOK_URL, message=""):
         data = {
             "text": message["text"],
             "attachments": [{"color": "ffaf4b", "fields": message["fields"]}],
+            "username": "EvalAI",
```
RishabhJain2018, line 218: Can you please re-organize all these keys alphabetically.

apps/jobs/sender.py
RishabhJain2018: Also, can you please use `get_submission_model` function here.
Ref: https://github.com/Cloud-CV/EvalAI/blob/master/apps/jobs/views.py#L910

PR #2763, opened 2020-04-26
Added appropriate error message for team creation

MarchingVoxels:
Added appropriate error messages for team creation: 
1. when there are illegal characters in team name
2. when the team name is blank

Before: 
team_before

After: 
For blank name: 
team_after1
For illegal name(with a single dot)
team_after2


frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -67,8 +67,8 @@
 
         vm.subErrors = {};
 
-        vm.isChallengeLeaderboardPrivate = false;
 
+        vm.isChallengeLeaderboardPrivate = false;
```
RishabhJain2018, line 71: Please don't make this change in this PR.

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -1142,7 +1143,13 @@
                     var error = response.data;
                     vm.team.error = error.team_name[0];
                     vm.stopLoader();
-                    $rootScope.notify("error", "New team couldn't be created.");
+                    if (vm.team.error == "This field is required.") {
+                        $rootScope.notify("error", "Illegal character in team name.");
```
RishabhJain2018, line 1147: Please also show which character is illegal in the name entered by the participant.

`Illegal character in team name.` -> It is way too generic.

PR #2768, opened 2020-04-28
Fix #2767: Adds ignore_submission field in submission model

### LINKED ISSUE #2767: Add models to disable/delete submissions
**Motivation**:
Some of the challenge hosts want to delete the 'canceled` submissions and the submission in which participants have uploaded non-standard submission files from their All Submissions view. Hence, we require this feature.

**Deliverables**:
- [ ] Add a boolean field `ignore_submission` with default as False in the Submissions model.
- [ ] Add migration file for the added models.

Ayukha:
- [X] Add a boolean field is_disabled with default as False in the Submissions model.
- [X] Add migration file for the added models (closes #2767)


Ayukha:

Issue #2783, opened 2020-05-06
Unordered list HTML tag <ul></ul> isn't getting rendered in evaluation details section


RishabhJain2018:
**Description**:
In the evaluation details section if a challenge host wants to render a `<ul></ul>` tag then it isn't getting rendered properly.

RishabhJain2018:
@Kajol-Kumari Can you please look into this?

Kajol-Kumari:
sure @RishabhJain2018 sir.

Kajol-Kumari:
@RishabhJain2018 @Sanji515 This issue is not only with `evaluation_detail_edit` but all the editing trix editors where angular-trix has been used i.e in `submission_guidelines_Edit`, `Terms_And_condition_Edit` etc. The trix-editor is working fine when i pasted it on plunker. So i thought maybe some .scss design is overwriting it. But even after removing all the styles from it. I am still just getting the blank space instead of bullet. Still can't figure out what is causing this problem. What am i missing?

RishabhJain2018:
okay.
Did you check by inputting the `<ul></ul>` manually in the HTML of overview, etc? I think it also doesn't get rendered from there.

Kajol-Kumari:
> okay.
> Did you check by inputting the `<ul></ul>` manually in the HTML of overview, etc? I think it also doesn't get rendered from there.

Yes i had tried this too. In this case also only blank space is appearing instead of bullets.

RishabhJain2018:
I don't think you're missing anything and it is mostly the CSS override issue. Did you check in the console by removing the CSS to this `<ul></ul>` element?

PR #2784, opened 2020-05-07
Fix:#2783 Bullet list in Trix editor

Kajol-Kumari:
Fix: #2783 

**Cause of Bug:**

CSS override in particular here `ul:not(.browser-default)` was being set with `list-style-type: none;` and it was overriding the bullet-list

**GIF of Fix:**
ezgif com-video-to-gif (2)


Kajol-Kumari:
@RishabhJain2018 please look at it. And if any changes are required, please let me know.

Issue #2785, opened 2020-05-07
Create unique URL for each phase under My Submission and All Submissions Page


RishabhJain2018:
**Description**:
If a challenge has multiple phases and the host reloads the My submission page / All Submissions page then the selected phase changes to blank and it becomes inconvenient for the host to select the phase again from the list again.

**Fix**:
Create a unique link for each phase using the slug for the phase in the URL.

RishabhJain2018:
@Kajol-Kumari Can you please add this :)

Kajol-Kumari:
> @Kajol-Kumari Can you please add this :)

Sure @RishabhJain2018 sir.

PR #2786, opened 2020-05-09
[#2785] Create unique URL for each phase under My Submission and All Submissions Page 

Kajol-Kumari:
fix: #2785 

- [x]  Added unique url using slug for `All_submissions` of a particular phase.

- [ ] Add unique url for get submissions under `my_submissions` tab

- [ ] Add the required tests.

GIF for `all_submissions` url:
ezgif com-video-to-gif (2)

### DIFF
--- a/apps/challenges/models.py
+++ b/apps/challenges/models.py
@@ -15,7 +15,7 @@ from base.models import (
     model_field_name,
     create_post_model_field,
 )
-from base.utils import RandomFileName, get_slug
+from base.utils import RandomFileName, get_slug, unique_slug_generator
 from participants.models import ParticipantTeam
 from hosts.models import ChallengeHost
 
@@ -444,3 +444,14 @@ class ChallengeEvaluationCluster(TimeStampedModel):
     class Meta:
         app_label = "challenges"
         db_table = "challenge_evaluation_cluster"
+
+# def pre_save_post_receiver(sender, instance, *args, **kwargs):
+#     if not instance.slug:
+#         # instance.slug = get_slug(instance)
+#         instance.slug = unique_slug_generator(instance)
+
+# pre_save.connect(pre_save_post_receiver, sender=ChallengePhase)
+@receiver(pre_save, sender="challenges.ChallengePhase")
+def submission_challenge_slug(sender, instance, **kwargs):
+    title = get_slug(instance.name)
+    instance.slug = "{}-{}".format(title, instance.pk)
--- a/apps/challenges/views.py
+++ b/apps/challenges/views.py
@@ -2148,6 +2148,87 @@ def get_challenge_phase_by_pk(request, pk):
 
 
 @api_view(["GET"])
+@throttle_classes([UserRateThrottle])
+@permission_classes((permissions.IsAuthenticated, HasVerifiedEmail))
+@authentication_classes((ExpiringTokenAuthentication,))
+def get_all_submissions_of_challenge_by_slug(
+    request, challenge_pk, slug
+):
+    """
+    Returns all the submissions for a particular challenge
+    """
+    # To check for the corresponding challenge from challenge_pk.
+    challenge = get_challenge_model(challenge_pk)
+    print(challenge)
+
+    # To check for the corresponding challenge phase from the slug and challenge.
+    try:
+        challenge_phase = ChallengePhase.objects.get(
+            slug=slug, challenge=challenge
+        )
+        print(challenge_phase)
+    except ChallengePhase.DoesNotExist:
+        response_data = {
+            "error": "Challenge Phase {} does not exist".format(
+                slug
+            )
+        }
+        return Response(response_data, status=status.HTTP_404_NOT_FOUND)
+
+    # To check for the user as a host of the challenge from the request and challenge_pk.
+    if is_user_a_host_of_challenge(
+        user=request.user, challenge_pk=challenge_pk
+    ):
+
+        # Filter submissions on the basis of challenge for host for now. Later on, the support for query
+        # parameters like challenge phase, date is to be added.
+        submissions = Submission.objects.filter(
+            challenge_phase=challenge_phase
+        ).order_by("-submitted_at")
+        filtered_submissions = SubmissionFilter(
+            request.GET, queryset=submissions
+        )
+        paginator, result_page = paginated_queryset(
+            filtered_submissions.qs, request
+        )
+        serializer = ChallengeSubmissionManagementSerializer(
+            result_page, many=True, context={"request": request}
+        )
+        print(serializer.data)
+        response_data = serializer.data
+        return paginator.get_paginated_response(response_data)
+
+    # To check for the user as a participant of the challenge from the request and challenge_pk.
+    elif has_user_participated_in_challenge(
+        user=request.user, challenge_id=challenge_pk
+    ):
+
+        # get participant team object for the user for a particular challenge.
+        participant_team_pk = get_participant_team_id_of_user_for_a_challenge(
+            request.user, challenge_pk
+        )
+
+        # Filter submissions on the basis of challenge phase for a participant.
+        submissions = Submission.objects.filter(
+            participant_team=participant_team_pk,
+            challenge_phase=challenge_phase,
+        ).order_by("-submitted_at")
+        paginator, result_page = paginated_queryset(submissions, request)
+        serializer = SubmissionSerializer(
+            result_page, many=True, context={"request": request}
+        )
+        response_data = serializer.data
+        return paginator.get_paginated_response(response_data)
+
+    # when user is neither host not participant of the challenge.
+    else:
+        response_data = {
+            "error": "You are neither host nor participant of the challenge!"
+        }
+        return Response(response_data, status=status.HTTP_400_BAD_REQUEST)
+
+
+@api_view(["GET"])
 @throttle_classes([AnonRateThrottle])
 def get_challenge_phase_by_slug(request, slug):
     """
--- a/apps/base/utils.py
+++ b/apps/base/utils.py
@@ -8,11 +8,14 @@ import re
 import requests
 import sendgrid
 import uuid
+import random
+import string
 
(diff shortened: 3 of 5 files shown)


Kajol-Kumari:
@RishabhJain2018 please let me know if it's OK so that i can go for `my_submission` url change.

Sanji515:
@Kajol-Kumari can you please fix the build here?

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -939,10 +942,13 @@
                     vm.stopLoader();
                 }
             };
-
             utilities.sendRequest(parameters);
         };
 
+        if(vm.mySubmissionPhaseSlug) {
```
Sanji515, line 948: Hey @Kajol-Kumari The problem is here, actually this thing is getting called before [this](https://github.com/Cloud-CV/EvalAI/blob/master/frontend/src/js/controllers/challengeCtrl.js#L505) API but in order to work as we are using `vm.phases` inside `vm.getResults` function so the above mentioned API should call first and then `vm.getResults` function should be called.

Kajol-Kumari: But @Sanji515 in the network tab i can see that [this API](https://github.com/Cloud-CV/EvalAI/blob/master/frontend/src/js/controllers/challengeCtrl.js#L504) being hit first and after that it's throwing that error of length of undefined.

Attached screenshort:
Screenshot from 2020-05-15 10-53-13

Sanji515: yes but after the execution of this(^) condition, that's reason of `vm.phases.results` is coming as `undefined`.

Kajol-Kumari: yes @Sanji515, Thanks for pointing this.I just restricted the call of `vm.getResult()` only after the value is assigned in vm.phases and now it works fine for me. Please check on your end once.

Kajol-Kumari:
@Sanji515 please look at it once and if any changes are required, please let me know.

apps/analytics/urls.py
```diff
@@ -24,7 +24,7 @@
         name="get_challenge_phase_submission_analysis",
     ),
     url(
-        r"^challenge/(?P<challenge_pk>[0-9]+)/challenge_phase/(?P<challenge_phase_pk>[0-9]+)/count$",
+        r"^challenge/(?P<challenge_pk>[0-9]+)/challenge_phase/(?P<version>(v1|v2|v3))/(?P<challenge_phase_pk>[-a-zA-Z0-9_]+)/count$",
```
Sanji515, line 27: @Kajol-Kumari Why there are 3 versions?

apps/analytics/views.py
```diff
@@ -141,14 +142,23 @@ def get_submission_count(request, challenge_pk, duration):
 )
 @authentication_classes((ExpiringTokenAuthentication,))
 def get_challenge_phase_submission_count_by_team(
-    request, challenge_pk, challenge_phase_pk
+    request, challenge_pk, challenge_phase_pk, version
```
Sanji515, line 145: As here it can be either slug or challenge phase id so can you please rename the variable to something like `challenge_phase_pk_or_slug`

apps/challenges/views.py
```diff
@@ -1308,7 +1308,7 @@ def create_challenge_using_zip_file(request, challenge_host_team_pk):
 @permission_classes((permissions.IsAuthenticated, HasVerifiedEmail))
 @authentication_classes((ExpiringTokenAuthentication,))
 def get_all_submissions_of_challenge(
-    request, challenge_pk, challenge_phase_pk
+    request, challenge_pk, challenge_phase_pk, version
```
Sanji515, line 1311: Same here.

apps/analytics/urls.py
Kajol-Kumari: 2 will be enough to serve our purpose.will remove the third one.

apps/challenges/models.py
```diff
@@ -444,3 +444,10 @@ class ChallengeEvaluationCluster(TimeStampedModel):
     class Meta:
         app_label = "challenges"
         db_table = "challenge_evaluation_cluster"
+
+
+def pre_save_post_receiver(sender, instance, *args, **kwargs):
```
Sanji515, line 449: Can you please tell why are you doing this? @Kajol-Kumari

apps/analytics/urls.py
```diff
@@ -24,7 +24,7 @@
         name="get_challenge_phase_submission_analysis",
     ),
     url(
-        r"^challenge/(?P<challenge_pk>[0-9]+)/challenge_phase/(?P<challenge_phase_pk>[0-9]+)/count$",
+        r"^challenge/(?P<challenge_pk>[0-9]+)/challenge_phase/(?P<version>(v1|v2))/(?P<challenge_phase_pk_or_slug>[-a-zA-Z0-9_]+)/count$",
```
Sanji515, line 27: Hey are we required to do this API also using slug in this PR?

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -792,25 +803,26 @@
...
-
+            
             vm.stopFetchingSubmissions = function() {
                 $interval.cancel(vm.poller);
             };
             vm.stopFetchingSubmissions();
             vm.isResult = true;
             if (phaseId !== undefined) {
                 vm.phaseId = phaseId;
+                vm.mySubmissionPhaseSlug = phaseId;
```
Sanji515, line 814: @Kajol-Kumari I think you have to work a liitle bit more here as now from the function `getResults`, its argument coming as slug not phase id in variable `phaseId` so you need to stop setting to controller variable `vm.phaseId` in line [813](https://github.com/Cloud-CV/EvalAI/pull/2786/files?file-filters%5B%5D=.js&file-filters%5B%5D=.py#diff-14a845646ed48757b5c7798b7ccdddb6R813) and also make sure that it is not breaking other functionalities where we might not able to get `vm.phaseId` after the removal of line 813.

apps/challenges/models.py
Kajol-Kumari: It is used to call the `unique_slug_generator` function just before the saving of the `challengePhaseSplits`in order to generated the slug based on the title of the challenge.

apps/analytics/urls.py
Kajol-Kumari: yes, as in `vm.getResults()` we are calling both submissions as well as count. So, both will be called using the slug only.

apps/challenges/models.py
Sanji515: But it is already done when the challenge gets created [here](https://github.com/Cloud-CV/EvalAI/blob/master/apps/challenges/views.py#L1013) then what's the need of it here.

Kajol-Kumari: Ah, i didn't noticed it. But yes, the `unique_slug_generator` i have used now gives preetier Url as the previously written one won't do much to an `id`

frontend/src/js/controllers/challengeCtrl.js
Kajol-Kumari: @Sanji515 we need to keep it as we will require it for download. Also as the submission extraction part is being done using slug. So, i think i need to change it for the `download_submission` also as now the `download_all_submission` will receive `slug` instead of the `id`

apps/analytics/views.py
```diff
@@ -141,14 +142,23 @@ def get_submission_count(request, challenge_pk, duration):
...
-    request, challenge_pk, challenge_phase_pk
+    request, challenge_pk, challenge_phase_pk_or_slug, version
 ):
     """
     Returns number of submissions done by a participant team in a challenge phase
     """
     challenge = get_challenge_model(challenge_pk)
 
-    challenge_phase = get_challenge_phase_model(challenge_phase_pk)
+    if version == 'v1':
```
Sanji515, line 152: Hey @Kajol-Kumari Can we change the slug based version to `v2` here and similar other places as it is the newer version of the API.

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -33,6 +35,7 @@
         vm.submissionVisibility = {};
         vm.baselineStatus = {};
         vm.showUpdate = false;
+        vm.showallSubmissionUpdate = false;
```
Sanji515, line 38: I think we are not using this variable, can we remove this?

apps/challenges/views.py
```diff
@@ -1336,17 +1336,30 @@ def get_all_submissions_of_challenge(
...
+        except ChallengePhase.DoesNotExist:
+            response_data = {
+                "error": "Challenge Phase {} does not exist".format(
+                    challenge_phase_pk_or_slug
+                )
+            }
+            return Response(response_data, status=status.HTTP_404_NOT_FOUND)
+    else:
+        try:
+            challenge_phase = ChallengePhase.objects.get(
```
Ram81, line 1353: @Kajol-Kumari you can use `get_challenge_phase_model` here right?

apps/challenges/views.py
```diff
@@ -1405,25 +1418,37 @@ def get_all_submissions_of_challenge(
...
```
Ram81, line 1441: same here?

apps/jobs/views.py
```diff
@@ -130,13 +130,22 @@ def challenge_submission(request, challenge_id, challenge_phase_id):
...
+        try:
+            challenge_phase = ChallengePhase.objects.get(
+                slug=challenge_phase_pk_or_slug, challenge=challenge
+            )
+        except ChallengePhase.DoesNotExist:
+            response_data = {"error": "Challenge Phase does not exist"}
+            return Response(response_data, status=status.HTTP_400_BAD_REQUEST)
+    else:
+        try:
+            challenge_phase = ChallengePhase.objects.get(
```
Ram81, line 143: same here?

RishabhJain2018:
@Ram81 Can you please review this.

PR #2791, opened 2020-05-12
Fix frontend issues

RishabhJain2018:
screencapture-127-0-0-1-8888-2020-05-12-00_18_56


deshraj:
Thanks for the fix. 🙏

PR #2792, opened 2020-05-13
Bold field when used to order leaderboard

MarchingVoxels:
Make the field with which the leaderboard is sorted is now bold
Before: 
leaderboard_before
After: 
leaderboardorder_after


MarchingVoxels:
> Hey @MarchingVoxels Can we please not use inline CSS.

PR #2802, opened 2020-05-19
[Frontend] Improve notifications layout and colors

deshraj:
## Current view
Screenshot from 2020-05-18 23-33-04

## Proposed view
Screenshot from 2020-05-18 23-31-51

Also added a bunch of typo fixes in the text shown in the notification.

### DIFF
--- a/frontend/src/js/controllers/challengeCtrl.js
+++ b/frontend/src/js/controllers/challengeCtrl.js
@@ -367,9 +367,9 @@
             // ev.stopPropagation();
             var participationState;
             if (isRegistrationOpen) {
-                participationState = 'Close';
+                participationState = 'closed';
             } else {
-                participationState = 'Open';
+                participationState = 'opened';
             }
             var confirm = $mdDialog.confirm()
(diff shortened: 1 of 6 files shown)


frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -367,9 +367,9 @@
             // ev.stopPropagation();
             var participationState;
             if (isRegistrationOpen) {
-                participationState = 'Close';
+                participationState = 'closed';
             } else {
-                participationState = 'Open';
+                participationState = 'opened';
```
Sanji515, line 372: @deshraj If you change this(^) text here then you also need to take care of the text which is shown over the confirmation modal [here](https://github.com/Cloud-CV/EvalAI/pull/2802/files#diff-14a845646ed48757b5c7798b7ccdddb6R375). Cause it will be incorrect to ask `opened/closed participation in the challenge?`

deshraj: Good point. Totally missed it. Incorporating now. Thanks!

PR #2803, opened 2020-05-19
[Feature] Restrict selection of one public submission for a challenge phase

Ram81:
### Feature Description

This feature is for the challenge phases with `is_restricted_to_one_submission=True`. It allows user to choose which submission will be shown on the leaderboard. The PR handles the following cases.

1. If a submission is already selected and the participant wants to change it to some other then we will show a pop up for confirmation. 

2. While challenge creation if `is_submission_public` is `True` then `is_restricted_to_one_submission` should be `False` or vice-versa.

3. We’re not making the latest submission public from our end -- Every team has to manually choose it.

### DIFF
--- a/apps/challenges/views.py
+++ b/apps/challenges/views.py
@@ -796,7 +796,7 @@ def create_challenge_using_zip_file(request, challenge_host_team_pk):
             return Response(
                 response_data, status=status.HTTP_406_NOT_ACCEPTABLE
             )
-        if "is_submission_public" in data.keys() and "is_restricted_to_select_one_submission" in data.keys():
+        if data.get("is_submission_public") and data.get("is_restricted_to_select_one_submission"):
             if data["is_submission_public"] and data["is_restricted_to_select_one_submission"]:
                 message = (
                     "is_submission_public can't be 'True' for for challenge phase '{}'"
@@ -806,7 +806,7 @@ def create_challenge_using_zip_file(request, challenge_host_team_pk):
                 )
                 response_data = {"error": message}
                 return Response(
-                    response_data, status=status.HTTP_406_NOT_ACCEPTABLE
+                    response_data, status=status.HTTP_400_BAD_REQUEST
                 )
 
     # Check for challenge image in yaml file.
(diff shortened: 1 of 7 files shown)


apps/jobs/views.py
```diff
@@ -372,6 +372,19 @@ def change_submission_data_and_visibility(
         if is_public is True:
             when_made_public = datetime.datetime.now()
             request.data["when_made_public"] = when_made_public
+
+        submissions_already_public = Submission.objects.filter(
```
RishabhJain2018, line 376: Hey @Ram81 Why are we using filter here? I think get should work here because a participant can only one submission public and if they change to another then the previous submission should be made private which is one in number.

apps/challenges/views.py
```diff
@@ -796,6 +796,18 @@ def create_challenge_using_zip_file(request, challenge_host_team_pk):
             return Response(
                 response_data, status=status.HTTP_406_NOT_ACCEPTABLE
             )
+        if "is_submission_public" in data.keys() and "is_restricted_to_select_one_submission" in data.keys():
```
deshraj, line 799: Instead of checking if these keys exist, shouldn't we check their values? 
something like:

```python
if data.get('is_submission_public') and data.get('is_restricted_to_select_one_submission'):
     # do something
```

apps/challenges/views.py
```diff
@@ -796,6 +796,18 @@ def create_challenge_using_zip_file(request, challenge_host_team_pk):
...
+            if data["is_submission_public"] and data["is_restricted_to_select_one_submission"]:
+                message = (
+                    "is_submission_public can't be 'True' for for challenge phase '{}'"
+                    " with is_restricted_to_select_one_submission 'True'. "
+                    " Please change is_submission_public to 'False'"
+                    " then try again!".format(data["name"])
+                )
+                response_data = {"error": message}
+                return Response(
+                    response_data, status=status.HTTP_406_NOT_ACCEPTABLE
```
deshraj, line 809: Can we please return 400 response instead since it's a bad data input from the user?

apps/jobs/views.py
```diff
@@ -372,6 +372,19 @@ def change_submission_data_and_visibility(
...
+            is_public=True,
+            participant_team=participant_team,
+            challenge_phase=challenge_phase
+        )
+
+        if (challenge_phase.is_restricted_to_select_one_submission
```
deshraj, line 382: Can we please add a comment before this line so that we remember later why this check was added ?

PR #2808, opened 2020-05-22
(Debugging - WIP) Add feature to restart active challenge workers every 3 days.

KhalidRmb:
- Included a cron job which runs everyday, to filter challenges which were last restarted 3 days ago.
- Restart the workers for the above challenges.

**TODO (In another PR):**
- Include a new field in the Challenge model to represent a defective challenge as an additional filter for this corn job. (Instead of restarting every challenge periodically.)


apps/challenges/migrations/0065_added_last_started_at_field_to_challenges.py
```diff
@@ -0,0 +1,20 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('challenges', '0064_add_is_restricted_to_select_one_submission_field_in_phase_model'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='challenge',
+            name='last_started_at',
```
deshraj, line 17: Can we please change this field to `worker_last_started_at` to be more clear ?

apps/challenges/models.py
```diff
@@ -164,6 +167,11 @@ def is_active(self):
             return True
         return False
 
+    @property
+    def three_days_since_last_start(self):
```
deshraj, line 171: nit: can we please change the name to `n_days_since_last_worker_start`

apps/challenges/restart_workers_cron.py
```diff
@@ -0,0 +1,36 @@
...
+    count, failures = response["count"], response["failures"]
+
+    if(count == challenges.count()):
+        logger.info("All challenge workers restarted succesfully.")
+    else:
+        for fail in failures:
+            logger.warning("Worker restart for challenge {} failed. Error: {}".format(fail["challenge_pk"], fail["message"]))
+
+
+'''
```
deshraj, line 21: Lets remove this commented part if we are not using it.

apps/challenges/aws_utils.py
```diff
@@ -1,3 +1,4 @@
+from datetime import date
```
Ram81, line 1: @KhalidRmb can we put all the `from` imports after `import` statements

PR #2809, opened 2020-05-23
[Deployment] Setup continuous deployment of webserver on AWS instances

deshraj:
Right now, we have to manually ssh to the EC2 instance in order to deploy the new changes on the staging or production server. 

This PR introduces the change where:
- Uses Travis file encryption to encrypt the private key
- Decrypt the private key during the `before_install` stage so that it can be used later for deployment
- If a commit lands on staging or production branch, it automatically triggers the deployment on the EC2 machine


scripts/deployment/deploy.sh
```diff
@@ -38,7 +17,32 @@ if [ -z ${COMMIT_ID} ]; then
...
+    exit 0
+fi
+
 case $opt in
+        auto_deploy)
+            ssh ubuntu@${HOSTNAME} -i scripts/deployment/evalai.pem -o StrictHostKeyChecking=no \
+            "cd ~/Projects/evalai && \
+            export AWS_ACCOUNT_ID=${AWS_ACCOUNT_ID} && \
+            export COMMIT_ID=${COMMIT_ID} && \
+            eval $(aws ecr get-login --no-include-email) && \
```
RishabhJain2018, line 41: Let's use `aws_login;` here.

scripts/deployment/deploy.sh
```diff
@@ -38,7 +17,32 @@ if [ -z ${COMMIT_ID} ]; then
...
+    HOSTNAME="evalai.cloudcv.org"
+elif [[ ${env} == "staging" ]]; then
+    HOSTNAME="evalai-staging.cloudcv.org"
+else
+    echo "Skipping deployment since commit not on staging or production branch."
+    exit 0
+fi
+
 case $opt in
+        auto_deploy)
```
RishabhJain2018, line 36: Can we please add docs for this command below - https://github.com/Cloud-CV/EvalAI/pull/2809/files#diff-2237c118ae9fb2de885c8df4c5c5f1d4R141

.travis.yml
```diff
@@ -11,7 +11,7 @@ addons:
     packages:
     - postgresql-10
     - postgresql-client-10
-    - default-jre # Java11 is used for elasticmq
+    - default-jre
```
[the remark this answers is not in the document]
deshraj: Getting rid of all of this in the next PR where we will run tests in docker. We really don't need any of this anymore.

scripts/deployment/deploy.sh
deshraj: Yeah, already tried that but sadly cannot use a function call inside since this is a string that is passed in after ssh to the machine.

PR #2814, opened 2020-05-24
Create EKS Cluster and nodegroup

Ayukha:
- [x] post hook when challenge is approved
- [x] create cluster
- [x] create nodegroup
- [x] Hardcoded disc space


apps/challenges/aws_utils.py
```diff
@@ -71,6 +71,18 @@
     ),
 }
 
+EKS_CLUSTER_ROLE_ARN = os.environ.get(
```
deshraj, line 74: Let's move these two constants to `settings.common`

apps/challenges/aws_utils.py
```diff
@@ -697,3 +709,57 @@ def restart_workers_signal_callback(sender, instance, field_name, **kwargs):
                 instance.pk, field_name
             )
         )
+
+
+def create_eks_nodegroup(instance, cluster_name):
+    """
+    Creates a nodegroup when a EKS cluster is created by the EvalAI admin
```
deshraj, line 716: Can we please follow numpy docstring style as given here: https://numpydoc.readthedocs.io/en/latest/format.html ?

apps/challenges/aws_utils.py
```diff
@@ -697,3 +709,57 @@ def restart_workers_signal_callback(sender, instance, field_name, **kwargs):
...
+    """
+    nodegroup_name = "{}-nodegroup".format(instance.title.replace(" ", "-"))
+    client = get_boto3_client("eks", aws_keys)
+    try:
```
deshraj, line 720: Would be great if you can add a comment before this saying something like:

```python
# TODO: Move the hardcoded cluster configuration such as the instance_type, subnets, AMI to challenge configuration later.
```

apps/challenges/aws_utils.py
```diff
@@ -697,3 +709,57 @@ def restart_workers_signal_callback(sender, instance, field_name, **kwargs):
...
+    cluster = ""
+    cluster_name = "{}-cluster".format(instance.title.replace(" ", "-"))
+    if instance.approved_by_admin:
+        client = get_boto3_client("eks", aws_keys)
+        try:
+            cluster = client.describe_cluster(
+                name='cluster_name'
+            )
+        except ClientError as e:
+            logger.exception(e)
```
deshraj, line 748: We should probably raise after the error occurs rather than proceeding further.

docker/dev/docker.env
```diff
@@ -1,6 +1,6 @@
 AWS_ACCESS_KEY_ID=x
 AWS_SECRET_ACCESS_KEY=x
-AWS_DEFAULT_REGION=us-east-1
+AWS_DEFAULT_REGION=us-east-2
```
deshraj, line 3: @RishabhJain2018 is this change going to affect any of the other running webserver or RDS connection if we change here?

@Ayukha to be safe, lets hard code the region in the function itself where you create the nodegroup.

apps/challenges/aws_utils.py
```diff
@@ -697,3 +699,121 @@ def restart_workers_signal_callback(sender, instance, field_name, **kwargs):
...
+            instanceTypes=["g4dn.xlarge", ],
+            amiType="AL2_x86_64_GPU",
+            nodeRole=settings.EKS_NODEGROUP_ROLE_ARN,
+        )
+    except ClientError as e:
+        logger.exception(e)
+        return response
+    waiter = client.get_waiter("nodegroup_active")
+    waiter.wait(name=nodegroup_name)
+    # create_eks_deployment()
```
deshraj, line 727: Umm, let's remove the function call if not going to be used?

apps/challenges/aws_utils.py
RishabhJain2018: Can we please add the docstrings in this format mentioned here:https://numpydoc.readthedocs.io/en/latest/format.html#sections
```
    """short description

    Arguments:
        x {type} -- description
        y {type} -- description

    Raises: (if applicable)
        error description and when it is raised

    Returns:
        z {type} -- description
    """
```

apps/challenges/aws_utils.py
```diff
@@ -697,3 +699,121 @@ def restart_workers_signal_callback(sender, instance, field_name, **kwargs):
...
+    Creates a nodegroup when a EKS cluster is created by the EvalAI admin
+    """
+    nodegroup_name = "{}-nodegroup".format(instance.title.replace(" ", "-"))
+    client = get_boto3_client("eks", aws_keys)
+    # TODO: Move the hardcoded cluster configuration such as the instance_type, subnets, AMI to challenge configuration later.
+    try:
+        response = client.create_nodegroup(
+            clusterName=cluster_name,
+            nodegroupName=nodegroup_name,
+            scalingConfig={"minSize": 1, "maxSize": 100, "desiredSize": 1},
```
RishabhJain2018, line 715: Can we limit the maxSize to 10 for now.

apps/challenges/aws_utils.py
```diff
@@ -697,3 +699,121 @@ def restart_workers_signal_callback(sender, instance, field_name, **kwargs):
...
+            diskSize=20,
```
RishabhJain2018, line 716: Let's make the diskSize to 100.

apps/challenges/aws_utils.py
```diff
@@ -697,3 +699,121 @@ def restart_workers_signal_callback(sender, instance, field_name, **kwargs):
...
+                challenge=instance.id,
+                name=cluster_name,
+                cluster_yaml=config_file,
+                kube_config=config_file
+            )
+            # Creating nodegroup
+            create_eks_nodegroup(instance, cluster_name)
+        except ClientError as e:
+            logger.exception(e)
+            return response
```
RishabhJain2018, line 819: This should be outside the` try block`.

apps/challenges/aws_utils.py
```diff
@@ -697,3 +699,121 @@ def restart_workers_signal_callback(sender, instance, field_name, **kwargs):
...
+    waiter = client.get_waiter("nodegroup_active")
+    waiter.wait(name=nodegroup_name)
+    # create_eks_deployment()
+
+
+def create_eks_cluster(sender, instance, field_name, **kwargs):
+    """
+    Called when Challenge is approved by the EvalAI admin
+    """
+    from .models import ChallengeEvaluationCluster
```
RishabhJain2018, line 734: Why are we importing it in the function?

settings/common.py
```diff
@@ -304,3 +304,16 @@
 SENDGRID_SETTINGS = {
     "TEMPLATES": {"CHALLENGE_INVITATION": "d-60825bcf014f4958bdb1b9173471d420"}
 }
+
+# EKS configs for Rl-worker
+EKS_CLUSTER_ROLE_ARN = os.environ.get(
+    "EKS_CLUSTER_ROLE_ARN",
+    # "arn:aws:iam::{}:role/evalaieksclusterRole".format(
+    #     aws_keys["AWS_ACCOUNT_ID"]
```
RishabhJain2018, line 312: Can we please remove these if we are not using it?

settings/common.py
```diff
@@ -304,3 +304,16 @@
...
+)
+
+EKS_NODEGROUP_ROLE_ARN = os.environ.get(
+    "EKS_NODEGROUP_ROLE_ARN",
+    # "arn:aws:iam::{}:role/evalaieksnodegroupRole".format(
```
RishabhJain2018, line 317: Can we please remove these if we are not using it?

apps/challenges/aws_utils.py
Ayukha: cyclic
When importing it outside the function, it was a cyclic import which gave the above error. I have removed this for now, will add it in the next PR

apps/challenges/aws_utils.py
```diff
@@ -697,3 +699,109 @@ def restart_workers_signal_callback(sender, instance, field_name, **kwargs):
                 instance.pk, field_name
             )
         )
+
+
+def create_eks_nodegroup(instance, cluster_name):
+    """
+    Creates a nodegroup when a EKS cluster is created by the EvalAI admin
+    """
+    nodegroup_name = "{}-nodegroup".format(instance.title.replace(" ", "-"))
```
vkartik97, line 708: How about `{0}-nodegroup` instead of `{}-nodegroup` ?

apps/challenges/aws_utils.py
```diff
@@ -697,3 +699,109 @@ def restart_workers_signal_callback(sender, instance, field_name, **kwargs):
...
+        return response
+    waiter = client.get_waiter("nodegroup_active")
+    waiter.wait(
+        clusterName=cluster_name,
+        nodegroupName=nodegroup_name,
+    )
+    # TODO:Deploy nvidia deamonset
+
+
+def create_eks_cluster(sender, instance, field_name, **kwargs):
```
vkartik97, line 733: How about replacing `instance` with `challenge` or something that explains it, as it only is used as `Challenge`'s instance.

apps/challenges/aws_utils.py
```diff
@@ -697,3 +699,109 @@ def restart_workers_signal_callback(sender, instance, field_name, **kwargs):
...
+    """
+    Called when Challenge is approved by the EvalAI admin
+    """
+    cluster_name = "{}-cluster".format(instance.title.replace(" ", "-"))
```
vkartik97, line 737: Same as https://github.com/Cloud-CV/EvalAI/pull/2814/files#r432973120

requirements/common.txt
```diff
@@ -1,6 +1,6 @@
 asgi-redis==1.4.3
-boto3==1.9.88
-botocore==1.12.88
+boto3==1.13.16
```
vkartik97, line 2: Just a caveat here, we need to make sure that this upgrade doesnt break any existing flow

Ayukha: @vkartik97  support for EKS APIs have been recently added to boto3. So had to upgrade it.

vkartik97: Sure. We just need to take care that it doesnt break anything 👍

apps/challenges/aws_utils.py
vkartik97: Can you add in this PR?
Its ok to have imports here because cyclic imports
From [Python Docs](https://docs.python.org/3.8/tutorial/modules.html#more-on-modules)
>It is customary but not required to place all import statements at the beginning of a module (or script, for that matter). The imported module names are placed in the importing module’s global symbol table.

docker/dev/docker.env
RishabhJain2018: @Ayukha Can you please remove this change from this PR and hardcode it in the function itself.

Ayukha:
@RishabhJain2018  sure

PR #2819, opened 2020-05-25
Add backend APIs for github badges feature

RishabhJain2018:
**Description:**
This feature will enable participants to add badges to their Github code repositories for the entry on the EvalAI leaderboard and the rank on the badge will be dynamic.
Sample badge image -
Screenshot 2020-05-25 at 3 21 01 AM

This PR contains -
- [x] Backend API for fetching data for GitHub Badges.
- [x] Refactor leaderboard API code.
- [x] Add API for fetching details of the participant team in a challenge from user auth token and challenge_pk.

### DIFF
--- a/apps/jobs/urls.py
+++ b/apps/jobs/urls.py
@@ -75,4 +75,9 @@ urlpatterns = [
         views.get_bearer_token,
         name="get_bearer_token",
     ),
+    url(
+        r"^phase_split/(?P<challenge_phase_split_pk>[0-9]+)/team/(?P<participant_team_pk>[0-9]+)/github_badge/$",
+        views.get_github_badge_data,
+        name="get_github_badge_data",
+    ),
 ]
--- a/apps/jobs/urls.py
+++ b/apps/jobs/urls.py
@@ -76,8 +76,8 @@ urlpatterns = [
         name="get_bearer_token",
     ),
     url(
-        r"^phase_split/(?P<challenge_phase_split_pk>[0-9]+)/team/(?P<participant_team_pk>[0-9]+)/github_badge/$",
-        views.github_badge_data,
-        name="github_badge_data",
+        r"^phase_splits/(?P<challenge_phase_split_pk>[0-9]+)/teams/(?P<participant_team_pk>[0-9]+)/github_badge/$",
+        views.get_github_badge_data,
+        name="get_github_badge_data",
     ),
 ]
--- a/apps/jobs/urls.py
+++ b/apps/jobs/urls.py
@@ -77,7 +77,7 @@ urlpatterns = [
     ),
     url(
         r"^phase_split/(?P<challenge_phase_split_pk>[0-9]+)/team/(?P<participant_team_pk>[0-9]+)/github_badge/$",
-        views.get_github_badge_data,
-        name="get_github_badge_data",
+        views.github_badge_data,
+        name="github_badge_data",
     ),
 ]
--- a/apps/jobs/utils.py
+++ b/apps/jobs/utils.py
@@ -231,7 +231,18 @@ def handle_submission_rerun(submission, updated_status):
 def calculate_distinct_sorted_leaderboard_data(
     user, challenge_obj, challenge_phase_split, only_public_entries
 ):
+    """
+    Function to calculate and return the sorted leaderboard data
 
+    Arguments:
+        user {[Class object]} -- User model object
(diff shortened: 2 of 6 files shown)


apps/jobs/urls.py
```diff
@@ -75,4 +75,9 @@
         views.get_bearer_token,
         name="get_bearer_token",
     ),
+    url(
+        r"^phase_split/(?P<challenge_phase_split_pk>[0-9]+)/team/(?P<participant_team_pk>[0-9]+)/github_badge/$",
```
deshraj, line 79: lets change the url to this 
`r"^phase_splits/(?P<challenge_phase_split_pk>[0-9]+)/teams/(?P<participant_team_pk>[0-9]+)/github_badge/$"`

deshraj: Notice the `s` at the end for phase split and team.

apps/jobs/utils.py
```diff
@@ -222,3 +226,142 @@ def handle_submission_rerun(submission, updated_status):
...
+
+    # Get the leaderboard associated with the Challenge Phase Split
+    leaderboard = challenge_phase_split.leaderboard
+
+    # Get the default order by key to rank the entries on the leaderboard
+    try:
+        default_order_by = leaderboard.schema["default_order_by"]
+    except KeyError:
+        response_data = {
+            "error": "Sorry, Default filtering key not found in leaderboard schema!"
```
deshraj, line 243: nit: please change the message to `Sorry, `default_order_by` key missing in leaderboard schema!`

apps/jobs/utils.py
```diff
@@ -222,3 +226,142 @@ def handle_submission_rerun(submission, updated_status):
                 "submitted_image_uri"
             ]
     return message
+
+
+def calculate_distinct_sorted_leaderboard_data(
```
deshraj, line 231: Can you please add numpy based docstring https://numpydoc.readthedocs.io/en/latest/format.html#id8?

apps/jobs/views.py
```diff
@@ -491,146 +492,21 @@ def leaderboard(request, challenge_phase_split_id):
...
-        sorted_leaderboard_data = sorted(
-            leaderboard_data,
-            key=lambda k: (
-                float(k["filtering_score"]),
-                float(-k["filtering_error"]),
-            ),
-            reverse=True
-            if challenge_phase_split.is_leaderboard_order_descending
-            else False,
+    if http_status_code == status.HTTP_400_BAD_REQUEST:
```
deshraj, line 505: I think I am missing something here. If response is 400, why we want to return the data? :/

Also, can you please add a comment above this line to say what action to take when response is 400?

apps/jobs/views.py
```diff
@@ -655,129 +531,30 @@ def get_all_entries_on_public_leaderboard(request, challenge_phase_split_pk):
...
```
deshraj, line 553: Same. Please add a comment before this line to explain action to be taken here.

apps/participants/urls.py
```diff
@@ -38,4 +38,9 @@
         views.participant_team_detail,
         name="get_participant_team_details",
     ),
+    url(
+        r"^challenge/(?P<challenge_pk>[0-9]+)/team_detail/$",
```
deshraj, line 42: Let's change the url pattern to `r"^challenges/(?P<challenge_pk>[0-9]+)/team_details/$"`

apps/participants/views.py
```diff
@@ -386,3 +395,33 @@ def remove_self_from_participant_team(request, participant_team_pk):
         if participants.count() == 0:
             participant_team.delete()
         return Response(status=status.HTTP_204_NO_CONTENT)
+
+
+@api_view(["GET"])
+@throttle_classes([UserRateThrottle])
+@permission_classes((permissions.IsAuthenticated, HasVerifiedEmail))
+@authentication_classes((ExpiringTokenAuthentication,))
+def participant_team_detail_in_challenge(request, challenge_pk):
```
deshraj, line 404: Please update the view function name accordingly to `get_participant_team_details_for_challenge`.

apps/jobs/views.py
RishabhJain2018: We want to return the error message. The variable name was wrong. I've changed it.

apps/jobs/views.py
RishabhJain2018: The action will be returned by the function.

PR #2822, opened 2020-05-27
[Backend] Use .filter for restrict one leaderboard submission challenge visibility update

Ram81:
### Description

To avoid exceptions when handling the case of updating challenge submission visibility for challenge phases with `is_restricted_to_one_leaderboard_submission` use `.filter` instead of `.get`

### DIFF
--- a/apps/jobs/views.py
+++ b/apps/jobs/views.py
@@ -372,8 +372,7 @@ def change_submission_data_and_visibility(
             when_made_public = datetime.datetime.now()
             request.data["when_made_public"] = when_made_public
 
-        try:
-            submissions_already_public = Submission.objects.get(
+            submissions_already_public = Submission.objects.filter(
                 is_public=True,
                 participant_team=participant_team,
                 challenge_phase=challenge_phase,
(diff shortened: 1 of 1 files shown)


apps/jobs/views.py
```diff
@@ -382,12 +381,12 @@ def change_submission_data_and_visibility(
...
-                and submissions_already_public
+                and submissions_already_public.count() == 1
             ):
-                submissions_already_public.is_public = False
-                submissions_already_public.save()
-        except Submission.DoesNotExist:
-            pass
+                # Assuming only 1 submission would be public
+                for submission_already_public in submissions_already_public:
+                    submission_already_public.is_public = False
```
RishabhJain2018, line 388: Can we please use serializer to update the data.

apps/jobs/views.py
```diff
@@ -382,12 +381,12 @@ def change_submission_data_and_visibility(
...
                 challenge_phase.is_restricted_to_select_one_submission
                 and is_public
-                and submissions_already_public
+                and submissions_already_public.count() == 1
             ):
-                submissions_already_public.is_public = False
-                submissions_already_public.save()
-        except Submission.DoesNotExist:
-            pass
+                # Assuming only 1 submission would be public
```
RishabhJain2018, line 386: # Case when the phase is restricted to make only one submission as public

apps/jobs/views.py
```diff
@@ -382,12 +381,12 @@ def change_submission_data_and_visibility(
...
+                for submission_already_public in submissions_already_public:
```
RishabhJain2018, line 387: We don't need a loop here, right? Since we know this list will only have 1 entry we can simply use the array indexing to parse it through the serializer.

Ram81: i tried using index but it wasn't reflecting in db seems like a access by reference issue that's why used for loop here

PR #2823, opened 2020-05-27
Add frontend for badges feature and minor API changes

RishabhJain2018:
Extension to #2819 
This PR contains -
- [x] Frontend for displaying badges for leaderboard entry.
- [x] API changes to remove authentication and add a check if the participant team exists or not.
Sample UI of the frontend -
Screenshot 2020-05-27 at 11 58 54 AM
Screenshot 2020-05-27 at 11 58 45 AM

### DIFF
--- a/apps/jobs/views.py
+++ b/apps/jobs/views.py
@@ -1488,11 +1488,9 @@ def get_bearer_token(request, challenge_pk):
 
 
 @api_view(["GET"])
-@throttle_classes([UserRateThrottle])
-@permission_classes((permissions.IsAuthenticated, HasVerifiedEmail))
-@authentication_classes((ExpiringTokenAuthentication,))
+@throttle_classes([AnonRateThrottle])
 def get_github_badge_data(
-    request, challenge_phase_split_pk, participant_team_pk
+    request, challenge_phase_split_pk, participant_team_name
 ):
     """
     Add API to get data for dynamically generating github badges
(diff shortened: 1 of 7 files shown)


apps/jobs/views.py
```diff
@@ -1488,11 +1488,9 @@ def get_bearer_token(request, challenge_pk):
 
 
 @api_view(["GET"])
-@throttle_classes([UserRateThrottle])
-@permission_classes((permissions.IsAuthenticated, HasVerifiedEmail))
-@authentication_classes((ExpiringTokenAuthentication,))
+@throttle_classes([AnonRateThrottle])
 def get_github_badge_data(
-    request, challenge_phase_split_pk, participant_team_pk
+    request, challenge_phase_split_pk, participant_team_name
```
deshraj, line 1493: Happy with this switch from participant_team_pk to participant_team_name but just curious what was the reasoning behind changing this?

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -73,6 +73,7 @@
 
         vm.isChallengeLeaderboardPrivate = false;
         vm.previousPublicSubmissionId = null;
+        vm.participantTeamDetail = "";
```
deshraj, line 76: Not a big fan of defining empty string. Maybe just declare it and don't set any value.

frontend/src/views/web/challenge/github-badge.html
```diff
@@ -0,0 +1,71 @@
...
+                        <div class="url">
+                            <a class="blue-text" target="_blank" href="https://img.shields.io/endpoint?url={{challenge.githubBadgeURL}}">
+                                https://img.shields.io/endpoint?url={{challenge.githubBadgeURL}}
+                            </a>
+                        </div>
+                    </div>
+                </div>
+            </div>
+            <div class="align-right text-med-black w-400">
+                <ul class="inline-list pointer">
```
deshraj, line 61: Probably doesn't make sense to use a `ul` element for a `Close` button. If the problem was with alignment, I am sure there is a better way to handle it using css.

frontend/src/views/web/challenge/leaderboard.html
```diff
@@ -124,6 +124,7 @@ <h5 class="w-300">Leaderboard</h5>
                             </td>
                         </tr>
                     </thead>
+                    {{challenge.phaseRemainingSubmissions.participant_team}}
```
deshraj, line 127: Wondering what is the significance of this?

deshraj:
Found a small bug. When logged in as a user who hasn't participated in the challenge, then it shows an error (see screenshot below)
Screenshot from 2020-05-28 00-27-50


Also, seems like we want to render the badge for everyone and not just the participant

Issue #2116, opened 2019-02-12
Add a progress bar showing in the percentage of file uploaded during a submission


RishabhJain2018:
## Deliverables

- [ ] A progress bar which will appear on the upload of a file showing the percentage of the file uploaded to the server.

sachinmukherjee:
Can I work on this?

RishabhJain2018:
Sure! but please keep me posted on this issue since we need it urgently.

sachinmukherjee:
Ok

sachinmukherjee:
As this is of high priority and due to other works I can't say when i will be able to submit a PR for this one. So this issue is still open.

yashdusing:
Can I take this up ?

ArshdeepSingh98:
Can I take this issue ?

pushkalkatara:
Hi @RishabhJain2018 The [PR](https://github.com/Cloud-CV/EvalAI/pull/2120) was closed due to inactivity. There was an issue with the color progress bar + merge conflicts. Should I fix the same and make a new PR?

ankitpal1029:
@RishabhJain2018 can i take this issue up?

ankitpal1029:
@Kajol-Kumari @lunayach  @RishabhJain2018  i thought the initial purpose of the issue was to show progress of the upload to the server rather than just a dummy progress bar(correct me if i'm wrong on this one).

since the actual file upload happens only when the user submits the form maybe we could show the progress bar right after submit instead of this loading screen:
![image](https://user-images.githubusercontent.com/49815751/114262059-8e68b580-99fb-11eb-9ab6-684ebfb07a1e.png)

What are your thoughts on this

DevalGarg129:
can i resolve this issue ?
@RishabhJain2018

devjayswal:
can i work in this

PR #2832, opened 2020-06-02
[WIP] Add progress bar on file selection under submission

Kajol-Kumari:
Fix: #2116 

**Changes introduced**
A progress bar will be shown after the submission file is selected.

GIF:
ezgif com-video-to-gif

### DIFF
--- a/frontend_v2/src/app/components/challenge/challengesubmit/challengesubmit.component.ts
+++ b/frontend_v2/src/app/components/challenge/challengesubmit/challengesubmit.component.ts
@@ -260,7 +260,6 @@ isSubmissionUsingUrl: any;
         sub.unsubscribe();
       }
     });
-
   }
 
   /**
--- a/frontend_v2/src/app/components/challenge/challengesubmit/challengesubmit.component.ts
+++ b/frontend_v2/src/app/components/challenge/challengesubmit/challengesubmit.component.ts
@@ -5,6 +5,7 @@ import { GlobalService } from '../../../services/global.service';
 import { ChallengeService } from '../../../services/challenge.service';
 import { Router, ActivatedRoute } from '@angular/router';
 import { EndpointsService } from '../../../services/endpoints.service';
+import { interval } from 'rxjs';
 
 /**
  * Component Class
@@ -165,6 +166,21 @@ isSubmissionUsingUrl: any;
   timer: any;
 
   /**
+   * File upload progress value
+   */
+  progressValue = 0;
+
+  /**
+   * File upload progress time
+   */
+  curSec = 0;
+
+  /**
+   * Is file seleced
+   */
+  isFileSelected = false;
+
+  /**
    * Component Class
    */
   @ViewChildren('formsubmit')
@@ -228,6 +244,21 @@ isSubmissionUsingUrl: any;
       this.displayDockerSubmissionInstructions(this.challenge.id, this.isParticipated);
     }
     this.authToken = this.globalService.getAuthToken();
+  }
+
+  progress() {
+    this.isFileSelected = true;
+    const timer$ = interval(100);
+
+    const sub = timer$.subscribe((sec) => {
+      this.progressValue = 0 + sec * 100 / 5;
+      this.curSec = sec;
+
+      if (this.curSec === 5) {
+        sub.unsubscribe();
+      }
+    });
+
   }
 
   /**
--- a/frontend_v2/src/app/components/challenge/challengesubmit/challengesubmit.component.ts
+++ b/frontend_v2/src/app/components/challenge/challengesubmit/challengesubmit.component.ts
@@ -246,15 +246,17 @@ isSubmissionUsingUrl: any;
     this.authToken = this.globalService.getAuthToken();
   }
 
-  progress() {
+  progress(files: FileList) {
     this.isFileSelected = true;
-    const timer$ = interval(100);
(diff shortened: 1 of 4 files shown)


frontend_v2/src/app/components/challenge/challengesubmit/challengesubmit.component.ts
```diff
@@ -230,6 +246,21 @@ isSubmissionUsingUrl: any;
     this.authToken = this.globalService.getAuthToken();
   }
 
+  progress() {
```
lunayach, line 249: Seems like the progress status is independent of the size of file (or the actual upload time). Please correct me if I am missing something here.

Kajol-Kumari: yes @lunayach I was just giving the 5 second time for uploading during which i was showing the slider. Now I have changed the time for upload to depends on the size of the file. Please take a look.

frontend_v2/src/app/components/challenge/challengesubmit/challengesubmit.component.ts
```diff
@@ -230,6 +246,22 @@ isSubmissionUsingUrl: any;
     this.authToken = this.globalService.getAuthToken();
   }
 
+  progress(files: FileList) {
+    this.isFileSelected = true;
+    const fileToUpload = files.item(0);
+    const seconds = fileToUpload.size;
```
lunayach, line 252: What's the unit of `fileToUpload.size`? Could this be taken equivalent to seconds?

frontend_v2/src/app/components/challenge/challengesubmit/challengesubmit.component.ts
```diff
@@ -230,6 +246,22 @@ isSubmissionUsingUrl: any;
...
+    const timer$ = interval(10);
+
+    const sub = timer$.subscribe((sec) => {
+      this.progressValue = 0 + sec * 100 / seconds;
+      this.curSec = sec;
+
+      if (this.curSec === seconds) {
```
lunayach, line 259: Seems like it's also possible to unsubscribe even before the progressValue becoming 100. Will this condition become true for every upload?

frontend_v2/src/app/components/challenge/challengesubmit/challengesubmit.component.ts
Kajol-Kumari: @lunayach what i am trying to do here is giving 10ms time for each byte of the filesize. As we are showing the progress bar as soon as we select the file so i couldn't find any way to get the status of file selection and so using this way to show the progress bar.

lunayach: How did you reach to the `10 ms` number?

frontend_v2/src/app/components/challenge/challengesubmit/challengesubmit.component.ts
Kajol-Kumari: @lunayach as mentioned in the condition [here](https://github.com/Cloud-CV/EvalAI/pull/2832) The `curSec` value will be equal to the `seconds` value only when the `progressValue` becomes 100. And yes for every upload, this `seconds` will change based on the file size and will only be subscribed if the `progressValue` becomes 100.

frontend_v2/src/app/components/challenge/challengesubmit/challengesubmit.component.ts
Kajol-Kumari: For this, i have considered the upload speed of 1Mbps

lunayach: https://usefulangle.com/post/67/pure-javascript-ajax-file-upload-showing-progess-percent
^ Can we have something like progress event of the `XMLHttpRequest.upload`?

Kajol-Kumari: @lunayach This will require the File to be uploaded to the server and here we just need to show the progress bar on file selection. Please correct me if i missed something.

lunayach: I see. This was the deliverable of the issue #2116, `A progress bar which will appear on the upload of a file showing the percentage of the file uploaded to the server.` So, I think the progress bar should be showing the progress of the file uploaded. @RishabhJain2018 @Sanji515 please clarify if I misunderstood this.

Kajol-Kumari:
> * Just want to know if you have checked the various file size upload.
> * Not sure about the submission backend part - but  I think it can be done using the already existing method. Let me know if you have already tried it.

yes @Shekharrajak I have tried uploading the files of different file size and they take time according to their file sizes in this current PR.

And regarding the current submission backend part, we want to show the progress bar as soon as user select the submission file not when user submits the submission form (as this will just increase the waiting time by stopping the loader and showing the progress bar also in the ngx UI the user won't be able to see that progress bar) and so we need to modify the API to get the upload status as soon as user selects the file.

RishabhJain2018:
This PR has been open for a long time and now has conflicts with the latest code. To keep the repository clean, we’re closing it for now. If you’d like to update the branch and resolve conflicts, feel free to reopen or submit a new PR. Thanks for your contribution!

Issue #2484, opened 2019-11-24
UI fades out for few seconds and then fades in 


vs74:
**The Observed Behaviour**
**GIF**
![ezgif com-video-to-gif](https://user-images.githubusercontent.com/31419210/69490542-b9f30a80-0eaf-11ea-8e35-a434d3476794.gif)

Here is a sample GIF wherein I tried to navigate between diff functionalists ( `AI Challenges`, `Hosted Challenge`s, `Create Challenge`). Notice when I click on `AI Challenges`, for a fraction of seconds the screen fades out and then it properly loads(fade in) the information on the web page. This behaviour is observed during first few seconds. Though I don't know what might the exact cause of the problem (perhaps response time, UI design or it might be the some effect or transition ) as I am not a expert in Web Development

**Steps to Reproduce the Problem**

1. Login in to EvalAI
2. Try Navigating between diff functionalities
3. You will notice for a few initial seconds fade out effect and then it fades in.

**Expected Behaviour**
1. If it's a bug, we should handle it. 

2. If this is probable effect which has been included in UI, I think it's better to reduce the delay between the fade in and fade out

aka-jain:
Hi @vs74 this is not an effect, just a loading screen waiting for api to getting finished. I guess somehow, loader is not showing up here. We can work on that. See other pages where you can see loader as well as fadeout screen.

PR #2839, opened 2020-06-09
Loader visible on navigating between functionalities

Kajol-Kumari:
Fix: #2484 [ UI fades out for few seconds and then fades in ]

**After Fix**
A loader will be shown till all the APIs get fetched along with the screen fading.

**GIF of fix:**
ezgif com-video-to-gif (1)


frontend/src/views/web/hosted-challenges.html
```diff
@@ -39,4 +39,17 @@
...
+        <div class="loader">
+            <div></div>
+            <div></div>
+            <div></div>
+            <div></div>
+            <div></div>
+        </div>
+    </div>
+    
 </section>
```
RishabhJain2018, line 55: Can you please add a new line here?

PR #2841, opened 2020-06-09
[Documentation] Fix readthedocs build failure

Ram81:
Readthedocs builds for EvalAI have been failing for quite a long time. You can find it [here](https://readthedocs.org/projects/evalai/builds/). 

### Build Failure Cause:

In our initial setup of readthedocs we were using `requirements/prod.txt` to install depedencies for readthedocs build and package `pycurl` was causing issue. 

### Fix:

We installed all prod environment python packages which were not required for readthedocs build. I changed the way we were managing config for our readthedocs setup. You can find the setup method [here](https://docs.readthedocs.io/en/stable/config-file/v2.html). I also separated requirements for readthedocs build to avoid future issues in dependencies. Only two dependencies which were required for the build to complete successfully were:

```
sphinx_rtd_theme
sphinx_markdown_tables
```


requirements/readthedocs.txt
```diff
@@ -0,0 +1,2 @@
+sphinx_rtd_theme
```
RishabhJain2018, line 1: Can we please add the specific versions for this?

requirements/readthedocs.txt
```diff
@@ -0,0 +1,2 @@
...
+sphinx_markdown_tables
```
RishabhJain2018, line 2: Same as above.

PR #2853, opened 2020-06-15
Add script for deploying EvalAI on VMs

yashdusing:
- [x] Uploaded the dev images to dockerhub
- [x] Adds new compose file to deploying locally  
- [x] Bash script to install requirements & pull images & deploy on VMs 
- [x] Database backup scheduled daily & backups deleted every 7 days (BACKUP_KEEP_DAYS flag). For manual backup use (`sudo docker exec -it $pgbackups_container_name /backup.sh -e POSTGRES_HOST=postgres -e POSTGRES_DB=db -e POSTGRES_USER=postgres -e POSTGRES_PASSWORD=postgres`) and manual restore use ( `sudo docker exec -it $pgbackups_container_name gunzip -c $backup_path | psql -h db -p 5432 -U postgres -d postgres` )
- [x] Setup nginx + nodejs with HTTP for frontend deployment
- [x] Setup nginx + nodejs with HTTPs & custom SSL certificates for frontend deployment

Steps : 
1. Launch an EC2 instance
2. Set up domain names (say `evalai.example.com` and `evalapi.example.com`) in route53 to point to EC2 instance (ideally its elastic IP)
3. Change `example.com` & `user@email.com` in codefiles with your domain name in codebase. Also, change ssl certifications config while generating for your organisation
4. ssh into EC2 instance and change directory to EvalAI
5. ./deploy-vm.sh


deploy-vm.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
+ 
+# Install requirements (docker)
+
+echo "Step 1"
```
Ram81, line 5: @yashdusing it would be a good idea to add more descriptive statements instead of just `Step 1` `Step 2`. You can use statements like `Pulling docker image...`, etc

docker-compose-local.yml
```diff
@@ -0,0 +1,82 @@
...
+    ports:
+      - "8888:8888"
+      - "35729:35729"
+    volumes:
+      - .:/code
+      - /code/node_modules
+      - /code/bower_components
+
+  # TODO: Add alternative for AWS SQS Celery (https://docs.celeryproject.org/en/stable/getting-started/brokers/sqs.html) in dev setup.
+  # celery:
```
Ram81, line 71: if the commented code is not needed can you remove it

deploy-vm.sh
```diff
@@ -0,0 +1,57 @@
...
+
+
+echo "### Installing docker-compose"
+sudo curl -L "https://github.com/docker/compose/releases/download/1.26.0/docker-compose-$(uname -s)-$(uname -m)" -o /usr/local/bin/docker-compose
+sudo chmod +x /usr/local/bin/docker-compose
+
+echo "### Finished installing docker-compose. Check docker-compose version"
+sudo docker-compose --version 
+
+export DJANGO_SERVER_URL=temp
```
Ram81, line 36: is this needed? I don't see it being used anywhere

deploy-vm.sh
```diff
@@ -0,0 +1,57 @@
...
+
+
+echo "### Initiating letsencrypt with certbot"
+chmod +x init-letsencrypt.sh
+./init-letsencrypt.sh
+
+# Pull images & run containers 
+sudo docker-compose -f docker-compose-vm.yml up -d --build
+
+#Restore database
```
Ram81, line 46: Add space after `#`

deploy-vm.sh
```diff
@@ -0,0 +1,57 @@
...
+LOCAL_DUMP_PATH=$(ls -t backups/* | head -1)
+
+if [ -n "${DUMP_FILE}" ]
+then    
+    echo "# Backup file exists: ${DUMP_FILE}"
+    echo "# Restoring latest postgres backup"
+    sudo docker exec -it "${DOCKER_DB_NAME}" gunzip -c "${LOCAL_DUMP_PATH}" | psql -h db -p 5432 -U postgres -d postgres
+else
+    echo "# No backup file exists"
+fi
```
Ram81, line 57: add newline

deploy-vm.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+
+# Install requirements (docker)
+echo "### Removing existing docker libs"
+sudo apt-get remove docker docker-engine docker.io containerd runc -y
```
Ram81, line 5: no need to remove docker if it already exists

deploy-vm.sh
```diff
@@ -0,0 +1,57 @@
...
+sudo apt-get update -y
+
+echo "### Installing docker requirements"
+sudo apt-get install \
+    apt-transport-https \
+    ca-certificates \
+    curl \
+    gnupg-agent \
+    software-properties-common -y
+curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
```
Ram81, line 15: add newline above this

deploy-vm.sh
```diff
@@ -0,0 +1,57 @@
...
+sudo add-apt-repository \
+   "deb [arch=amd64] https://download.docker.com/linux/ubuntu \
+   $(lsb_release -cs) \
+   stable"
+sudo apt-get update -y
```
Ram81, line 20: samer here

docker-compose-vm.yml
```diff
@@ -0,0 +1,83 @@
...
+      - ./certbot/www:/var/www/certbot
+    command: "/bin/sh -c 'while :; do sleep 6h & wait $${!}; nginx -s reload; done & nginx -g \"daemon off;\"'"
+
+  certbot:
+    image: certbot/certbot
+    restart: unless-stopped
+    volumes:
+      - ./certbot/conf:/etc/letsencrypt
+      - ./certbot/www:/var/www/certbot
+    entrypoint: "/bin/sh -c 'trap exit TERM; while :; do certbot renew; sleep 12h & wait $${!}; done;'"
```
Ram81, line 83: add newline

docker/prod/nodejs/Dockerfile
```diff
@@ -26,4 +26,4 @@ FROM nginx:1.13-alpine
 ARG NODE_ENV
 COPY docker/prod/nodejs/nginx_${NODE_ENV}.conf /etc/nginx/conf.d/default.conf
 COPY --from=node /code /code
-COPY /ssl /code/ssl
+# COPY /ssl /code/ssl
```
Ram81, line 29: this is the prod docker file is this change relevant to deployment change?

docker/prod/nodejs/nginx_vm.conf
```diff
@@ -0,0 +1,78 @@
...
+  location /.well-known/acme-challenge/ {
+    root /var/www/certbot;
+  }
+}
+
+server {
+  listen 443 ssl;
+  server_name evalapi.example.com;
+  client_max_body_size 400M;
+  # rewrite ^/(.*) http://evalapi.example.com/$1 permanent;
```
Ram81, line 23: is this a comment? if not needed can you remove this

yashdusing:
@Ram81 sure I’ll start work on this this saturday.

docker/prod/nodejs/nginx_vm.conf
```diff
@@ -0,0 +1,78 @@
...
+  gzip_min_length   256;
+  gzip_vary         on;
+  gzip_proxied      expired no-cache no-store private auth;
+  gzip_types        text/plain text/css application/json application/javascript application/x-javascript text/xml application/xml application/xml+rss text/javascript;
+  gzip_comp_level   9;
+  root /code/frontend;
+  location / {
+    try_files $uri $uri/ /index.html =404;
+  }
+}
```
Ram81, line 78: add newline

init-letsencrypt.sh
```diff
@@ -0,0 +1,79 @@
+#!/bin/bash
+
+domains=(example.com evalai.example.com evalapi.example.com)
+rsa_key_size=4096
+data_path="./certbot"
+email="user@email.com" # Adding a valid address is strongly recommended
```
Ram81, line 6: take this as a parameter in `./deploy-vm.sh`.

init-letsencrypt.sh
```diff
@@ -0,0 +1,79 @@
...
+sudo docker-compose -f docker-compose-vm.yml run --rm --entrypoint "\
+  rm -Rf /etc/letsencrypt/live/$domains && \
+  rm -Rf /etc/letsencrypt/archive/$domains && \
+  rm -Rf /etc/letsencrypt/renewal/$domains.conf" certbot
+echo
+
+
+echo "### Requesting Lets Encrypt certificate for $domains ..."
+echo
+#Join $domains to -d args
```
Ram81, line 52: be consistent with comments

init-letsencrypt.sh
```diff
@@ -0,0 +1,79 @@
...
+    $email_arg \
+    $domain_args \
+    --rsa-key-size $rsa_key_size \
+    --agree-tos \
+    --force-renewal" certbot
+echo
+
+echo "### Reloading nginx ..."
+echo
+sudo docker-compose -f docker-compose-vm.yml exec nodejs nginx -s reload
```
Ram81, line 79: add newline

deploy-vm.sh
```diff
@@ -0,0 +1,57 @@
...
+echo "### Initiating letsencrypt with certbot"
+chmod +x init-letsencrypt.sh
+./init-letsencrypt.sh
+
+# Pull images & run containers 
+sudo docker-compose -f docker-compose-vm.yml up -d --build
+
+#Restore database
+DOCKER_DB_NAME="$(sudo docker-compose -f docker-compose-vm.yml ps -q db)"
+LOCAL_DUMP_PATH=$(ls -t backups/* | head -1)
```
Ram81, line 48: maybe you should add a check for `backup` folder. If it doesn't exist create one

docker/prod/nodejs/nginx_vm.conf
```diff
@@ -0,0 +1,78 @@
+upstream django_app {
+  server django:8000 fail_timeout=0;
+}
+
+# Django server
+server {
+  listen 80;
+  server_name evalapi.example.com;
```
Ram81, line 8: we have to take domain name as a parameter, hardcoding it doesn't make sense

init-letsencrypt.sh
```diff
@@ -0,0 +1,79 @@
+#!/bin/bash
+
+domains=(example.com evalai.example.com evalapi.example.com)
```
Ram81, line 3: take this as a parameter in ./deploy-vm.sh.

yashdusing:
`ls: cannot access 'backups/*': No such file or directory
# No backup file exists` is okay. That means that it worked. I'll try getting all inputs from script itself. 
I was doing that with environment variables initially but the env variables weren't getting substituted in several places like gulp.js, nginx config, etc after docker-compose command. I'll retry that however

Ram81:
@yashdusing I tried accessing the frontend and backend using http://ip:port. I can access the django backend on (http://ip:8000) but the frontend doesn't seem to be running on http://ip:80. Can you check the configurations once

docker/prod/nodejs/Dockerfile
yashdusing: no. I'll create a different Dockerfile. But, I'm not sure where it should go. Should I create a different env folder ?

yashdusing:
So nginx will only accept traffic from domain pointing to machine & not directly through ip itself. You'll have to make a route53 entry for evalai.domain.com, evalapi.domain.com, domain.com to point to machine io

Ram81:
But if the IP is public you should allow access right? Any specific reason why you strictly want to allow access using domain only?

yashdusing:
You can do that. But, you'll have to add nginx config entry for that as well

Ram81:
Yes, we should have that too. I don't see any reason to why you are not allowing access directly using IP.

docker/prod/nodejs/Dockerfile
yashdusing: Should I create a different env folder here ? like /vm/nodejs/Dockerfile ?

yashdusing:
@Ram81 actually I just remembered. With the way certbot works, certbot requires you to complete a task on server end & serve it at a particular endpoint of the domain. So, for ssl you'll need certbot & domain. So, I'm not there's a way to add nginx config entry for IP.

yashdusing:
@Ram81 could you check this ?

init-letsencrypt.sh
```diff
@@ -0,0 +1,78 @@
...
+touch .env
+echo "DOMAIN_NAME=$DOMAIN_NAME" > .env
+
+rsa_key_size=4096
+data_path="./certbot"
+
+# Adding a valid address is strongly recommended
+read -p "Enter your Email ID : " email
+
+staging=1 # Set to 1 if you're testing your setup to avoid hitting request limits
```
Ram81, line 15: why is this hardcoded to 1? can you take this a parameter to the script

deploy-vm.sh
Ram81: why is this not resolved?

init-letsencrypt.sh
```diff
@@ -0,0 +1,78 @@
+#!/bin/bash
+
+read -p "Enter your domain name (example.com) : " DOMAIN_NAME
```
Ram81, line 3: @yashdusing this is not clear. Should we enter space separated domain names? Please add appropriate comment

init-letsencrypt.sh
```diff
@@ -0,0 +1,78 @@
...
+domains=($DOMAIN_NAME evalai.$DOMAIN_NAME evalapi.$DOMAIN_NAME)
```
Ram81, line 4: this can be change to only `domains=$DOMAIN_NAME`. Accept both the domain names not just the suffix. We can't add fixed prefix

docker/vm/nodejs/nginx_vm.conf
```diff
@@ -0,0 +1,77 @@
...
+
+  location /.well-known/acme-challenge/ {
+    root /var/www/certbot;
+  }
+}
+
+server {
+  listen 443 ssl;
+  sendfile on;
+  default_type application/octet-stream;
```
Ram81, line 56: please add appropriate comments over all the lines which have to be changed. Or add a setup.md

init-letsencrypt.sh
```diff
@@ -0,0 +1,78 @@
...
+
+echo "### Deleting dummy certificate for $domains ..."
+sudo docker-compose -f docker-compose-vm.yml run --rm --entrypoint "\
+  rm -Rf /etc/letsencrypt/live/$domains && \
+  rm -Rf /etc/letsencrypt/archive/$domains && \
+  rm -Rf /etc/letsencrypt/renewal/$domains.conf" certbot
+
+
+echo "### Requesting Lets Encrypt certificate for $domains ..."
+domain_args=""
```
Ram81, line 53: isn't this supposed to be accepted as params to script? if yes, when I run the script I don't get a prompt to enter paths

init-letsencrypt.sh
```diff
@@ -0,0 +1,78 @@
...
+    -keyout '$path/privkey.pem' \
+    -out '$path/fullchain.pem' \
+    -subj '/CN=localhost'" certbot
+
+
+echo "### Starting nginx ..."
+sudo docker-compose -f docker-compose-vm.yml up --force-recreate -d nodejs
+
+echo "### Deleting dummy certificate for $domains ..."
+sudo docker-compose -f docker-compose-vm.yml run --rm --entrypoint "\
```
Ram81, line 46: why are we creating dummy certificate and deleting it?

docker-compose-vm.yml
```diff
@@ -0,0 +1,86 @@
...
+
+  certbot:
+    image: certbot/certbot
+    restart: unless-stopped
+    environment: 
+      DOMAIN_NAME: ${DOMAIN_NAME}
+    volumes:
+      - ./certbot/conf:/etc/letsencrypt
+      - ./certbot/www:/var/www/certbot
+    entrypoint: "/bin/sh -c 'trap exit TERM; while :; do certbot renew; sleep 12h & wait $${!}; done;'"
```
Ram81, line 86: no newline?

Ram81:
@yashdusing If there's no ssl cert specified then can't we just host on IP address with `http`?

yashdusing:
@Ram81 we can do it with just http. But that would require changing the nginx config.
As for the issue you're facing, I think that's because for ssl certification I explicitly kept evalai.${DOMAIN} & evalapi.${DOMAIN} as nodejs/django. So anything like evalai-test.${DOMAIN} wouldn't work unfortunately. You can change that in domains for init-letscrypt.

init-letsencrypt.sh
yashdusing: no. So in our case it would be `cloudcv.org`

Ram81:
@yashdusing I think we decided on accepting domain names as Params to shell script. Can you please make that change.

yashdusing:
Yeah. I thought we would have fixed prefixes. Also, could you post the nodejs logs here please

init-letsencrypt.sh
Ram81: this is not correct. We already host our production evalai at `evalai.cloudcv.org` we should not accept suffix. We should accept complete domain names

Ram81:
```
2020/08/15 15:23:34 [emerg] 7#7: BIO_new_file("/etc/letsencrypt/live/evalai-test-ec2-frontend.cloudcv.org/fullchain.pem") failed (SSL: error:02001002:system library:fopen:No such file or directory:fopen('/etc/letsencrypt/live/evalai-test-ec2-frontend.cloudcv.org/fullchain.pem','r') error:2006D080:BIO routines:BIO_new_file:no such file)
nginx: [emerg] BIO_new_file("/etc/letsencrypt/live/evalai-test-ec2-frontend.cloudcv.org/fullchain.pem") failed (SSL: error:02001002:system library:fopen:No such file or directory:fopen('/etc/letsencrypt/live/evalai-test-ec2-frontend.cloudcv.org/fullchain.pem','r') error:2006D080:BIO routines:BIO_new_file:no such file)
```

init-letsencrypt.sh
yashdusing: you can read about that here : https://medium.com/@pentacent/nginx-and-lets-encrypt-with-docker-in-less-than-5-minutes-b4b8a60d3a71

yashdusing:
Could you try fixing the domain names for now manually in init-letscrypt & retrying ? also it would be helpful if you could tell me exactly which instruction it failed on

yashdusing:
also in your nginx conf, the folder directory of ssl certs is wrong. It still has example.com

Ram81:
@yashdusing I am trying to setup without ssl. Ideally without ssl certs the deploy-vm.sh shouldn't fail. It should allow deploying on http. I made changes to nginx for http but it seems https isn't optional

yashdusing:
`
ssl_certificate /etc/letsencrypt/live/example.com/fullchain.pem;
ssl_certificate_key /etc/letsencrypt/live/example.com/privkey.pem;
`
to 
`
ssl_certificate /etc/letsencrypt/live/evalai-test-ec2-frontend.cloudcv.org/fullchain.pem;
ssl_certificate_key /etc/letsencrypt/live/evalai-test-ec2-frontend.cloudcv.org/privkey.pem;
`

similarly for nodejs config

yashdusing:
you're still listening on 443 though. For http it is 80 and you would have to make a few more changes I believe

Ram81:
That is the problem. Idea behind using `deploy-vm.sh` was to make it easier for anyone to deploy it on a VM with or without https. The way current code is structured in your PR requires a lot of changes which are not documented as comments or instructions.

yashdusing:
I can add http nginx config later. 
But for now if you want you can message me the ec2 instance IP & you can use my domain & then I can add route53 entry for it. Will that be fine to check functionality ?

Ram81:
That's not necessary, I think its better to change the nginx conf for the domains I created. Just to confirm once, I haven't explicitly created ssl certs for my domain do I have create them manually?

yashdusing:
No. init-letsencrypt will handle the ssl certs. You do not have to do it manually

init-letsencrypt.sh
yashdusing: We're iterating over all domains & including it in domain args. This is specified to certbot when requesting for certs in final call. So, `domain=$DOMAIN_NAME` would be incorrect I believe

docker/vm/nodejs/nginx_vm.conf
yashdusing: Added commments

init-letsencrypt.sh
Ram81: But you can accept array of domain names as a input and assign it to `DOMAIN_NAME`

yashdusing: then how would we differentiate in frontend & backend domains ?
I added a read command for explicitly taking frontend & backend domains as input so now you won't have to fiddle around with the nginx config. You can check that

docker/vm/nodejs/nginx_vm.conf
```diff
@@ -0,0 +1,82 @@
...
+  }
+}
+
+server {
+  listen 443 ssl;
+  server_name BACKEND_DOMAIN;
+  client_max_body_size 400M;
+
+  # SSL certificates generated by init-letsencrypt
+  ssl_certificate /etc/letsencrypt/live/DOMAIN_NAME/fullchain.pem;
```
Ram81, line 28: Why do we have to replace the paths for ssl certs from lets encrypt? Can't this be placed at a fixed location say `/etc/certs/frontend/` and `/etc/certs/backend`?

init-letsencrypt.sh
```diff
@@ -0,0 +1,84 @@
...
+
+if [ ! -e "$data_path/conf/options-ssl-nginx.conf" ] || [ ! -e "$data_path/conf/ssl-dhparams.pem" ]; then
+  echo "### Downloading recommended TLS parameters ..."
+  mkdir -p "$data_path/conf"
+  curl -s https://raw.githubusercontent.com/certbot/certbot/master/certbot-nginx/certbot_nginx/_internal/tls_configs/options-ssl-nginx.conf > "$data_path/conf/options-ssl-nginx.conf"
+  curl -s https://raw.githubusercontent.com/certbot/certbot/master/certbot/certbot/ssl-dhparams.pem > "$data_path/conf/ssl-dhparams.pem"
+fi
+
+echo "### Creating dummy certificate for $domains ..."
+path="/etc/letsencrypt/live/$domains"
```
Ram81, line 39: please change this to a fixed path. We want to help the user do minimal changes to deploy on vm. If we can't use ssl certs from shell script parameters then we should ask them to place the certs at a fixed location that will be easier than asking them to make changes to nginx config

yashdusing: I don't think that is possible 
Certbot places the certs in a new folder named according to the first domain passed.  
Source : https://certbot.eff.org/docs/using.html#where-are-my-certificates

init-letsencrypt.sh
yashdusing: It is being accepted as parameter at the start

init-letsencrypt.sh
yashdusing: Yes, I can. The reason why I set it as 1 was that lets encrypt renews certs only a limited number of times (I think 7). So, for testing purpose we should use this

deploy-vm.sh
yashdusing: Added a check

init-letsencrypt.sh
yashdusing: Changed now. Accepting backend & frontend separately

docker/vm/nodejs/nginx_vm.conf
yashdusing: Already answered in different comment

docker/vm/nodejs/nginx_vm.conf
```diff
@@ -0,0 +1,84 @@
...
+
+server {
+  listen 443 ssl;
+  server_name BACKEND;
+  client_max_body_size 400M;
+
+  # SSL certificates generated by init-letsencrypt
+  # Don't change the FRONTEND. Certbot places certificates in the first domain passed in init-letsencrypt. 
+  # See https://certbot.eff.org/docs/using.html#where-are-my-certificates to know more
+  ssl_certificate /etc/letsencrypt/live/FRONTEND/fullchain.pem;
```
Ram81, line 30: this should be `BACKEND`

docker/vm/nodejs/nginx_vm.conf
```diff
@@ -0,0 +1,84 @@
...
+  ssl_certificate_key /etc/letsencrypt/live/FRONTEND/privkey.pem;
```
Ram81, line 31: this should be `BACKEND`

docker/vm/nodejs/nginx_vm.conf
```diff
@@ -0,0 +1,84 @@
...
+    root /var/www/certbot;
+  }
+}
+
+server {
+  listen 443 ssl;
+  sendfile on;
+  default_type application/octet-stream;
+  
+  server_name FRONTEND;
```
Ram81, line 65: instead of using `FRONTEND` or `BACKEND` use `FRONTEND_DOMAIN` `BACKEND_DOMAIN`

docker/vm/nodejs/nginx_vm.conf
yashdusing: https://certbot.eff.org/docs/using.html#where-are-my-certificates
In the case of creating a SAN certificate with multiple alternative names, $domain is the first domain passed in via -d parameter. Rather than copying, please point your (web) server configuration directly to those files (or create symlinks)

Since Frontend was passed first, all certs are available in frontend

init-letsencrypt.sh
```diff
@@ -0,0 +1,84 @@
+#!/bin/bash
+
+read -p "Enter your domain name (example.com) : " DOMAIN
```
Ram81, line 3: no need to accept domain suffix

docker/vm/nodejs/nginx_vm.conf
Ram81: wait, but we have two domain names. Are certs for both domains at the same location?

yashdusing: Yes. We're using the same certs

docker/vm/nodejs/nginx_vm.conf
yashdusing: Done

yashdusing:
@Ram81 could you please check now ? I added the http config & made the other changes

deploy-vm.sh
```diff
@@ -0,0 +1,88 @@
+#!/bin/bash
+
+
+if ! [[ $(sudo which docker) && $(sudo docker --version) ]]; then   #[ -x "$(command -v docker)" ];
```
deshraj, line 4: Please remove the comment after `then`

deploy-vm.sh
```diff
@@ -0,0 +1,88 @@
+#!/bin/bash
```
deshraj, line 1: Let's move the deploy-vm.sh script to `scripts/ec2_deployment/`

deshraj: Also, can you please change the name of the script to `deploy.sh` only after moving.

deploy-vm.sh
```diff
@@ -0,0 +1,88 @@
...
+
+
+if ! [[ $(sudo which docker) && $(sudo docker --version) ]]; then   #[ -x "$(command -v docker)" ];
+    # Install requirements (docker)
+    echo "### Removing existing docker libs"
```
deshraj, line 6: Not a big fan of echo statements like `echo "### Hello world"`. Can you please change the statements to something like: `echo "Hello world"`

deploy-vm.sh
```diff
@@ -0,0 +1,88 @@
...
+
+    echo "### Installing docker"
+    sudo apt-get install docker-ce docker-ce-cli containerd.io -y
+
+    echo "### Finished installing docker. Check docker version"
+    sudo docker --version
+
+fi
+
+if ! [[ $(sudo which docker-compose) && $(sudo docker-compose --version) ]]; then   #[ -x "$(command -v docker)" ];
```
deshraj, line 35: Same, please remove the comment after `then`

deploy-vm.sh
```diff
@@ -0,0 +1,88 @@
...
+
+fi
+
+touch .env
+
+read -p "Enter Nodejs (frontend) domain (evalai.example.com) : " FRONTEND_DOMAIN
+echo "FRONTEND_DOMAIN=$FRONTEND_DOMAIN" > .env
+
+read -p "Enter Django (backend) domain (evalapi.example.com) : " BACKEND_DOMAIN
+echo "BACKEND_DOMAIN=$BACKEND_DOMAIN" >> .env
```
deshraj, line 52: We don't want to ask user to input both frontend and backend domain name. Ideally, you want to make changes in the nginx configuration so that:
- Django server gets the request when the URL starts with `/api`
- Frontend server gets the requests for all other requested paths

deploy-vm.sh
```diff
@@ -0,0 +1,88 @@
...
+
+DOCKER_COMPOSE_FILE=""
+
+read -p "Do you want to enable automatic HTTPs with certbot ? (y/N) " decision
+if [ "$decision" != "Y" ] && [ "$decision" != "y" ]; then
+
+    # Normal HTTP
+    DOCKER_COMPOSE_FILE="docker-compose-vm-http.yml"
```
deshraj, line 60: I assume that we are only testing this with EC2 ubuntu AMI only. So lets change the name of the file to `docker-compose-vm-http.yml`

deploy-vm.sh
```diff
@@ -0,0 +1,88 @@
...
+
+read -p "Enter Django (backend) domain (evalapi.example.com) : " BACKEND_DOMAIN
+echo "BACKEND_DOMAIN=$BACKEND_DOMAIN" >> .env
+
+DOCKER_COMPOSE_FILE=""
+
+read -p "Do you want to enable automatic HTTPs with certbot ? (y/N) " decision
+if [ "$decision" != "Y" ] && [ "$decision" != "y" ]; then
+
+    # Normal HTTP
```
deshraj, line 59: This comment is not making much sense here. Please remove.

deploy-vm.sh
deshraj: Also, I don't think we should put these two docker-compose files in the root folder since it is already overwhelmed. Let's move these to `scripts/ec2_deployment/`

docker-compose-vm-http.yml
```diff
@@ -0,0 +1,77 @@
...
+    image: pakohan/elasticmq
+    hostname: sqs
+    ports:
+      - 9324:9324
+
+  django:
+    hostname: django
+    env_file:
+      - docker/dev/docker.env
+    image: yashdusing/evalai_django:latest
```
deshraj, line 41: @Ram81 @RishabhJain2018 can one of you create a docker registry for EvalAI and give Yash access to that so that we can put the actual address here?

docker-compose-vm-http.yml
```diff
@@ -0,0 +1,77 @@
...
+    ports:
+      - "8000:8000"
+    depends_on:
+      - db
+      - sqs
+    volumes:
+      - .:/code
+
+  worker:
+    image: yashdusing/evalai_worker:latest
```
deshraj, line 51: Same. Please change the registry to `evalai`

deploy-vm.sh
deshraj: Lets also change the name of the docker compose files to `docker-compose-http.yml` and `docker-compose-https.yml`

docker-compose-vm.yml
```diff
@@ -0,0 +1,93 @@
...
+    image: pakohan/elasticmq
+    hostname: sqs
+    ports:
+      - 9324:9324
+
+  django:
+    hostname: django
+    env_file:
+      - docker/dev/docker.env
+    image: yashdusing/evalai_django:latest
```
deshraj, line 41: Same. Please change the registry to `evalai`

docker-compose-vm.yml
```diff
@@ -0,0 +1,93 @@
...
+    volumes:
+      - /code/node_modules
+      - /code/bower_components
+      - ./certbot/conf:/etc/letsencrypt
+      - ./certbot/www:/var/www/certbot
+    command: "/bin/sh -c 'while :; do sleep 6h & wait $${!}; nginx -s reload; done & nginx -g \"daemon off;\"'"
+
+  certbot:
+    image: certbot/certbot
+    restart: unless-stopped
```
deshraj, line 83: Shouldn't this be `always`?

docker/vm/nodejs/Dockerfile
```diff
@@ -0,0 +1,43 @@
+FROM node:8.11.2 as node
```
deshraj, line 1: Can we reuse the existing `Dockerfile` rather than creating a new one? It feels like a bad design decision since we would have to maintain multiple copies of the Dockerfile.

The way it should work is that we have one image for prod and that is what gets pushed to Docker Hub and which is what is used for the vm deployment as well.

docker-compose-vm.yml
yashdusing: I think the reason for putting 'unless-stopped' might be because certbot is the one responsible for requesting actual certs. If for some reason it fails after requesting actual certs, there might be a case that you might hit the request limit for certs.

docker/vm/nodejs/Dockerfile
RishabhJain2018: What changes do you  need to make for the prod one?

RishabhJain2018 (changes requested):

yashdusing:
For prod version changes :
1. The last line of copying SSL folder needs to be removed.
2. I will have to shift the 'SED' commands to bash & create template files for config.json & nginx configs instead of running 'SED' in place.
3. Moreover, in v2, we will have to necessarily remove the Dockerfile for django, celery, worker because they have ECR image as base image. So, if we're going to make Dockerfiles for django, celery, worker explicitly, it makes sense to make the nodejs version as well.

RishabhJain2018:
I see..Can you please open up an issue for it and link it here. Let's merge this PR and then you can make those enhancements.

yashdusing:
Added to list : https://github.com/Cloud-CV/EvalAI/issues/2968

RishabhJain2018:
Closing this since it is very old and deployment setup is changed.

PR #2860, opened 2020-06-20
[Backend] Add feature to support partial submission evaluation

Ram81:
## Description

This PR adds support to evaluate submissions incrementally. It allows challenge workers to update `LeaderboardData` for a specific `Submission` incrementally. As soon as the worker evaluates 1 or more metrics for submission it can update the leaderboard data incrementally to show all the evaluated metrics as soon as they get computed rather than waiting for all metrics to be computed all together.


apps/jobs/urls.py
```diff
@@ -45,6 +45,11 @@
         views.update_submission,
         name="update_submission",
     ),
+    url(
+        r"^challenge/(?P<challenge_pk>[0-9]+)/update_partially_evaluated_submission/$",
```
RishabhJain2018, line 49: `r"^challenge/(?P<challenge_pk>[0-9]+)/update_partially_evaluated_submission/$",` -> `r"^challenges/(?P<challenge_pk>[0-9]+)/update_partially_evaluated_submission/$",`

Issue #2773, opened 2020-05-03
Tiles in Challenges List page not aligned properly


vkartik97:
- The Observed Behaviour: 
The Challenge Listing is not aligned properly
<img width="1280" alt="Screenshot 2020-05-03 at 4 10 08 PM" src="https://user-images.githubusercontent.com/3920286/80926919-cca33100-8db7-11ea-9f38-a6c384979330.png">
- Steps to reproduce the problem: Open https://evalai.cloudcv.org/web/challenges/list <br>
- OS: macOS Mojave<br>
- Browser: Firefox 75.0<br>
<br><br>
Thanks!

RishabhJain2018:
@Kajol-Kumari Would you like to wok on this?

Kajol-Kumari:
sure @RishabhJain2018 sir.

PR #2861, opened 2020-06-22
[Frontend_v2] Fix challenge card UI

### LINKED ISSUE #2813: Fix the card UI for challenges
The challenge logo is required to be a square
Screen Shot 2020-05-23 at 2 00 15 PM

however, the card display for each challenge uses a rectangular preview which results in the logo getting cropped from top and bottom. 
Screen Shot 2020-05-23 at 2 00 10 PM

It'd be nice to update the card UI so that it shows the full logo instead of a cropped version.

Kajol-Kumari:
**Changes introduced through this PR**

- Fixes #2813 [ Fix the card UI for challenges ]
Now complete image will be displayed in the challenge card.

- Fixes: #2773 [ Tiles in Challenges List page not aligned properly ] for Frontend_v2

**Screenshot:**
Screenshot from 2020-06-22 16-04-50

\cc: @RishabhJain2018


dexter1691:
This seems to mess up the aspect ratio. Ideally, we'd want to keep the aspect ratio the same. My suggestion would be to center the image such that the full image is shown, even though it might mean there is some whitespace on the side

PR #2863, opened 2020-06-22
Send logs from Django, Celery & NodeJS containers to Cloudwatch.

KhalidRmb:
@deshraj @Ram81 @RishabhJain2018 
https://docs.docker.com/config/containers/logging/awslogs/


docker-compose-production.yml
```diff
@@ -9,6 +9,12 @@ services:
       dockerfile: docker/prod/django/Dockerfile
     ports:
       - "8000:8000"
+    logging:
+      driver: awslogs
+      options:
+        awslogs-region: ${AWS_DEFAULT_REGION}
+        awslogs-group: django_production
```
Ram81, line 16: rest of the `awslogs-group` field are in format `"<log_group>"`. This one isn't, can we keep it consistent for all fields.

docker-compose-production.yml
```diff
@@ -9,6 +9,11 @@ services:
       dockerfile: docker/prod/django/Dockerfile
     ports:
       - "8000:8000"
+    logging:
+      driver: awslogs
+      options:
+        awslogs-region: ${AWS_DEFAULT_REGION}
+        awslogs-group: "django_production"
```
RishabhJain2018, line 16: I think we also need to add this option here: https://docs.docker.com/config/containers/logging/awslogs/#awslogs-create-group

Same for all below options

KhalidRmb: I've created those log groups on AWS, since it's not dynamic. We don't need the option here.

KhalidRmb:
@RishabhJain2018 Let's get this merged?

docker-compose-production.yml
RishabhJain2018: But what if someone forks EvalAI and tries to deploy in the production setting or we change our AWS account then we will have to create them manually. What is the issue in adding this option?

KhalidRmb: Right, I didn't realize that. I'll add the option.

RishabhJain2018: Thanks!

PR #2867, opened 2020-06-26
Restart workers on updating Evaluation script or Test annotation.

KhalidRmb:
ToDo:

- [x] Write a new template on Sendgrid to send the notification email, and enter the ID in the settings file.


KhalidRmb:
@deshraj @Ram81 @RishabhJain2018

apps/challenges/aws_utils.py
```diff
@@ -689,17 +689,53 @@ def restart_workers_signal_callback(sender, instance, field_name, **kwargs):
...
+            challenge_url = "https://{}/web/challenges/challenge-page/{}".format(settings.HOSTNAME, challenge.id)
+            challenge_manage_url = "https://{}/web/challenges/challenge-page/{}/manage".format(settings.HOSTNAME, challenge.id)
+
+            if field_name is "test_annotation":
+                file_updated = "Test Annotation"
+            else if field_name is "evaluation_script":
+                file_updated = "Evaluation script"
+
+            template_data = {
+            "CHALLENGE_NAME": challenge.title,
```
Ram81, line 719: fix indentation

apps/challenges/aws_utils.py
```diff
@@ -689,17 +689,53 @@ def restart_workers_signal_callback(sender, instance, field_name, **kwargs):
...
+        count, failures = response["count"], response["failures"]
+
         logger.info(
             "The worker service for challenge {} was restarted, as {} was changed.".format(
-                instance.pk, field_name
+                challenge.pk, field_name
             )
         )
 
+        if (count != 1):
```
Ram81, line 707: what happens in case the challenge has more than 1 workers? can we fetch the total number of workers to compare from models?

apps/challenges/aws_utils.py
```diff
@@ -689,17 +689,53 @@ def restart_workers_signal_callback(sender, instance, field_name, **kwargs):
...
+            logger.warning("Worker(s) for challenge {} couldn't restart! Error: {}".format(challenge.id, failures[0]["message"]))
+        else:
+            challenge_url = "https://{}/web/challenges/challenge-page/{}".format(settings.HOSTNAME, challenge.id)
+            challenge_manage_url = "https://{}/web/challenges/challenge-page/{}/manage".format(settings.HOSTNAME, challenge.id)
+
+            if field_name is "test_annotation":
+                file_updated = "Test Annotation"
+            else if field_name is "evaluation_script":
```
Ram81, line 715: is there a case for multiple field updates simultaneously? do we need to handle that separately?

KhalidRmb: I mentioned this in the channel as well, Rishabh asked to do this later.

apps/challenges/aws_utils.py
KhalidRmb: Here, count refers to the number of challenges actually. It's due to the start_workers method implementation. So, it should be 1 since we're doing this only for 1 challenge.

RishabhJain2018:
@KhalidRmb Please resolve the conflicts here.

RishabhJain2018:
@KhalidRmb Can you please resolve conflicts here?

PR #2868, opened 2020-06-27
[BugFix] Fix dataset split codename in seeding script

Ram81:
### Description

1. Database seeding script had a issue with `DatasetSplit` `codenames` which was causing submission evaluation failures as the codenames were not matching with the values hardcoded in evaluation script for all the seeded challenges. Evaluation script was expecting following splits:

```
[
          {
                'split1': {
                    'score': score,
                }
            },
            {
                'split2': {
                    'score': score,
                }
            }
]
```

but the seeding script had a unique codename for every `DatasetSplit`.

2. For every seeded challenge host was not already added as a participant team. This PR fixes that too.


scripts/seed.py
```diff
@@ -308,7 +334,7 @@ def create_dataset_splits(number_of_splits):
     for i in range(number_of_splits):
```
RishabhJain2018, line 334: Can we please change `i` to something meaningful like `split`

scripts/seed.py
```diff
@@ -308,7 +334,7 @@ def create_dataset_splits(number_of_splits):
...
         global DATASET_SPLIT_ITERATOR
         name = "Split {}".format(DATASET_SPLIT_ITERATOR + 1)
-        codename = "{}{}".format("split", DATASET_SPLIT_ITERATOR + 1)
+        codename = "{}{}".format("split", i + 1)
```
RishabhJain2018, line 337: Can we please change `i` to something meaningful like `split`

Ram81:
@RishabhJain2018  done

Issue #2870, opened 2020-06-27
Check if the cluster is already created for a challenge in database


RishabhJain2018:
Complete details for the error: https://sentry.io/share/issue/1c744d1830cd4ead8f1f234be5524920/

uzair-ali10:
Hey does this issue still persist??

PR #2873, opened 2020-06-27
#2870: Fixed model queries

Ayukha:
Fixes condition when a cluster is already created for a challenge in database   fixes(#2870 )


Ayukha:

apps/challenges/aws_utils.py
```diff
@@ -802,12 +802,21 @@ def create_eks_cluster(sender, challenge, field_name, **kwargs):
...
             config_file = NamedTemporaryFile(delete=True)
             config_file.write(config_text.encode())
-            ChallengeEvaluationCluster.objects.create(
-                challenge=challenge,
-                name=cluster_name,
-                cluster_endpoint=cluster_ep,
-                cluster_ssl=cluster_cert,
-            )
+            try:
+                ChallengeEvaluationCluster.objects.create(
```
RishabhJain2018, line 806: @Ayukha Let's try to get the object first and if it is not found then please create the object by serializing the data using a serializer. Also, the update should happen using the serializer as well.

apps/challenges/aws_utils.py
```diff
@@ -952,12 +953,19 @@ def create_eks_cluster(challenge):
...
             config_file = NamedTemporaryFile(delete=True)
             config_file.write(config_text.encode())
-            ChallengeEvaluationCluster.objects.create(
-                challenge=challenge_obj,
-                name=cluster_name,
-                cluster_endpoint=cluster_ep,
-                cluster_ssl=cluster_cert,
-            )
+            try:
+                cluster_details =  ChallengeEvaluationCluster.objects.get(
```
RishabhJain2018, line 957: @Ayukha If this block gets executed, then where are we using the `cluster_details` variable?

RishabhJain2018:
@Ayukha What is the ETA on this?

apps/challenges/aws_utils.py
```diff
@@ -956,12 +958,19 @@ def create_eks_cluster(challenge):
...
             config_file = NamedTemporaryFile(delete=True)
             config_file.write(config_text.encode())
-            ChallengeEvaluationCluster.objects.create(
-                challenge=challenge_obj,
-                name=cluster_name,
-                cluster_endpoint=cluster_ep,
-                cluster_ssl=cluster_cert,
-            )
+            try:
+                ChallengeEvaluationCluster.objects.get(
```
RishabhJain2018, line 962: Why are we executing this try block?

RishabhJain2018:
@Ayukha ?

apps/challenges/aws_utils.py
Ayukha: @RishabhJain2018  The use case is to create the fields in the model when they don't already exist. And we are not updating anything in the try block because, if a challenge name is changed, the Kubernetes cluster won't be renamed.

PR #2885, opened 2020-07-01
Change awslogs-create-group to string & hardcode region.

KhalidRmb:
@RishabhJain2018


docker/prod/docker_staging.env
```diff
@@ -1,6 +1,6 @@
 AWS_ACCESS_KEY_ID=x
 AWS_ACCOUNT_ID=x
-AWS_DEFAULT_REGION=x
+AWS_DEFAULT_REGION=us-east-1
```
RishabhJain2018, line 3: Please don't change this value.

docker-compose-production.yml
```diff
@@ -14,7 +14,7 @@ services:
       options:
         awslogs-region: ${AWS_DEFAULT_REGION}
```
RishabhJain2018, line 15: Please add "us-east-1" here in place of ${AWS_DEFAULT_REGION}

PR #2891, opened 2020-07-02
Fix template ID & fix signal.

KhalidRmb:


apps/challenges/aws_utils.py
```diff
@@ -260,10 +260,12 @@ def client_token_generator():
     Returns:
     str: string of size 32 composed of digits and letters
     """
-
-    client_token = "".join(
-        random.choices(string.ascii_letters + string.digits, k=32)
+    remaining_chars = 32 - len(str(challenge_pk))
+    s = "".join(
```
RishabhJain2018, line 264: Can we please change `s` to something meaningful like - `random_char_string`

apps/challenges/aws_utils.py
```diff
@@ -260,10 +260,12 @@ def client_token_generator():
...
+        random.choices(string.ascii_letters + string.digits, k=remaining_chars)
     )
+    client_token = str(challenge_pk) + s
```
RishabhJain2018, line 267: Can we please use `f"{str(challenge_pk)}{random_char_string}"`

PR #2906, opened 2020-07-13
Add backend to support uploading large files.

KhalidRmb:
Adds APIs and util methods to enable generation of a presigned url and enable the pipeline to upload large annotation files and submission files through CLI.

Related: https://github.com/Cloud-CV/evalai-cli/pull/277


apps/base/utils.py
```diff
@@ -168,6 +168,28 @@ def get_boto3_client(resource, aws_keys):
         logger.exception(e)
 
 
+def get_presigned_url_for_file_upload(filename, file_key):
+    """
+    """
+    try:
+        aws_keys = {
```
Ram81, line 175: @KhalidRmb aren't these loaded in settings? can't we use those? If not can you create a method in utils and move this piece of code there so that we avoid redundancy

apps/challenges/views.py
```diff
@@ -2705,3 +2706,29 @@ def manage_worker(request, challenge_pk, action):
...
+    key = ""
+    presigned_url = get_presigned_url_for_file_upload(filename, key)
+
+    challenge_phase.test_annotation.path = presigned_url
+    challenge_phase.save()
+
+    response_data = {"presigned_url": presigned_url}
+    return Response(response_data, status=status.HTTP_200_OK)
+
+
```
Ram81, line 2733: remove extra newlines

apps/base/utils.py
```diff
@@ -168,6 +168,28 @@ def get_boto3_client(resource, aws_keys):
...
+        response = s3.generate_presigned_url(
+            ClientMethod='upload_file',
+            Params={'Filename': filename, 'Bucket': settings.AWS_STORAGE_BUCKET_NAME, 'Key': file_key},
+            ExpiresIn=settings.PRESIGNED_URL_EXPIRY,
+        )
+        return response
+    except Exception as e:
+        logger.exception(e)
+        # Handle the error case by sending back a response here?
+
```
Ram81, line 191: what are we doing in case of exception?

apps/challenges/views.py
```diff
@@ -2705,3 +2706,29 @@ def manage_worker(request, challenge_pk, action):
...
+def get_presigned_url_for_annotations(request, challenge_phase_pk):
+    response_data = {
+            "error": "Sorry, you are not authorized for uploading an annotation file."
+        }
+        return Response(response_data, status=status.HTTP_400_BAD_REQUEST)
+
+    challenge_phase = get_challenge_phase_model(challenge_phase_pk)
+
+    filename = ""
+    key = ""
```
Ram81, line 2724: why is the key empty string for every request? is this intentional?

apps/jobs/views.py
```diff
@@ -2073,3 +2074,170 @@ def get_github_badge_data(
...
+        response_data = {
+            "error": "Sorry, cannot accept submissions since challenge phase is not active"
+        }
+        return Response(
+            response_data, status=status.HTTP_406_NOT_ACCEPTABLE
+        )
+
+    # check if user is a challenge host or a participant
+    if not is_user_a_host_of_challenge(request.user, challenge_pk):
+        # check if challenge phase is public and accepting solutions
```
Ram81, line 2121: shouldn't we check if challenge is approved by admin here?

KhalidRmb: Yes...made the change.

apps/challenges/views.py
KhalidRmb: Added the proper values.

apps/base/utils.py
KhalidRmb: I've changed the values to those from settings instead of using env variables.

apps/base/utils.py
KhalidRmb: I'm returning an empty string and sending back a 400 response to the CLI after logging it in our backend.

KhalidRmb:
@RishabhJain2018 @deshraj @Ram81 PTAL.

apps/base/utils.py
```diff
@@ -168,6 +168,29 @@ def get_boto3_client(resource, aws_keys):
...
 
+def get_presigned_url_for_file_upload(file_name, file_key):
+    """
+    Function to get the presigned url to upload a file to s3 with name file_name and key as file_key.
+    """
+    try:
+        aws_keys = {
+            "AWS_ACCESS_KEY_ID": settings.AWS_ACCESS_KEY_ID,
+            "AWS_SECRET_ACCESS_KEY": settings.AWS_SECRET_ACCESS_KEY,
+            "AWS_REGION": os.environ.get("AWS_DEFAULT_REGION", "us-east-1"),
```
Ram81, line 179: @KhalidRmb can you load the region in settings too. Doesn't make sense to do it using env variable here. Also please create a helper method `get_aws_secret_keys` use that to get this `dict`. The method can be used in future

apps/base/utils.py
Ram81: instead of doing that handle exception in API or send a error response from this method. Doesn't make sense to send response on success and empty string on failure. Let's be consistent.

apps/challenges/views.py
```diff
@@ -2705,3 +2707,34 @@ def manage_worker(request, challenge_pk, action):
...
+    challenge_phase = get_challenge_phase_model(challenge_phase_pk)
+
+    file_ext = os.path.splitext(request.data["file_name"])[-1]
+    file_name = "test_annotations/presigned_url_files/challenge_phase_{}/{}{}".format(
+        challenge_phase_pk, uuid.uuid4(), file_ext
+    )
+    file_key = file_name
+
+    presigned_url = get_presigned_url_for_file_upload(file_name, file_key)
+    if presigned_url == "":
```
Ram81, line 2732: please change this to `response['error'] == true` for handling exception case

apps/challenges/views.py
```diff
@@ -2705,3 +2707,34 @@ def manage_worker(request, challenge_pk, action):
...
+        response_data = {"error": "Could not fetch presigned url."}
+        return Response(response_data, status=status.HTTP_400_BAD_REQUEST)
+
+    challenge_phase.test_annotation = file_name
+    challenge_phase.save()
+
+    response_data = {"presigned_url": presigned_url}
+    return Response(response_data, status=status.HTTP_200_OK)
```
Ram81, line 2740: add newline

apps/jobs/views.py
```diff
@@ -2073,3 +2076,157 @@ def get_github_badge_data(
...
+    request, challenge_pk, challenge_phase_pk
+):
+    # Abstract out the validation
+    try:
+        challenge = Challenge.objects.get(pk=challenge_pk)
+    except Challenge.DoesNotExist:
+        response_data = {"error": "Challenge does not exist"}
+        return Response(response_data, status=status.HTTP_400_BAD_REQUEST)
+
+    if not challenge.approved_by_admin:
```
Ram81, line 2095: what about the case when host is making a submission? we should add this check for participants right? not for hosts

apps/jobs/views.py
```diff
@@ -2073,3 +2076,157 @@ def get_github_badge_data(
...
+        submission_message["submission_pk"] = submission.pk
+
+        file_ext = os.path.splitext(request.data["file_name"])[-1]
+        file_name = "submission_files/presigned_url_files/submission_{}/{}{}".format(
+            submission.pk, uuid.uuid4(), file_ext
+        )
+        file_key = file_name
+
+        presigned_url = get_presigned_url_for_file_upload(file_name, file_key)
+        if presigned_url == "":
```
Ram81, line 2209: same here, either handle exception in the api or send a error response dict with error flag from method

apps/jobs/views.py
```diff
@@ -2073,3 +2077,154 @@ def get_github_badge_data(
...
+    request, challenge_pk, challenge_phase_pk
+):
+    # Abstract out the validation
+    try:
+        challenge = Challenge.objects.get(pk=challenge_pk)
+    except Challenge.DoesNotExist:
+        response_data = {"error": "Challenge does not exist"}
+        return Response(response_data, status=status.HTTP_400_BAD_REQUEST)
+
+    if not is_user_a_host_of_challenge(request.user, challenge_pk) and not challenge.approved_by_admin:
```
Ram81, line 2096: @KhalidRmb do the challenge approval check in the code block at line 2125. The check is redundant `if not is_user_a_host_of_challenge(request.user, challenge_pk)`. See the submission view for how to order these checks

apps/challenges/views.py
Ram81: no newline?

apps/base/utils.py
```diff
@@ -168,6 +177,25 @@ def get_boto3_client(resource, aws_keys):
...
+def get_presigned_url_for_file_upload(file_key):
+    """
+    Function to get the presigned url to upload a file to s3 with name file_name and key as file_key.
+    """
+    try:
+        aws_keys = get_aws_secret_keys()
+
+        s3 = get_boto3_client("s3", aws_keys)
+        response = s3.generate_presigned_post(
+            Bucket=settings.AWS_STORAGE_BUCKET_NAME,
```
Ram81, line 189: why are you using the bucket explicitly from settings? shouldn't this also go in `get_aws_secret_keys`

apps/jobs/views.py
```diff
@@ -2073,3 +2077,154 @@ def get_github_badge_data(
...
+
+@api_view(["POST"])
+@throttle_classes([UserRateThrottle])
+@permission_classes((permissions.IsAuthenticated, HasVerifiedEmail))
+@authentication_classes((ExpiringTokenAuthentication,))
+def publish_submission_message_api(request):
+    message = json.loads(request.data.get("submission_message"))
+    publish_submission_message(message)
+    response_data = {"id": message.get("submission_pk")}
+    return Response(response_data, status=status.HTTP_200_OK)
```
Ram81, line 2230: add newline

KhalidRmb: It gives a flake8 error.

apps/jobs/urls.py
```diff
@@ -85,4 +85,14 @@
         views.get_github_badge_data,
         name="get_github_badge_data",
     ),
+    url(
+        r"^challenge/(?P<challenge_pk>[0-9]+)/challenge_phase/(?P<challenge_phase_pk>[0-9]+)/get_presigned_url_for_submission/$",
+        views.get_presigned_url_for_submission,
+        name="get_presigned_url_for_submission",
+    ),
+    url(
+        r"^publish_submission_message/$",
```
Ram81, line 94: @RishabhJain2018  can you confirm if this API url is okay?

apps/jobs/views.py
```diff
@@ -2073,3 +2077,161 @@ def get_github_badge_data(
...
+        return Response(response_data, status=status.HTTP_406_NOT_ACCEPTABLE)
+
+    # check if challenge phase is active
+    if not challenge_phase.is_active:
+        response_data = {
+            "error": "Sorry, cannot accept submissions since challenge phase is not active"
+        }
+        return Response(response_data, status=status.HTTP_406_NOT_ACCEPTABLE)
+
+    # check if user is a challenge host or a participant
```
Ram81, line 2120: change `# check if user is a challenge host or a participant` to `# Check if user is a challenge host or a participant`

apps/jobs/views.py
```diff
@@ -2073,3 +2077,161 @@ def get_github_badge_data(
...
+    if not is_user_a_host_of_challenge(request.user, challenge_pk):
+        # check if challenge phase is public and accepting solutions
```
Ram81, line 2122: change `# check if challenge phase is public and accepting solutions` to `# Check if challenge phase is public and accepting solutions`

apps/jobs/views.py
```diff
@@ -2073,3 +2077,161 @@ def get_github_badge_data(
...
+
+    participant_team_id = get_participant_team_id_of_user_for_a_challenge(
+        request.user, challenge_pk
+    )
+
+    if not challenge.is_active:
+        response_data = {"error": "Challenge is not active"}
+        return Response(response_data, status=status.HTTP_406_NOT_ACCEPTABLE)
+
+    # check if challenge phase is active
```
Ram81, line 2113: change `    # check if challenge phase is active` to ` # Check if challenge phase is active`

apps/jobs/views.py
```diff
@@ -2073,3 +2077,161 @@ def get_github_badge_data(
...
+    request, challenge_pk, challenge_phase_pk
+):
+    # Abstract out the validation
+    try:
+        challenge = Challenge.objects.get(pk=challenge_pk)
+    except Challenge.DoesNotExist:
+        response_data = {"error": "Challenge does not exist"}
+        return Response(response_data, status=status.HTTP_400_BAD_REQUEST)
+
+    # check if the challenge phase exists or not
```
Ram81, line 2096: change `# check if the challenge phase exists or not` to `# Check if the challenge phase exists or not`

apps/jobs/views.py
```diff
@@ -2073,3 +2077,160 @@ def get_github_badge_data(
...
+        }
+        return Response(response_data, status=status.HTTP_201_CREATED)
+    return Response(serializer.errors, status=status.HTTP_406_NOT_ACCEPTABLE)
+
+
+@api_view(["POST"])
+@throttle_classes([UserRateThrottle])
+@permission_classes((permissions.IsAuthenticated, HasVerifiedEmail))
+@authentication_classes((ExpiringTokenAuthentication,))
+def publish_submission_message_api(request):
```
RishabhJain2018, line 2232: `publish_submission_message_api` -> `send_submission_message`

apps/jobs/views.py
```diff
@@ -2073,3 +2077,160 @@ def get_github_badge_data(
...
+    message = json.loads(request.data.get("submission_message"))
```
RishabhJain2018, line 2233: What if the `submission_message` is blank?

apps/jobs/views.py
```diff
@@ -2073,3 +2077,160 @@ def get_github_badge_data(
...
+
+@api_view(["POST"])
+@throttle_classes([UserRateThrottle])
+@permission_classes((permissions.IsAuthenticated, HasVerifiedEmail))
+@authentication_classes((ExpiringTokenAuthentication,))
+def get_presigned_url_for_submission(
+    request, challenge_pk, challenge_phase_pk
+):
+    try:
+        challenge = Challenge.objects.get(pk=challenge_pk)
```
RishabhJain2018, line 2090: Can we use `get_challenge_model`

apps/jobs/views.py
```diff
@@ -2073,3 +2077,160 @@ def get_github_badge_data(
...
+    except Challenge.DoesNotExist:
+        response_data = {"error": "Challenge does not exist"}
+        return Response(response_data, status=status.HTTP_400_BAD_REQUEST)
+
+    # Check if the challenge phase exists or not
+    try:
+        challenge_phase = ChallengePhase.objects.get(
```
RishabhJain2018, line 2097: Can we use `get_challenge_phase_model`

apps/jobs/views.py
```diff
@@ -2073,3 +2077,160 @@ def get_github_badge_data(
...
             data["message"] = f"{challenge_obj.title}"
     return Response(data, status=http_status_code)
+
+
+@api_view(["POST"])
+@throttle_classes([UserRateThrottle])
+@permission_classes((permissions.IsAuthenticated, HasVerifiedEmail))
+@authentication_classes((ExpiringTokenAuthentication,))
+def get_presigned_url_for_submission(
+    request, challenge_pk, challenge_phase_pk
```
RishabhJain2018, line 2087: Why do we need challenge_pk and challenge_phase_pk both? We can simply take and use challenge_phase_pk, no?

apps/jobs/views.py
RishabhJain2018, line 2232: Can we have all the checks of creating a submission by the participant here as well otherwise I can simply write a script to post vague messages into the queue?

apps/jobs/urls.py
```diff
@@ -85,4 +85,14 @@
         views.get_github_badge_data,
         name="get_github_badge_data",
     ),
+    url(
+        r"^challenge/(?P<challenge_pk>[0-9]+)/challenge_phase/(?P<challenge_phase_pk>[0-9]+)/get_presigned_url_for_submission/$",
```
RishabhJain2018, line 89: ` r"^challenge/(?P<challenge_pk>[0-9]+)/challenge_phase/(?P<challenge_phase_pk>[0-9]+)/get_presigned_url_for_submission/$",` ->
`r"^phases/(?P<challenge_phase_pk>[0-9]+)/get_submission_file_presigned_url/$",`

apps/jobs/views.py
```diff
@@ -2073,3 +2077,160 @@ def get_github_badge_data(
         else:
             data["message"] = f"{challenge_obj.title}"
     return Response(data, status=http_status_code)
+
+
+@api_view(["POST"])
+@throttle_classes([UserRateThrottle])
+@permission_classes((permissions.IsAuthenticated, HasVerifiedEmail))
+@authentication_classes((ExpiringTokenAuthentication,))
+def get_presigned_url_for_submission(
```
RishabhJain2018, line 2086: `get_presigned_url_for_submission` -> `get_submission_file_presigned_url`

apps/jobs/urls.py
RishabhJain2018: I don't think it is as it needs to have all the checks and should accept phase_pk parameter.
cc: @Ram81

apps/challenges/views.py
```diff
@@ -2740,3 +2742,36 @@ def manage_worker(request, challenge_pk, action):
...
+    if not is_user_a_host_of_challenge(request.user, challenge_pk):
+        response_data = {
+            "error": "Sorry, you are not authorized for uploading an annotation file."
+        }
+        return Response(response_data, status=status.HTTP_400_BAD_REQUEST)
+
+    challenge_phase = get_challenge_phase_model(challenge_phase_pk)
+
+    file_ext = os.path.splitext(request.data["file_name"])[-1]
+    file_name = "test_annotations/presigned_url_files/challenge_phase_{}/{}{}".format(
```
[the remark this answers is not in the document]
KhalidRmb: We're just hardcoding the logical parent folders of the file. The file name itself is random. This is consistent with the way the ChallengePhase model stores the files. The only additional thing I've specified is a `presigned_url_files` parent folder.
What do you think?

apps/jobs/views.py
KhalidRmb: Yeah, thanks for pointing out. I'll make the change.

apps/base/utils.py
```diff
@@ -168,6 +179,31 @@ def get_boto3_client(resource, aws_keys):
         logger.exception(e)
 
 
+def get_presigned_url_for_file_upload(file_key):
+    """
+    Function to get the presigned url to upload a file to s3 with name file_name and key as file_key.
+    """
+    if settings.DEBUG:
```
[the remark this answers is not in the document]
KhalidRmb: Right, it shouldn't run even then. Will make the change.

apps/jobs/views.py
KhalidRmb: Right...I'll make the change to catch this edge case.

apps/base/utils.py
```diff
@@ -168,6 +179,31 @@ def get_boto3_client(resource, aws_keys):
         logger.exception(e)
 
 
+def get_presigned_url_for_file_upload(file_key):
+    """
+    Function to get the presigned url to upload a file to s3 with name file_name and key as file_key.
```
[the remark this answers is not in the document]
KhalidRmb: It's clear that it's the key for the file on S3 right. What else can we call it... `file_key_on_s3`? cc: @Ram81

apps/base/utils.py
```diff
@@ -138,6 +138,17 @@ def send_email(
     return
 
 
+def get_aws_secret_keys():
```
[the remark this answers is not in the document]
KhalidRmb: Ca you point me to this? Didn't find one in `base/utils.py`

apps/challenges/views.py
```diff
@@ -2740,3 +2742,36 @@ def manage_worker(request, challenge_pk, action):
...
+        challenge_phase_pk, uuid.uuid4(), file_ext
+    )
+    file_key = "{}/{}".format(settings.MEDIAFILES_LOCATION, file_name)
+
+    response = get_presigned_url_for_file_upload(file_key)
+    if response.get("error"):
+        response_data = response
+        return Response(response_data, status=status.HTTP_400_BAD_REQUEST)
+
+    challenge_phase.test_annotation = file_name
```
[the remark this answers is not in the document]
KhalidRmb: Do you mean a separate serializer just for updating the file name?

apps/base/utils.py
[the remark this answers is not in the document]
RishabhJain2018: I think it is in challenge/utils.py

apps/challenges/views.py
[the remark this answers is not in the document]
RishabhJain2018: I still didn't get this or maybe I missing something here -- When you will call the save method would that automatically put the file in the respective folder? Why do we need a hardcoded name?

apps/challenges/views.py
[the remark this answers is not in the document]
RishabhJain2018: Not the submission serializer but updating the fields using the serializer by using `partial=True` as a parameter in serializers.

apps/base/utils.py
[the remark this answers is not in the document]
KhalidRmb: Okay...I've changed the function to get the storage bucket name as well when we are not using host's credentials.
But, do we need to add the bucket name field to the challenge model as well, to get the host's bucket name to the aws_keys when using host credentials? (I can make a separate PR for this.) @RishabhJain2018 @Ram81

apps/challenges/views.py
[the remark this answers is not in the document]
KhalidRmb: It'll automatically put the dummy file in the default folder like so: 
`test_annotation = models.FileField(upload_to=RandomFileName("test_annotations"), default=False)`.

What's being done here is the following:
1. Dummy file is being uploaded automatically like usual (back when they create the challenge.)
2. Now, when uploading through CLI: This API will generate the presigned url to upload the new annotation file to a different `presigned_url` subfolder within the `test_annotation` folder. (I guess I can remove the` challenge_phase_{}`) folder to have just one additional new folder.
3. After this, the file_name for the current dummy file should be changed so that when the worker fetches the file's url, it can point to the new annotation file. (And this is what works to change the name of a file in Django: `challenge_phase.test_annotation = file_name`)

Are you saying upload the actual annotation file using the Presigned to the same location as the dummy file with the same `uuid.uuid4()` name?  (That way there's no need to change the file name in the ChallengePhase object since the new annotation file will be superimposed on the previous dummy file.)

apps/challenges/views.py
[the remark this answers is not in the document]
KhalidRmb: I'll have to see is this works. But, no need to do this if we want the new annotation file to be uploaded to the same location as the dummy file. (Essentially, replace the dummy file.)

apps/base/utils.py
```diff
@@ -138,6 +138,17 @@ def send_email(
...
+    aws_keys = {
+        "AWS_ACCOUNT_ID": os.environ.get("AWS_ACCOUNT_ID"),
```
Ram81, line 143: what if the environment variable is not set? you have to pass a default value too right?

KhalidRmb: @Ram81 I've changed this to how RIshabh suggested. Haven't pushed because I'm waiting for confirmation on some other comments.

RishabhJain2018:
@KhalidRmb Is this complete?
cc: @Ram81

KhalidRmb:
> @KhalidRmb Is this complete?
> cc: @Ram81

@RishabhJain2018 Yes

Ram81:
@RishabhJain2018 I haven't tested the PR yet. Are we planning to test this one staging?

RishabhJain2018:
> @RishabhJain2018 I haven't tested the PR yet. Are we planning to test this one staging?

Yes!

Ram81:
@KhalidRmb did you also test these changes with auto launch workers feature when challenge is approved? Because we will be auto approving challenges and launching workers before challenge approval would this break submission worker launch flow?

KhalidRmb:
> @KhalidRmb did you also test these changes (for test annotations) with auto launch workers feature when challenge is approved? Because we will be auto approving challenges and launching workers before challenge approval would this break submission worker launch flow?

I thought the admin won't approve a challenge without an annotation file. But if we're auto approving a challenge, the worker will fail after starting since there's no file yet to fetch. We'll have to catch this case in the auto approval pipeline, where we should check for annotation file being present before auto approving a challenge?

Ram81:
@KhalidRmb can you resolve conflicts so that we can test this on staging

PR #2907, opened 2020-07-14
[Backend] Worker auto launch without EvalAI admin in loop

Ram81:
## Description

This PR adds support for following features:
```
- Approving of challenge without EvalAI admin in the loop -
    - Override the challenge save method and launch the worker when a challenge is created.
    - Check in the submissions API if the submission is from the challenge host then pass is to the submission worker else doesn't pass.
    - Show to challenge host on the UI that they can make submissions and other participants can't. For that please email EvalAI admin.
```

@RishabhJain2018 can you share the exact message that we should add for point 3?

From my testing I could see that support for point 2 was already there. I was able to make submissions from challenge host account even when challenge was not approved.


apps/challenges/views.py
```diff
@@ -1210,6 +1210,14 @@ def create_challenge_using_zip_file(request, challenge_host_team_pk):
                 }
                 send_slack_notification(message=message)
 
+            response = start_workers([challenge])
+            count, failures = response["count"], response["failures"]
+            logging.info("Total worker start count is {} and failures are: {}".format(count, failures))
+            if count:
+                logging.info("{} workers started successfully".format(count))
```
Ram81, line 1217: @RishabhJain2018 @deshraj  need to know what to do in case of worker startup failure

Ram81:
@RishabhJain2018 please don't merge this. There are few more changes I will add related to celery tasks, didn't get time this week to work on them.

apps/challenges/views.py
```diff
@@ -822,6 +824,7 @@ def create_challenge_using_zip_file(request, challenge_host_team_pk):
         response_data = {"error": message}
         return Response(response_data, status=status.HTTP_406_NOT_ACCEPTABLE)
 
+    is_test_annotation_missing = False
```
RishabhJain2018, line 827: Can we also add the same check for the missing evaluation script?

apps/challenges/views.py
```diff
@@ -1244,6 +1249,28 @@ def create_challenge_using_zip_file(request, challenge_host_team_pk):
                 }
                 send_slack_notification(message=message)
 
+            template_data = get_challenge_template_data(zip_config.challenge)
+            if not is_test_annotation_missing:
```
RishabhJain2018, line 1253: Please also check for missing evaluation script file.

apps/challenges/utils.py
```diff
@@ -323,3 +328,42 @@ def get_unique_alpha_numeric_key(length):
...
+    )
+    template_data = {
+        "CHALLENGE_NAME": challenge.title,
+        "CHALLENGE_URL": challenge_url,
+        "CHALLENGE_MANAGE_URL": challenge_manage_url,
+    }
+    return template_data
+
+
+def send_multiple_emails(emails, template_id, template_data):
```
RishabhJain2018, line 355: send_emails

settings/common.py
```diff
@@ -322,6 +322,8 @@
         "CHALLENGE_APPROVAL_EMAIL": "d-45e0adc0597b4b60bd7c384aa903c488",
         "WORKER_RESTART_EMAIL": "d-3d9a474a5e2b4ac4ad5a45ba9c0b84bd",
         "CLUSTER_CREATION_TEMPLATE": "d-6de90fd760df4a41bb9bff1872eaab82",
+        "MISSING_ANNOTATIONS_EMAIL": "",
```
RishabhJain2018, line 325: Can you please create and add the template IDs here?

settings/common.py
```diff
@@ -322,6 +322,8 @@
...
+        "WORKER_START_EMAIL": "",
```
RishabhJain2018, line 326: Can you please create and add the template IDs here?

RishabhJain2018 (changes requested):

RishabhJain2018:
Also, please fix the Travis build.

apps/challenges/views.py
Ram81: @RishabhJain2018 we won't allow challenge creation without evaluation script right? the feature we added was for updating the evaluation script correct?

RishabhJain2018: > we won't allow challenge creation without evaluation script right?

Yes!

> the feature we added was for updating the evaluation script correct?
I'm missing something here, which feature are you talking about?

RishabhJain2018:
@Ram81 Any updates on this?

apps/challenges/views.py
Ram81: I thought we would allow creating challenge without evaluation_script similar to annotations file that's why you are asking for the check.

Ram81:
@RishabhJain2018 can you tell me what will be the content of the banner? and should we remove the banner we added earlier to inform hosts about challenge approval will be done in 24 hours from challenge overview page? After adding the banner for auto worker launch they seem redundant.

Screen Shot 2020-09-12 at 3 45 54 PM

here's the sendgrid email template. let me know if we should change anything

RishabhJain2018:
This template looks okay for now and we can make changes to it later on SendGrid directly.

RishabhJain2018:
Can you please add the banner thing here so that I can merge this PR.

RishabhJain2018 (approved):

PR #2911, opened 2020-07-17
[Frontend_v2] Leaderboard precision value feature added

Kajol-Kumari:
**Changes introduced**

- Challenge host can now adjust the precision value for the scores displayed on the leaderboard.

GIF:
leaderboard-precison

\cc: @Sanji515 @lunayach @RishabhJain2018 @Shekharrajak


frontend_v2/src/app/components/challenge/challengeleaderboard/challengeleaderboard.component.html
```diff
@@ -27,6 +27,20 @@
...
+        <div class="leaderboard-precision" *ngIf="leaderboard.length>0 && selectedPhaseSplit!=null">
+          <span class="spacer"></span>
+          <span class="fw-regular"> Leaderboard precision value:</span>
+            <div class="rangeslider fa-stack fa-2x">
+              <mat-slider thumbLabel min="0" max="20"  step="1" thumbLabel placeholder="leaderboardPrecisionValue" [(ngModel)]="leaderboardPrecisionValue"
+                            (change)="updateLeaderboardDecimalPrecision($event)">
+              </mat-slider>
+              <span class="fa-stack-2x value" *ngIf="leaderboardPrecisionValue">
+                <span class="val">
+                  {{ leaderboardPrecisionValue }}
```
lunayach, line 39: When the slider is showing precision value, do we also need to show it here again?

Kajol-Kumari: The slider will show the value only if we have focus on it. But yes, we can remove this as it looks cleaner to me that way. @lunayach @Sanji515 Need your views on this.

lunayach: Yes, I think we should remove it. Presence of both seems redundant. @RishabhJain2018 @Sanji515 may also pitch in.

PR #2912, opened 2020-07-17
Auto delete workers on challenge unapproval.

### LINKED ISSUE #2901: Stop and delete the challenge workers on AWS automatically when the challenge is un-approved by admin
**Description**
The challenge worker on Fargate keeps on running even when the challenge is unapproved by the EvalAI Admin.

**Fix**
Ideally, when the challenge is unapproved by the admin then the workers on Fargate should be stopped and deleted.

KhalidRmb:
This is to delete workers automatically when the EvalAI admin unapproves a previously approved challenge.
Fixes #2901 

@RishabhJain2018 @Ram81


apps/challenges/aws_utils.py
```diff
@@ -988,3 +984,13 @@ def challenge_workers_start_notifier(sender, instance, field_name, **kwargs):
                 )
             else:
                 construct_and_send_worker_start_mail(challenge)
+
+        if prev and not curr:
+            response = delete_workers([challenge])
+            count, failures = response["count"], response["failures"]
+            if count != 1:
```
RishabhJain2018, line 991: @KhalidRmb What is this condition checking? Can you please mention in the comment?

What if I have 5 workers for a challenge running and I unapproved a challenge.
- Would all the 5 workers will be deleted?
- Do I need to manually reduce them to 1 and then unapprove the challenge?

KhalidRmb: @RishabhJain2018 It basically checks that the number of successful deletions is same as the number of challenges passed in the array.

Yes, all 5 workers will be deleted. (The implementation is that the workers will be scaled down to 0, and then the service is deleted.)

RishabhJain2018: okay, can you please add it as comment.

PR #2915, opened 2020-07-19
[Frontend_v2] Added logger for dev environment

Kajol-Kumari:
**Changes Introduced**

- `ngx-logger` used for logging to the console in development environment.

- Logging disabled for production environment.

cc: @Shekharrajak @RishabhJain2018 @Sanji515


Kajol-Kumari:
> Thanks! It is looking good. Try to use debug, error, fatal levels accordingly in future, wherever it is needed. Can we log it in some file as well ? So that if there is any error happened then we can check the log file and understand the root cause ?

sure! @Shekharrajak

PR #2918, opened 2020-07-20
Send 2 mails for worker auto restart.

### LINKED ISSUE #2898: Send email to challenge hosts only when the worker reload is complete and successful
**Current Scenario**: Whenever a challenge hosts updates an evaluation script or the test annotation file then the email for reloading is sent to them immediately after the upload of annotations file or the evaluation script file which gives a wrong sense to challenge hosts that the worker is reloaded but actually it is not because it takes 2 mins on AWS to reload it successfully.

**Fix**
1. Send the email to challenge hosts immediately after upload saying that worker reload is started and it will take 2 mins to complete and they will be notified again when it is successfully reloaded.
2. Send another email to challenge hosts when the worker reload is successful on AWS.

KhalidRmb:
For auto reloading the workers, send 2 mails to the hosts:
1. At the time of initiating the restart.
2. When the new workers have been launched.
Fixes #2898 

@RishabhJain2018 @Ram81


apps/challenges/aws_utils.py
```diff
@@ -766,57 +767,110 @@ def restart_workers_signal_callback(sender, instance, field_name, **kwargs):
...
-        count, failures = response["count"], response["failures"]
 
-        logger.info(
-            "The worker service for challenge {} was restarted, as {} was changed.".format(
-                challenge.pk, field_name
-            )
+@app.task
+def restart_workers_on_file_change(field_name, serialized_challenge):
+    """
+    Restarts the challenge workers, and sends email notifications to the hosts.
```
Ram81, line 777: @KhalidRmb please use proper docstrings.

apps/challenges/aws_utils.py
```diff
@@ -766,57 +767,110 @@ def restart_workers_signal_callback(sender, instance, field_name, **kwargs):
...
+    """
+    for obj in serializers.deserialize("json", serialized_challenge):
```
Ram81, line 779: please rename `obj` to something relevant

apps/challenges/aws_utils.py
```diff
@@ -766,57 +767,110 @@ def restart_workers_signal_callback(sender, instance, field_name, **kwargs):
...
+        challenge = obj.object
+    response = restart_workers([challenge])
```
Ram81, line 781: add new line above

apps/challenges/aws_utils.py
```diff
@@ -766,57 +767,110 @@ def restart_workers_signal_callback(sender, instance, field_name, **kwargs):
...
+
```
Ram81, line 782: remove newline

apps/challenges/aws_utils.py
```diff
@@ -766,57 +767,110 @@ def restart_workers_signal_callback(sender, instance, field_name, **kwargs):
...
+    count, failures = response["count"], response["failures"]
+
+    logger.info(
+        "The worker service for challenge {} was restarted, as {} was changed.".format(
```
Ram81, line 786: Message should be `Worker(s) for challenge {} were restarted as {} was changed.`

apps/challenges/aws_utils.py
```diff
@@ -766,57 +767,110 @@ def restart_workers_signal_callback(sender, instance, field_name, **kwargs):
...
-                    sender=settings.CLOUDCV_TEAM_EMAIL,
-                    recipient=email,
-                    template_id=template_id,
-                    template_data=template_data,
+        ecs = get_boto3_client("ecs", aws_keys)
+        queue_name = challenge.queue
+        service_name = "{}_service".format(queue_name)
+
+        # Checking if the new workers are launched and old workers are stopped.
+        for i in range(100):
```
Ram81, line 839: this doesn't seem to be the right way to poll for worker status. Can you look for alternative methods

KhalidRmb: There was another way to use a waiter, but it seems suitable only for checking start and stop workers. The way restart works is more complex since it first starts new workers (thereby doubling the number of workers for a brief period) and then scaling back to normal by deleting the old workers.

Some of the waiters use descibe_service internally to know how many tasks are running, so I did the same approach but I'm checking for the unique combination of parameters which signals us that the workers have fully restarted since the time we initiated it.

RishabhJain2018:
@KhalidRmb @Ram81 Is this ready for review?

RishabhJain2018:
This is WIP from last 26 days?

RishabhJain2018:
@Ram81 @KhalidRmb Can we please merge this first?
cc: @deshraj

Ram81:
@RishabhJain2018 the way polling/ worker status checking is implemented in this PR isn't feasible. I had asked @KhalidRmb to look for alternate ways for this. @KhalidRmb can you update here?

KhalidRmb:
@RishabhJain2018 @deshraj @Ram81 So, Ram and I had a discussion on this where we addressed this point I'd made earlier: https://github.com/Cloud-CV/EvalAI/pull/2918#discussion_r460539020

Ram mainly had concerns regarding the time [this](https://github.com/Cloud-CV/EvalAI/blob/63c8cde1c6f1faf002044261f842546f8d6fc46a/apps/challenges/aws_utils.py#L839) loop may take in worst case is 200 seconds in a celery task. So he wanted to explore using boto3 waiters.

I get this, but we’ll be using more resources even if were able to use Waiters since they loop for worst case 600 seconds as opposed to 200 for mine. 
That said, we can’t use waiters since checking for restart is more complex than what the waiters are there for. (And I'm using the same way the waiters are implemented, but tailored for checking restart workers completion)

So, I don't think there's a better alternative. 

I'll fix the tests on this and push it this week so we can close this. The test is failing because celery is not there in Test environment I think. That's all from my side.

RishabhJain2018:
okay, I think I'm missing some context here, let's discuss it in tomorrow' scrum call.

Ram81:
@KhalidRmb @RishabhJain2018 and I discussed about this feature. The current implementation as a celery task doesn't seem feasible. So we made few modifications to this feature. Instead of polling in a celery task we will show worker status on manage tab. 

We will show list of workers with worker status and you can just move your celery polling code to a polling API. And set polling interval between 5s-10s. Idea is that challenge host will be able to monitor worker status directly in manage tab.

Here's a list of changes:

```
- Show the worker status on the manage tab.
- Disable the start/stop/restart buttons when the status is RESTARTING or STARTING to avoid multiple API calls.
- Don’t send the restart worker email but show a notification when that button is clicked.
```

PR #2920, opened 2020-07-21
#Fix[Bug]: challenge approval posthook fixed

Ayukha:


KhalidRmb:
@Ayukha You need to call `challenge_workers_start_notifier` with the appropriate prams for it to be executed. Maybe that's why the travis is failing.
@RishabhJain2018 It doesn't interfere with [this](https://github.com/Cloud-CV/EvalAI/pull/2912) PR, if that's what you meant.

PR #2930, opened 2020-08-01
[BugFix] Modify SQS queue name generation logic

Ram81:
### Description

For the [Table Recognition challenge](https://evalai.cloudcv.org/web/challenges/challenge-page/673/overview) we observed that sqs queue name was identical for both staging and prod versions of challenge. This caused issues with submission processing, submissions made on prod version were being evaluated by staging evaluation workers which were creating issues. Because challenge title was >= 80 characters existing logic couldn't append random uuid suffix for queue name.

### Fix

1. Use <= 50 characters from challenge title in queue name generation
2. Append challenge pk suffix to ensure uniqueness
3. Use uuid for characters at index >= 50 till < 80


apps/base/utils.py
```diff
@@ -207,12 +207,17 @@ def get_slug(param):
     return slug
 
 
-def get_queue_name(param):
-    queue_name = param.replace(" ", "-").lower()
+def get_queue_name(param, unique_key):
```
RishabhJain2018, line 210: `unique_key` -> `challenge_pk`

apps/base/utils.py
```diff
@@ -207,12 +207,17 @@ def get_slug(param):
...
+    # The max-length for queue-name is 80 in SQS
```
RishabhJain2018, line 211: Can we also add docstrings.

PR #2931, opened 2020-08-02
[Backend] Rerank submissions In Progress on My Submissions page

Ram81:
### Description

Add custom sorting for submissions with status `RUNNING` `SUBMITTED` or `SUBMITTING` for `My submissions` page. Sorting logic (for submitting/submitted/running status challenge submissions -> sort by created_at asc -> return rank of current submission)

@RishabhJain2018 @deshraj

### DIFF
--- a/apps/jobs/utils.py
+++ b/apps/jobs/utils.py
@@ -446,7 +446,7 @@ def reorder_submissions_comparator_to_key(comparator):
         Returns:
             [class] -- key class object for lamdbda
     """
-    class K:
+    class ComparatorToLambdaKey:
         def __init__(self, obj, *args):
             self.obj = obj
 
@@ -467,4 +467,4 @@ def reorder_submissions_comparator_to_key(comparator):
 
         def __ne__(self, other):
             return comparator(self.obj, other.obj) != 0
-    return K
+    return ComparatorToLambdaKey
--- a/apps/jobs/utils.py
+++ b/apps/jobs/utils.py
@@ -438,7 +438,7 @@ def reorder_submissions_comparator(submission1, submission2):
 
 def reorder_submissions_comparator_to_key(comparator):
     """
-        Convert a cmp= function into a key= function'
+        Convert a cmp= function into a key= function for lambda
 
         Arguments:
              comparator {[function]} -- comparator function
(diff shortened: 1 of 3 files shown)


apps/jobs/utils.py
```diff
@@ -413,3 +413,58 @@ def get_leaderboard_data_model(submission_pk, challenge_phase_split_pk):
...
+    """
+        Convert a cmp= function into a key= function'
+
+        Arguments:
+             comparator {[function]} -- comparator function
+
+        Returns:
+            [class] -- key class object for lamdbda
+    """
+    class K:
```
RishabhJain2018, line 449: Can we please add a meaningful name here.

apps/jobs/views.py
```diff
@@ -160,8 +161,12 @@ def challenge_submission(request, challenge_id, challenge_phase_id):
         filtered_submissions = SubmissionFilter(
             request.GET, queryset=submission
         )
+        reordered_submissions = sorted(filtered_submissions.qs,
```
RishabhJain2018, line 164: Can we please add a comment here with the logic.

apps/jobs/utils.py
```diff
@@ -413,3 +413,58 @@ def get_leaderboard_data_model(submission_pk, challenge_phase_split_pk):
...
+        def __init__(self, obj, *args):
+            self.obj = obj
+
+        def __lt__(self, other):
```
RishabhJain2018, line 453: I didn't get what `__lt__` means. It would be nice if we can have some meaningful names.

Ram81: `__lt__` is for less than comparison. I don't think we can change comparator key function names, I'll look at it once

apps/jobs/utils.py
```diff
@@ -413,3 +413,64 @@ def get_leaderboard_data_model(submission_pk, challenge_phase_split_pk):
         challenge_phase_split__pk=challenge_phase_split_pk,
     )
     return leaderboard_data
+
+
+def reorder_submissions_comparator(submission1, submission2):
```
RishabhJain2018, line 418: Minor change here -> `submission_1` and `submission_2`.

RishabhJain2018:
@Ram81 Can you please fix the build here?

PR #2947, opened 2020-08-08
[Backend] Submission data cleanup after processing

Ram81:
### Description

Add submission data cleanup after a submission is processed in the submission worker. We are doing this because after a submission is evaluated the submission data occupies space on the instance. Because submission data cleanup was missing we saw `No space left on device` errors while processing new submissions for LVIS challenge. [Issue Thread](https://cloudcv.slack.com/archives/C2P84STC6/p1596778303064600). (We fixed this by manually cleaning up space on the worker)

### DIFF
--- a/scripts/workers/submission_worker.py
+++ b/scripts/workers/submission_worker.py
@@ -170,12 +170,11 @@ def delete_submission_data_directory(location):
     try:
         shutil.rmtree(location)
     except Exception as e:
-        logger.error(
+        logger.exception(
             "{} Failed to delete submission data directory {}, error {}".format(
                 WORKER_LOGS_PREFIX, location, e
             )
         )
-        traceback.print_exc()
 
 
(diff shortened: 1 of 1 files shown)


scripts/workers/submission_worker.py
```diff
@@ -161,6 +160,24 @@ def delete_zip_file(download_location):
...
+    """
+        Helper function to delete submission data from location `location`
+
+        Arguments:
+            location {[string]} -- Location of directory to be removed.
+    """
+    try:
+        shutil.rmtree(location)
+    except Exception as e:
+        logger.error(
```
deshraj, line 173: Instead of logger.error, you can directly use logger.exception and that will automatically print the message that you pass in followed by the traceback. Then you can also get rid of the `traceback.print_exc()`

Ram81:
@RishabhJain2018 @deshraj PTAL

PR #2955, opened 2020-08-14
Feature: script to deploy calico into the cluster

Ayukha:
**Why do we need it?**
To limit the out going request calls from the cluster

**Changes Proposed:**
Calico is being deployed into the cluster to provide networking and network policy.
Deploying AWS container-insights


docker/kubernetes.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+sudo apt install -y awscli  # Install the AWS CLI.
+aws --version               # Confirm the AWS CLI was installed.
+
+aws configure set a us-east-1
+aws configure set aws_access-key.id
+aws configure set aws_secret_access_key.key
+
+#install iam-authenticator
+curl -o aws-iam-authenticator https://amazon-eks.s3.us-west-2.amazonaws.com/1.17.7/2020-07-08/bin/linux/amd64/aws-iam-authenticator
```
RishabhJain2018, line 10: Why do we need aws-iam-authenticator when we have the config file?

Ayukha: @RishabhJain2018 Config uses `get token` command from `iam-authenticator`

docker/kubernetes.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
```
RishabhJain2018, line 1: Can we please keep this script in the worker folder itself. Basically at --
`scripts/workers/code_upload_worker_utils/deploy_calico.sh`

RishabhJain2018:
Also, @Ayukha Can you please update the PR description here.

PR #2956, opened 2020-08-14
Feature: added resource limits for the pods to specify the no of GPUs used

Ayukha:
**Why do we need it?**
Since we want every submission should run on 1 GPU as running parallel submissions on a GPU causes the submissions to run slow and they time out most of the time and multiple pods should not share the same GPU while running.

**Changes Proposed:**
Adding a `resources` parameter while deploying the pods should fix this.


RishabhJain2018:
@Ayukha Can you please fix the Travis build?

PR #2959, opened 2020-08-16
FIX: Prod and staging env added for code-upload worker pipeline

Ayukha:
Added the missiing environment keys used by code-upload-worker pipeline


Ayukha:

PR #2961, opened 2020-08-18
[Backend] Handle challenge not found exception in submission worker

Ram81:
### Description

This PR handles gracefully handles the exception at https://github.com/Cloud-CV/EvalAI/blob/master/scripts/workers/submission_worker.py#L693. As we now have `Start worker` feature available feature in `Manage Tab` whenever a challenge host tries to start a challenge worker before it is approved the above line fails and raises an exception. This has caused our sentry to be flooded with exceptions. Once auto-launch challenge worker before approval feature is done we will modify the query params used in submission worker which will allow us to start workers even when the challenge is not approved. This fix is to reduce messages being sent to sentry

### DIFF
--- a/scripts/workers/submission_worker.py
+++ b/scripts/workers/submission_worker.py
@@ -688,13 +688,7 @@ def get_or_create_sqs_queue(queue_name):
 
 
 def load_challenge_and_return_max_submissions(q_params):
-    try:
-        challenge = Challenge.objects.get(**q_params)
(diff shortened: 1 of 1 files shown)


burnerlee:
@Ram81 i would like to work on this issue, could you please assign it to me and give some more insights on the task please

RishabhJain2018:
@Ram81 Should we close this PR?

PR #2964, opened 2020-08-20
Fix: Added Auth_token value for every challenge in the code-upload pipeline

Ayukha:
**Why do we need it?**
Since every challenge needs a different host token to run the code-upload worker

**Changes Proposed:**
Removed Auth_token from env variable and fetched it from the database


apps/challenges/aws_utils.py
```diff
@@ -343,11 +343,18 @@ def register_task_def_by_challenge_pk(client, queue_name, challenge):
 
     if execution_role_arn:
         if challenge.is_docker_based:
+            # host auth token used by Code-upload-worker
```
RishabhJain2018, line 346: `# challenge host auth token to be used by code-upload-worker`

RishabhJain2018:
Please add the PR description as well.

Ayukha:
@RishabhJain2018 Done

PR #2965, opened 2020-08-20
Add backend for GitHub challenges

KhalidRmb:
This PR includes the backend changes to enable GitHub based challenges.

Changes: 

- Refactor Abstract out and make relevant changes to Challenge config validation API
- Add new API for GitHub challenges MVP, and a field in Challenge model for storing the github_repository identifier

Screenshot 2020-08-20 at 9 36 43 PM


apps/challenges/models.py
```diff
@@ -132,6 +132,8 @@ def __init__(self, *args, **kwargs):
         null=True, blank=True, max_length=2048, default=""
     )
     slack_webhook_url = models.URLField(max_length=200, blank=True, null=True)
+    # Identifier for the github repository of a challenge in format: account_name/repository_name
+    github_repository = models.CharField(max_length=1000, null=True, blank=True, default="")
```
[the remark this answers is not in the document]
KhalidRmb: No...this is the correct name. Just a combination of account name and repo name. It's not a url. I'm using the same name as that in GitHub actions docs.

apps/challenges/views.py
```diff
@@ -2808,3 +2543,366 @@ def get_annotation_file_presigned_url(request, challenge_phase_pk):
...
+    error_messages, yaml_file_data, files = validate_challenge_config_method(
+        request,
+        validate_challenge_config_method,
+        BASE_LOCATION,
+        unique_folder_name,
+        zip_ref,
+    )
+
+    if not len(error_messages):
+        if not challenge_queryset:
```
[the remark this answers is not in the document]
KhalidRmb: An empty queryset gives false on bool..so it works

apps/challenges/views.py
```diff
@@ -2808,3 +2543,366 @@ def get_annotation_file_presigned_url(request, challenge_phase_pk):
...
+            try:
+                with transaction.atomic():
+                    uploaded_zip_file = challenge_config_serializer.save()
+
+                    yaml_file_data["github_repository"] = data[
```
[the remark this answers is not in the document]
KhalidRmb: I'm moving this to happen in serializer. The repo name data is sent as part of request.data. No change required in the config file.

apps/challenges/challenge_config_utils.py
```diff
@@ -112,14 +135,11 @@ def is_challenge_phase_config_yaml_html_field_valid(yaml_file_data, key, base_lo
     message = ""
     is_valid = False
     if value:
-        file_path = join(base_location, value)
-        if file_path.endswith(".html") and isfile(file_path):
-            is_valid = True
-        else:
-            message = (" ERROR: There is no html file present for key"
-                       " {} in phase {}.".format(key, yaml_file_data["name"]))
+        is_valid = True
```
Ram81, line 138: @KhalidRmb why is this check removed?

apps/challenges/challenge_config_utils.py
```diff
@@ -177,3 +198,389 @@ def is_challenge_phase_split_mapping_valid(phase_ids, leaderboard_ids, dataset_s
         else:
             return False
     return False
+
+
+def get_value_from_field(data, base_location, field_name):
+    file_path = join(base_location, data.get(field_name))
```
Ram81, line 204: what about exception handling for missing values for a field name?

Ram81 (changes requested):

apps/challenges/challenge_config_utils.py
KhalidRmb: Because the validation inside the challenge create API is such that it allows an invalid value for the file. It just replaces it with null before moving on. Validation fails if the entire key is missing. @Ram81

apps/challenges/challenge_config_utils.py
KhalidRmb: If the value is missing, this function will return None. In accordance with the validation. (This is also what happens in the create challenge API)

apps/challenges/challenge_config_utils.py
RishabhJain2018: I didn't get the reason for the removal of this check. Can we please discuss about this on slack channel.

apps/challenges/models.py
```diff
@@ -196,6 +200,7 @@ def create_eks_cluster_for_challenge(sender, instance, created, **kwargs):
 class DatasetSplit(TimeStampedModel):
     name = models.CharField(max_length=100)
     codename = models.CharField(max_length=100)
+    config_id = models.IntegerField(default=None, blank=True, null=True)
```
RishabhJain2018, line 203: Can we please add comments for these fields answering why are they needed.

apps/challenges/models.py
```diff
@@ -263,6 +268,7 @@ def __init__(self, *args, **kwargs):
     is_partial_submission_evaluation_enabled = models.BooleanField(
         default=False
     )
+    config_id = models.IntegerField(default=None, blank=True, null=True)
```
RishabhJain2018, line 271: Same as above

apps/challenges/models.py
```diff
@@ -317,6 +323,7 @@ def save(self, *args, **kwargs):
 class Leaderboard(TimeStampedModel):
 
     schema = JSONField()
+    config_id = models.IntegerField(default=None, blank=True, null=True)
```
RishabhJain2018, line 326: Same as above.

apps/challenges/views.py
```diff
@@ -2808,3 +2560,421 @@ def get_annotation_file_presigned_url(request, challenge_phase_pk):
...
     response_data = {"presigned_url": response.get("presigned_url")}
     return Response(response_data, status=status.HTTP_200_OK)
+
+
+@api_view(["POST"])
+@throttle_classes([UserRateThrottle])
+@permission_classes((permissions.IsAuthenticated, HasVerifiedEmail))
+@authentication_classes((ExpiringTokenAuthentication,))
+def create_or_update_github_challenge(request, challenge_host_team_pk):
+    challenge_queryset = Challenge.objects.filter(
```
RishabhJain2018, line 2570: Can we please add check to verify who can use this API?

RishabhJain2018: It should be accessed by hosts only.

RishabhJain2018 (changes requested):

apps/challenges/views.py
```diff
@@ -2808,3 +2560,421 @@ def get_annotation_file_presigned_url(request, challenge_phase_pk):
...
+                        ] = serializer.instance.pk
+
+                    # Create Challenge Phase
+                    challenge_phase_ids = {}
+                    challenge_phases_data = yaml_file_data["challenge_phases"]
+                    for data, challenge_test_annotation_file in zip(
+                        challenge_phases_data,
+                        files["challenge_test_annotation_files"],
+                    ):
+                        data["slug"] = "{}-{}-{}".format(
```
Ram81, line 2671: isn't there a util method to get slug?

Ram81 (changes requested):

apps/challenges/views.py
KhalidRmb: That's different. That just replaces empty spaces with '-', and is used only for challenge. I can use it here only to modify the codename.

KhalidRmb:
@RishabhJain2018 
Done

RishabhJain2018 (approved):

RishabhJain2018:
@KhalidRmb Can you please resolve the conflicts here?

KhalidRmb:
@RishabhJain2018 Done

RishabhJain2018 (approved):

PR #2972, opened 2020-08-26
Feature: Block outgoing connections from pods

Ayukha:
**Why do we need it?**
To limit the out going request calls from the cluster

**Changes Proposed:**
Setting the network policies

- [X] Deny all outgiong requests

- [X] Allow the requests only going to evalai CIDR


docker/prod/docker_production.env
```diff
@@ -36,6 +36,8 @@ SECRET_KEY=some-secret-key
 SENTRY_URL=x
 SENDGRID_API_KEY=x
 
+CIDR=X
```
RishabhJain2018, line 39: `CIDR=x`

docker/prod/docker_staging.env
```diff
@@ -36,6 +36,8 @@ SECRET_KEY=some-secret-key
 SENTRY_URL=x
 SENDGRID_API_KEY=x
 
+CIDR=X
```
RishabhJain2018, line 39: `CIDR=x`

Ayukha: @RishabhJain2018  Updated

PR #2974, opened 2020-08-26
Feature: Export logs for agent and environment containers in kubernetes to s3

Ayukha:
- [X] Added a post-hook for `submission status = Finished`

- [X] Added a method `export_cloudwatch_logs()` as a background task

- [X] Created a export job for agent and environment container


apps/challenges/aws_utils.py
```diff
@@ -1126,3 +1127,40 @@ def challenge_approval_callback(sender, instance, field_name, **kwargs):
...
+def export_cloudwatch_logs(submission):
+    """ 
+    This is to export cloudwatch logs of Agent and environment for a submission into S3 bucket.
+    """
+
+    for obj in serializers.deserialize("json", submission):
+        submission_obj = obj.object
+
+    try:
+        challenge = get_challenge_model(submission.challenge_phase.challenge.id)
```
RishabhJain2018, line 1141: I don't think you need `try/except` while using the `get_challenge_model` function. Please take a look at its definition.

apps/challenges/aws_utils.py
```diff
@@ -1126,3 +1127,40 @@ def challenge_approval_callback(sender, instance, field_name, **kwargs):
...
+                logStreamNamePrefix=log_stream_prefix,
+                fromTime=submission_obj.started_at,
+                to=submission_obj.completed_at,
+                destination=settings.AWS_STORAGE_BUCKET_NAME,
+                destinationPrefix=destination_prefix
+            )
+
+        except ClientError as e:
+            logger.exception(e)
+            return response
```
RishabhJain2018, line 1166: This return should be outside of except block. 
Also, what happens when except block is executed, what is the value of `response` in that case?

RishabhJain2018 (changes requested):

apps/challenges/aws_utils.py
```diff
@@ -1126,3 +1127,40 @@ def challenge_approval_callback(sender, instance, field_name, **kwargs):
                         challenge.id, failures[0]["message"]
                     )
                 )
+
+@app.task
+def export_cloudwatch_logs(submission):
+    """ 
+    This is to export cloudwatch logs of Agent and environment for a submission into S3 bucket.
+    """
```
RishabhJain2018, line 1135: Can you please write proper docstrings with arguments and return type.

RishabhJain2018 (changes requested):

RishabhJain2018:
@Ayukha Can you please resolve the conflicts as well.

PR #2992, opened 2020-09-10
Remove code upload worker from Dev setup

RishabhJain2018:
This PR --
- [x] Remove code-upload-worker container creation in dev setup.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -37,15 +37,16 @@ services:
     volumes:
       - .:/code
 
-  code-upload-worker:
-    hostname: rl-worker
-    env_file:
-      - docker/dev/docker_code_upload_worker.env
-    build:
-      context: ./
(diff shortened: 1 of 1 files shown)


Ram81 (approved):

docker-compose.yml
```diff
@@ -37,15 +37,16 @@ services:
...
-  code-upload-worker:
-    hostname: rl-worker
-    env_file:
-      - docker/dev/docker_code_upload_worker.env
-    build:
-      context: ./
-      dockerfile: docker/dev/code-upload-worker/Dockerfile
-    depends_on:
-      - django
+  # Since we don't use code-upload-worker in dev setup, hence commenting this code
```
deshraj, line 40: Umm, I am not sure why we are keeping the commented code in master branch. Ideally if it is not getting used, then we should remove it. We can always revery it back since all of it is version controlled.

PR #2995, opened 2020-09-11
Fix for the template zip file url on staging and prod

KhalidRmb:
Fix for the template zip file url on staging and prod
- Do not append host url on staging and prod

### DIFF
--- a/apps/challenges/views.py
+++ b/apps/challenges/views.py
@@ -728,7 +728,10 @@ def create_challenge_using_zip_file(request, challenge_host_team_pk):
                 response_data, status=status.HTTP_406_NOT_ACCEPTABLE
             )
 
-        template_zip_s3_url = settings.API_HOST_URL + challenge_template.template_file.url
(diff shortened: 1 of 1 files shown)


apps/challenges/views.py
```diff
@@ -728,7 +728,10 @@ def create_challenge_using_zip_file(request, challenge_host_team_pk):
                 response_data, status=status.HTTP_406_NOT_ACCEPTABLE
             )
 
-        template_zip_s3_url = settings.API_HOST_URL + challenge_template.template_file.url
+        if settings.DEBUG:
```
RishabhJain2018, line 731: What about settings.TEST?

RishabhJain2018 (changes requested):

apps/challenges/views.py
KhalidRmb: RIght! Pushed the fix

RishabhJain2018 (approved):

PR #3008, opened 2020-09-19
[Backend]: Add default worker config in Challenge model

Ram81:
### Description

This PR adds default worker cpu core and memory config in Challenge model and removes the environment variables.


apps/challenges/models.py
```diff
@@ -137,9 +137,9 @@ def __init__(self, *args, **kwargs):
         max_length=1000, null=True, blank=True, default=""
     )
     # The number of vCPU for a Fargate worker for the challenge.
-    worker_cpu_cores = models.IntegerField(null=True, blank=True, default=None)
+    worker_cpu_cores = models.IntegerField(null=True, blank=True, default=256)
```
RishabhJain2018, line 140: Can you please mention in the comments what is 256?

apps/challenges/models.py
```diff
@@ -137,9 +137,9 @@ def __init__(self, *args, **kwargs):
...
     # Memory size of a Fargate worker for the challenge.
-    worker_memory = models.IntegerField(null=True, blank=True, default=None)
+    worker_memory = models.IntegerField(null=True, blank=True, default=512)
```
RishabhJain2018, line 142: same as above.

RishabhJain2018 (changes requested):

RishabhJain2018 (approved):

PR #3021, opened 2020-10-03
Add missing file url check in submission API

Ram81:
### Description

This PR adds the missing `file_url` check in submission api. Currently, we are not handling this case because of which the submission API throws an exception. Refer [this](https://sentry.io/organizations/cloudcv/issues/1929968923/?project=152529&referrer=slack).


RishabhJain2018 (commented):
Can we please add tests for this as well.

RishabhJain2018 (approved):

PR #3030, opened 2020-10-12
Handle submission public/private corner case for restrict one LB submission

Ram81:
### Description

This PR handles the case when a participant chooses public/private submission visibility for challenge phases with `is_restricted_to_select_one_submission=True`. If a participant has 1 submission already public in such phases and tries to make another public submission we will make latest submission public


apps/jobs/views.py
```diff
@@ -304,6 +304,34 @@ def challenge_submission(request, challenge_id, challenge_phase_id):
             request.data["is_public"] = (
                 True if challenge_phase.is_submission_public else False
             )
+        else:
+            request.data["is_public"] = json.loads(request.data["is_public"])
+            if request.data.get("is_public") and challenge_phase.is_restricted_to_select_one_submission:
+                # handle corner case for restrict one public lb submission
+                try:
```
RishabhJain2018, line 311: Why do we need `try/except` here with a blank except? I might be missing something here.

RishabhJain2018 (changes requested):

apps/jobs/views.py
Ram81: It was needed when using `.get` now it isn't

RishabhJain2018:
@Ram81 Can you please fix the travis build?

RishabhJain2018 (approved):

PR #3038, opened 2020-10-15
Disable manage tab for remote evaluation challenge

Ram81:
### Description

This PR disables manage tab access for remote evaluation challenges. Earlier we were spinning up workers on ECS for remote evaluation challenges too, this PR disables that as remote challenges have their own setup for challenge workers.

### DIFF
--- a/apps/challenges/serializers.py
+++ b/apps/challenges/serializers.py
@@ -59,6 +59,7 @@ class ChallengeSerializer(serializers.ModelSerializer):
             "slug",
             "max_docker_image_size",
             "cli_version",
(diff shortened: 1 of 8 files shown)


RishabhJain2018 (changes requested):
I think not displaying the `Manage` Tab would make more sense for challenge hosts in remote challenges.

Ram81:
@RishabhJain2018 that's what I did for old UI and for new UI I don't display manage tab features in settings tab

RishabhJain2018 (approved):

PR #3041, opened 2020-10-16
Add visibility control for default meta attributes while creation

Ram81:
### Description

This PR adds feature to allow challenge hosts change the visibility of default submission meta attributes `method_name`, `method_description` , `project_url` `publication_url`. EvalAI-starters [PR](https://github.com/Cloud-CV/EvalAI-Starters/pull/29)

### Approach

1. We allow challenge hosts to specify the visibility of default submission meta attributes for every challenge phase in the challenge config.

2. If the challenge host has set visibility for any of the default submission meta attributes we use the provided visibility option.

3. If the challenge host doesn't provide visibility options for any of the default submission meta attributes we set default visibility to `True`


RishabhJain2018:
@Ram81 Can you please fix the Travis build and conflicts.

apps/challenges/models.py
```diff
@@ -277,6 +277,8 @@ def __init__(self, *args, **kwargs):
     )
     # Id in the challenge config file. Needed to map the object to the value in the config file while updating through Github
     config_id = models.IntegerField(default=None, blank=True, null=True)
+    # Store the metadata for the default submission meta attributes of this challenge phase.
+    default_submission_meta_attributes = JSONField(default=None, blank=True, null=True)
```
RishabhJain2018, line 281: `# Store the metadata for the default submission meta attributes of this challenge phase.` -> `# Store the default metadata for a submission meta attributes of a challenge phase.`

RishabhJain2018, line 281: Can we please add the default value so that the current running challenges are not facing any issues as for current challenges, this field will be None and no fields will get displayed.

apps/challenges/views.py
```diff
@@ -1002,6 +1002,31 @@ def create_challenge_using_zip_file(request, challenge_host_team_pk):
                         response_data, status=status.HTTP_406_NOT_ACCEPTABLE
                     )
 
+        if data.get("default_submission_meta_attributes"):
+            for attribute in data["default_submission_meta_attributes"]:
+                keys = ["name", "visible"]
```
RishabhJain2018, line 1007: `visible` -> `is_visible`

RishabhJain2018 (changes requested):

apps/challenges/models.py
Ram81: @RishabhJain2018 default value is null which is already handled on UI side.

RishabhJain2018 (dismissed):

RishabhJain2018:
@Ram81 Can you please resolve the conflicts here.

RishabhJain2018 (approved):

RishabhJain2018 (changes requested):
@Ram81 Can you please write the approach of how we are doing this in the description. This will be a good way to keep track of how we implemented things rather than going through the codebase everytime.

RishabhJain2018 (approved):

PR #3047, opened 2020-10-22
Accept npz file afor submission input

Ram81:
### Description

Allow `.npz` input files in submissions.

### DIFF
--- a/frontend/src/views/web/challenge/submission.html
+++ b/frontend/src/views/web/challenge/submission.html
@@ -227,7 +227,7 @@
                                         <span>Upload file</span>
                                         <input type="file" ngf-select ng-model="challenge.input_file"
                                             name="challenge.input_file"
-                                            accept=".json, .zip, .txt, .tsv, .gz, .csv, .h5, .npy">
+                                            accept=".json, .zip, .txt, .tsv, .gz, .csv, .h5, .npy, .npz">
                                     </div>
                                     <div class="file-path-wrapper">
                                         <input class="file-path validate" type="text" readonly>


frontend/src/views/web/challenge/submission.html
```diff
@@ -227,7 +227,7 @@ <h5 class="w-300">Make Submission</h5>
                                         <span>Upload file</span>
                                         <input type="file" ngf-select ng-model="challenge.input_file"
                                             name="challenge.input_file"
-                                            accept=".json, .zip, .txt, .tsv, .gz, .csv, .h5, .npy">
+                                            accept=".json, .zip, .txt, .tsv, .gz, .csv, .h5, .npy, .npz">
```
RishabhJain2018, line 230: There is a field in the backend using which we can control it. Can you please change it to that?

RishabhJain2018 (changes requested):

apps/challenges/models.py
```diff
@@ -265,7 +265,7 @@ def __init__(self, *args, **kwargs):
         max_length=2128, null=True, blank=True
     )  # Max length of repository name and tag is 2000 and 128 respectively
     allowed_submission_file_types = models.CharField(
-        max_length=200, default=".json, .zip, .txt, .tsv, .gz, .csv, .h5, .npy"
+        max_length=200, default=".json, .zip, .txt, .tsv, .gz, .csv, .h5, .npy, .npz"
```
RishabhJain2018, line 268: Why are we changing the default value?

RishabhJain2018: We can change the database value of this particular challenge, no?

RishabhJain2018 (changes requested):

apps/challenges/models.py
Ram81: Sure, I thought you meant we should add `npz` to default values too

RishabhJain2018: No no, the whole point of having a database field is to have different values for different challenges.

frontend/src/views/web/challenge/submission.html
Ram81: @RishabhJain2018 We want to add `.npz` manually for all phases of open catalyst challenge right?

RishabhJain2018: Yes, but in `accept` we want that fields should come from backend rather than being hardcoded as it is now.

RishabhJain2018 (approved):

PR #3049, opened 2020-10-23
Use allowed submission type field from backend

Ram81:
### Description

This PR adds changes to use `allowed_submission_types` field value for file upload restriction on submit page. EvalAI-Starters [PR](https://github.com/Cloud-CV/EvalAI-Starters/pull/30)

### DIFF
--- a/apps/challenges/serializers.py
+++ b/apps/challenges/serializers.py
@@ -96,6 +96,7 @@ class ChallengePhaseSerializer(serializers.ModelSerializer):
             "is_restricted_to_select_one_submission",
             "submission_meta_attributes",
             "is_partial_submission_evaluation_enabled",
+            "allowed_submission_file_types",
         )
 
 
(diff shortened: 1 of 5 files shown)


RishabhJain2018 (approved):

RishabhJain2018 (changes requested):
@Ram81 Can you please add the corresponding change in challenge create API.

RishabhJain2018:
@Ram81 We also missed the change in uploading files through URLs. Can you please add that as well.

Ram81:
>@Ram81 Can you please add the corresponding change in challenge create API.

Please see the serializer changes this is already done.

>@Ram81 We also missed the change in uploading files through URLs. Can you please add that as well.

Done

RishabhJain2018:
>Please see the serializer changes this is already done.

Do we not need to parse the values from the challenge YAML in the challenge creation API or am I missing something here?

Ram81:
>Do we not need to parse the values from the challenge YAML in the challenge creation API or am I missing something here?

Once you expose the field in the `ChallengePhaseCreateSerializer` it will initialize the serializer values from yaml data.

RishabhJain2018 (approved):

PR #3058, opened 2020-10-31
Add download submission presigned url API

Ram81:
### Description

This PR

- [x] Add presigned URL generation API for submission file download.


Ram81:
@RishabhJain2018 how do we want to authenticate this API request? If we use user token then each worker will need an auth token

RishabhJain2018:
>@RishabhJain2018 how do we want to authenticate this API request? If we use user token then each worker will need an auth token

The ideal way would be to use challenge creators auth token but for that, we need to add an API token package (we discussed that earlier). Let's not block this feature with that.

Ram81:
@RishabhJain2018 then how do we handle authentication? If we remove authentication then anyone with submission id can get a presigned url. Or do you want to add submission file name as a request data param to make sure no one can get access to the file without the uuid?

PR #3062, opened 2020-11-02
Deployment monitoring script

Ram81:
### Description

This PR adds
- [x] Script to monitor status of staging and production docker container. If any container goes down the script will send a notification to `#evalai-deployment-notifications`

### Crontab entry

```
0 */1 * * * /home/ubuntu/evalai/scripts/monitoring/run_monitor_containers.sh
```

### DIFF
--- a/scripts/monitoring/run_monitor_containers.sh
+++ b/scripts/monitoring/run_monitor_containers.sh
@@ -1,2 +1,2 @@
 #!/bin/bash
-python scripts/monitoring/monitor_containers.py
+python /home/ubuntu/evalai/scripts/monitoring/monitor_containers.py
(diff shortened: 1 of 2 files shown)


RishabhJain2018 (approved):

scripts/monitoring/run_monitor_containers.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+python /home/ubuntu/evalai/scripts/monitoring/monitor_containers.py
```
RishabhJain2018, line 2: Can we please use the bash command to get the absolute path for this folder and then append it with the relative path to this file?

RishabhJain2018 (changes requested):

scripts/monitoring/run_monitor_containers.sh
Ram81: @RishabhJain2018  done

RishabhJain2018 (approved):

PR #3082, opened 2020-11-21
Migration to Django 2.2

Ram81:
### Description

This PR adds:

- [x] Upgrade `django==1.11.3` to `django=2.2.8`
- [x] Add respective models changes (specify `on_delete` for `ForeignKey` fields, refer [this](https://docs.djangoproject.com/en/3.1/ref/models/fields/))
- [x] Change `user.is_anonymous()` to `user.is_anonymous`, refer [this](https://docs.djangoproject.com/en/2.2/ref/contrib/auth/)
- [x] Upgrade `drf-yasg` `pyyaml` `django-allauth` `djangorestframework` `django-silk` `psycopg2`
- [x] Fix tests to use`django.urls.reverse_lazy` imports.
- [x] Fix `urlsafe_base64_encode` usage
- [x] Add `app_name` in `urls.py` for app url resolution.


apps/challenges/views.py
```diff
@@ -627,6 +627,8 @@ def challenge_phase_detail(request, challenge_pk, pk):
         return Response(response_data, status=status.HTTP_406_NOT_ACCEPTABLE)
 
     if request.method == "GET":
+        print("In request")
```
RishabhJain2018, line 630: Please remove the print statement.

apps/challenges/views.py
```diff
@@ -627,6 +627,8 @@ def challenge_phase_detail(request, challenge_pk, pk):
...
+        print(request.data)
```
RishabhJain2018, line 631: Please remove the print statement.

apps/web/views.py
```diff
@@ -29,7 +29,7 @@ def home(request, template_name="index.html"):
     return render(request, template_name)
 
 
-def page_not_found(request):
+def page_not_found(request, exception):
```
RishabhJain2018, line 32: Why are we passing `exception` as a parameter when we are not using it?

RishabhJain2018 (changes requested):

apps/web/views.py
Ram81: @RishabhJain2018 please refer [this](https://docs.djangoproject.com/en/3.1/ref/views/). In django 2.0 `page_not_found` function has exception as mandatory param.

Ram81:
@RishabhJain2018 I've made the changes

RishabhJain2018 (changes requested):
@Ram81 Since you changed the models, we also need to have migration files right?

Ram81:
@RishabhJain2018 no, I ran `makemigrations` and it didn't detect any changes. By default value of `on_delete` was `models.CASCADE` which is why we don't need any migrations for these changes.

RishabhJain2018 (commented):
@Ram81 I can merge this one, right?

Ram81:
@RishabhJain2018  yes

RishabhJain2018 (approved):

PR #3084, opened 2020-11-23
Add Tags to Challenges hosted 

anilchoudary:
- Add a tags field (postgres array field, default=[], null=True) to Challenges Model

- List tags associated with a challenge on the challenge list page, challenge details page

- Add tags support in Challenge creation templates so that tags automatically gets created when configuration is uploaded to EvalAI

- Update/add unit tests for tags


anilchoudary:
Need more time

PR #3087, opened 2020-11-28
Add staging tags in staging worker log group names

Ram81:
### Description

Currently we create log group with `challenge-pk-{challenge_pk}-worker" format which can cause log group name conflicts between production and staging log groups.

This PR adds - 
- [x] Add `staging` tag in staging worker log group names to solve above issue.


settings/common.py
```diff
@@ -331,3 +331,5 @@
 EKS_CLUSTER_ROLE_ARN = os.environ.get("EKS_CLUSTER_ROLE_ARN")
 
 EKS_NODEGROUP_ROLE_ARN = os.environ.get("EKS_NODEGROUP_ROLE_ARN")
+
+ENVIRONMENT = os.environ.get("ENVIRONMENT", "production")
```
RishabhJain2018, line 335: Let's set it to `dev` by default and add it in the production and staging env files.

apps/challenges/aws_utils.py
```diff
@@ -338,8 +338,7 @@
 
 def get_log_group_name(challenge_pk):
     log_group_name = "challenge-pk-{}-workers".format(challenge_pk)
-    is_staging = "staging" in settings.EVALAI_API_SERVER
-    if is_staging:
+    if settings.ENVIRONMENT == "staging":
```
RishabhJain2018, line 341: Let's also add explicitly for `production`.

docker/prod/docker_staging.env
```diff
@@ -53,3 +53,4 @@ EKS_CLUSTER_ROLE_ARN=x
 EKS_NODEGROUP_ROLE_ARN=x
 
 SLACK_WEB_HOOK_URL=x
+ENVIRONMENT=staging
```
RishabhJain2018, line 56: Please add the same in production env file as well.

RishabhJain2018 (changes requested):

RishabhJain2018 (approved):
@Ram81 Could you please add this variable in the environment files on our s3 bucket before we merge this.

Ram81:
@RishabhJain2018 sure, but this change will break manage tab logs for existing challenges. We have to create the workers and log groups again to make sure they have name with new format

RishabhJain2018:
> @RishabhJain2018 sure, but this change will break manage tab logs for existing challenges. We have to create the workers and log groups again to make sure they have name with new format

How about we add a change in API to handle both the names for the log groups as in try to fetch from the older ones first and if found then return otherwise use this one.

Ram81:
That seems hacky. It would be better if we switch for all of them now so that we don't have to keep handling this case in future.

RishabhJain2018:
Alright, then we should this now. I am assuming you will have to stop the existing workers and then start again, right?

Ram81:
We have to delete workers and create again

RishabhJain2018:
cool, then we can do that before moving to production. Let's test this on staging first and once it works on staging then we can do that.

RishabhJain2018:
Can you please add the change in the API based on environment as well.

Ram81:
We don't need any changes in API right? it takes log group name from same util function. Am I missing something?

Ram81:
This line https://github.com/Cloud-CV/EvalAI/pull/3087/files#diff-3ce6744a64dac7bb4f8172e9ceeb1a3b2717399cbeec67483b1ac98574601eb8R2627

RishabhJain2018:
Got it!

Ram81:
@RishabhJain2018 I updated the s3 file you can deploy now.

RishabhJain2018:
Let's merge the other PR as well and then will deploy both.

PR #3089, opened 2020-12-03
Update dependencies

Ram81:
### Description

This PR - 

- [x] Updates all dependencies to the latest stable version after dajngo 2 migration.
- [x] Not updating moto because of this [issue](https://github.com/getsentry/responses/issues/358)


Ram81:
Not updating moto because of this [issue](https://github.com/getsentry/responses/issues/358)

RishabhJain2018 (changes requested):
@Ram81 Could you please resolve the conflicts here?

RishabhJain2018 (changes requested):
@Ram81 Could you please resolve the conflicts here?

Ram81:
Sure, will do

settings/common.py
```diff
@@ -61,6 +61,7 @@
 THIRD_PARTY_APPS = [
     "allauth",
     "allauth.account",
+    'allauth.socialaccount',
```
RishabhJain2018, line 64: Why are we adding this app?

requirements/worker.txt
```diff
@@ -1,7 +1,7 @@
...
-networkx==2.1
-pandas==0.25.3
-pycocotools==2.0.0
-scipy==1.5.2
+matplotlib==3.3.3
+networkx==2.5
+pandas==1.1.4
+pycocotools==2.0.2
+scipy==1.5.4
 sklearn==0.0
```
RishabhJain2018, line 6: Can we please change it to some relevant version than `0.0`?

RishabhJain2018 (changes requested):

settings/common.py
Ram81: After upgrading the allauth package we had to add this app explicitly to get it working. I can't find the git issue where I found the solution. I will update if I find it here

RishabhJain2018 (approved):

PR #3090, opened 2020-12-04
BugFix: Evaluation script exception handling when evaluating submissions

Ram81:
### Description

Currently when the evaluation script throws an exception while evaluating any submission the files are not being cleaned up.

This PR adds:

- [x] Exception handling code for case when evaluation fails because of exception thrown by evaluation script.
- [x] Move `submission.save()` on line 578 in if block to avoid redundant save for a failed submission.

### DIFF
--- a/scripts/workers/submission_worker.py
+++ b/scripts/workers/submission_worker.py
@@ -546,6 +546,9 @@ def run_submission(
     except Exception:
         stderr.write(traceback.format_exc())
         successful_submission_flag = False
+        # Set submission_output to None to handle case when evaluation script throws exception
+        # In case of exception from evaluation script submission_output is assigned exception object
+        submission_output = None
 
     submission_status = (
         Submission.FINISHED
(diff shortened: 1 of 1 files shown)


scripts/workers/submission_worker.py
```diff
@@ -574,8 +577,7 @@ def run_submission(
         submission.submission_metadata_file.save(
             "submission_metadata.json", ContentFile(submission_metadata)
         )
-
-    submission.save()
```
RishabhJain2018, line 578: Why are we changing this line to `if` block? How is it causing redundancy?

RishabhJain2018 (changes requested):

scripts/workers/submission_worker.py
Ram81: we need above save only if `submission_output != None` see this https://github.com/Cloud-CV/EvalAI/pull/3090/files#diff-ae6210f157b1ef66d873b08fe30f65f36c92c1e3d8aa61e3c8580910f5f06695L560. For a failed submission we are calling save twice.

PR #3091, opened 2020-12-05
Handle docker client corner cases in monitoring script

Ram81:
### Description

This PR adds -

- [x] Check if docker client is running or not and send slack notification if client is not up.
- [x] Handle case when all containers are removed i.e no container is in start/stop state
- [x] Add path argument in `run_monitor_containers.sh` (This is to handle triggering script from crontab. crontab doesn't run script from the project folder)


RishabhJain2018 (changes requested):
Can you please fix the travis build.

RishabhJain2018 (approved):

RishabhJain2018 (approved):

PR #3094, opened 2020-12-09
BugFix: Container monitoring script

Ram81:
### Description

This PR adds -

- [x] Add `MONITORING_SLACK_WEBHOOK_URL` env variable explicitly in `run_monitor_containers.sh` for crontab to be able to access it. Crontab is not able to get env variables from bash_profile and bashrc
- [x] Fix exception handling for case when docker client is down


RishabhJain2018 (changes requested):
@Ram81 Can you please fix the Travis build.

RishabhJain2018 (approved):

PR #3097, opened 2020-12-17
Update dependencies

Ram81:
### Description

This PR - 

- [x] Updates all dependencies to the latest stable version after dajngo 2 migration.
- [x] Not updating moto because of this [issue](https://github.com/getsentry/responses/issues/358)
- [x] Handle ArrayField warning after updating django to 2.2 using [this](https://github.com/geodesign/django-raster/issues/34)
- [x]  Update django version from `2.2.8` to `2.2.13`


requirements/common.txt
```diff
@@ -1,29 +1,29 @@
...
+docker-compose==1.27.4
 drfdocs==0.0.11
 drf-yasg==1.17.0
-kubernetes==10.0.1
-moto==1.3.8
-pika==0.10.0
-pickleshare==0.7.4
-Pillow==6.2.0
+kubernetes==12.0.1
+moto==1.3.14
```
RishabhJain2018, line 20: You said in your previous PR that `moto` has some issue, right?

RishabhJain2018 (changes requested):
Can you please check we are updating everything from here: https://requires.io/github/Cloud-CV/EvalAI/requirements/?branch=master

requirements/common.txt
Ram81: we are not updating moto to 1.3.15 that is the latest version which has isses

RishabhJain2018 (approved):

PR #3101, opened 2020-12-19
[Backend] Add JWT access token support

Ram81:
### Description

This PR adds - 

- [x] SimpleJWT authentication [package](https://django-rest-framework-simplejwt.readthedocs.io/en/latest/getting_started.html) support to solve the issue of auth tokens expiring after logout for cli and API calls
- [x] Add API to refresh the access token.
- [x] Add frontend changes to refresh token in frontend v1 & v2 
- [x] Add API token blacklisting on token refresh

EvalAI-cli [PR](https://github.com/Cloud-CV/evalai-cli/pull/299)

### Usage

To use the jwt token for API calls use `Bearer` instead of `Token` prefix in the authorization header of the api call followed by actual jwt token. Refer following example:

```
curl --location --request GET 'http://eval.ai/api/auth/user/' \
--header 'Authorization: Bearer <auth_token>'
```

To blacklist a token EvalAI admins or user can make a refresh token API call. Here's an example:

```
curl --location --request GET 'http://eval.ai/api/accounts/user/refresh_auth_token' \
--header 'Authorization: Bearer <auth_token>'
```
The refresh token API call will blacklist the existing token and generate a new token for the user. This won't affect login state of the user but the old `<auth_token>` will blacklisted and can't be used to get data from evalai APIs.


apps/accounts/views.py
```diff
@@ -46,23 +47,56 @@ def get_auth_token(request):
         Response(response_data, status.HTTP_404_NOT_FOUND)
 
     try:
-        token = Token.objects.get(user=user)
-    except Token.DoesNotExist:
-        token = Token.objects.create(user=user)
+        token = JwtToken.objects.get(user=user)
+    except JwtToken.DoesNotExist:
+        jwt_refresh_token = RefreshToken.for_user(user)
+        token = JwtToken.objects.create(
```
RishabhJain2018, line 53: Can we please use serializers to save this.

apps/accounts/views.py
```diff
@@ -46,23 +47,56 @@ def get_auth_token(request):
...
+        token = JwtToken.objects.get(user=user)
+        existing_token = RefreshToken(token.refresh_token)
+        existing_token.blacklist()
+    except JwtToken.DoesNotExist:
+        token = JwtToken.objects.create(user=user)
+
+    jwt_refresh_token = RefreshToken.for_user(user)
+    token.access_token = str(jwt_refresh_token.access_token)
+    token.refresh_token = str(jwt_refresh_token)
+    token.save()
```
RishabhJain2018, line 99: Can we please use serializers to save this.

apps/accounts/views.py
```diff
@@ -16,15 +15,17 @@
...
+from .models import JwtToken
 from .permissions import HasVerifiedEmail
 
 from .throttles import ResendEmailThrottle
 
 
 @api_view(["POST"])
 @permission_classes((permissions.IsAuthenticated,))
-@authentication_classes((ExpiringTokenAuthentication,))
+@authentication_classes((JWTAuthentication, ExpiringTokenAuthentication))
```
RishabhJain2018, line 28: Should we remove the `ExpiringTokenAuthentication` class from here now? We will anyway do it in future. What do you think ?

RishabhJain2018 (changes requested):

apps/accounts/views.py
Ram81: We can't. I mentioned in our call that the frontend will use `ExpiringTokenAuthentication` because we are not planning to override the login view. Otherwise we have to do a db lookup on login to get refresh tokens for login too which we decided not to do.

RishabhJain2018: Yes, I remember now. Got it.

apps/accounts/migrations/0002_add_jwt_access_token_model.py
```diff
@@ -0,0 +1,30 @@
+# Generated by Django 2.2.13 on 2021-01-01 06:23
```
RishabhJain2018, line 1: It seems like the `pre-commit` hook for black and flake8 isn't working. Can you please install the hook and use it.

RishabhJain2018 (changes requested):

settings/common.py
```diff
@@ -333,3 +337,26 @@
...
 ENVIRONMENT = os.environ.get("ENVIRONMENT", "dev")
+
+SIMPLE_JWT = {
+    "ACCESS_TOKEN_LIFETIME": timedelta(days=365),
+    "REFRESH_TOKEN_LIFETIME": timedelta(days=365),
+    "ROTATE_REFRESH_TOKENS": False,
+    "UPDATE_LAST_LOGIN": False,
+
+    "ALGORITHM": "HS256",
+    "SIGNING_KEY": "test-key",
```
RishabhJain2018, line 348: I think keeping the signing key as `test-key` is way too simple. Moreover, it should be passed as an environment variable.

settings/common.py
```diff
@@ -333,3 +337,26 @@
...
+    "VERIFYING_KEY": None,
```
RishabhJain2018, line 349: Why is this set as "None"?

RishabhJain2018 (changes requested):
Also, could you please write in the description about the blacklisting of the tokens?

settings/common.py
Ram81: As we are using HMAC as the encryption algorithm value of `VERIFYING_KEY` will be same as `SIGNING_KEY`. See this https://django-rest-framework-simplejwt.readthedocs.io/en/latest/settings.html#verifying-key

settings/common.py
Ram81: Actually we can remove this, by default it will use `SECRET_KEY` from existing django settings.

apps/accounts/migrations/0002_add_jwt_access_token_model.py
Ram81: @RishabhJain2018 no files were updated after running `pre-commit`

RishabhJain2018: Just confirming, these are the steps that you followed --
1. `pre-commit install`
2. `pre-commit run -a`

RishabhJain2018 (changes requested):

docker-compose.yml
```diff
@@ -52,15 +52,15 @@ services:
...
-      dockerfile: docker/dev/nodejs_v2/Dockerfile
-    environment:
-      ANGULAR_ENV: development
-    ports:
-      - "9999:4200"
-      - "4200:4200"
-    volumes:
-    - ./frontend_v2:/code
-    - /code/node_modules
+  # nodejs_v2:
```
RishabhJain2018, line 55: I think this is done by mistake. Can you please revert this?

RishabhJain2018 (changes requested):

RishabhJain2018 (approved):

RishabhJain2018 (changes requested):
image

@Ram81 If I click on `Refresh Token` then the visible token doesn't change. Could you please take a look at this?

Also, please add details of making an API call using this token and also the steps to blacklist a token by an admin. It would be helpful if we might need to take a look at this on some later stages.

Rest looks good!

Ram81:
@RishabhJain2018 the token value is being changed. The token has 3 parts:

1. Encoded header
2. Encoded payload which has user info
3. Encoded signing key

These 3 values are arranged in following order `<encoded_header>.<encoded_payload>.<encoded_verifying_key>`  separated by `.` what you are seeing are the first 2 parts of the token which will be constant as we have fixed header which has info about algorithm being used for hashing and second part is user info which is also constant.

If you scroll horizontally you'll see that when you refresh the token the signing key changes and is reflected on frontend. Changes are working as expected.

>Also, please add details of making an API call using this token and also the steps to blacklist a token by an admin. It would be helpful if we might need to take a look at this on some later stages.

Where do we want to add these details? PR description?

RishabhJain2018:
>@RishabhJain2018 the token value is being changed. The token has 3 parts:

Encoded header
Encoded payload which has user info
Encoded signing key
These 3 values are arranged in following order <encoded_header>.<encoded_payload>.<encoded_verifying_key> separated by . what you are seeing are the first 2 parts of the token which will be constant as we have fixed header which has info about algorithm being used for hashing and second part is user info which is also constant.

If you scroll horizontally you'll see that when you refresh the token the signing key changes and is reflected on frontend. Changes are working as expected.

I see..got it. Thanks for clarifying.

>Where do we want to add these details? PR description?

Yes.

Ram81:
@RishabhJain2018 done.

Ram81:
Created new PR #3161 . Closing this one

PR #3106, opened 2020-12-22
Pass slack webhook url as param to deployment monitoring script 

Ram81:
### Description

This PR adds `MONITORING_SLACK_WEBHOOK_URL` as a bash script param instead of hardcoding it in staging and production environment scripts.

### DIFF
--- a/scripts/monitoring/run_monitor_containers.sh
+++ b/scripts/monitoring/run_monitor_containers.sh
@@ -1,11 +1,17 @@
 #!/bin/bash
 path=$PWD
+webhook_url=''
 if [ ! -z "$1" ]
   then
     path=$1
 fi
 
+if [ ! -z "$2" ]
+  then
+    webhook_url=$2
+fi
+
 # crontab doesn't have access to env variable, define explicitly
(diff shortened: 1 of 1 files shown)


scripts/monitoring/run_monitor_containers.sh
```diff
@@ -1,11 +1,17 @@
...
+webhook_url=''
 if [ ! -z "$1" ]
   then
     path=$1
 fi
 
+if [ ! -z "$2" ]
+  then
+    webhook_url=$2
+fi
```
RishabhJain2018, line 12: Can we please throw an error saying "Please enter webhook URL" if it is not passed.

RishabhJain2018 (changes requested):

scripts/monitoring/run_monitor_containers.sh
Ram81: The python script throws an exception. Do we need additional logs?

RishabhJain2018: I see..then it is fine. I thought that it might not be handled.

RishabhJain2018 (approved):

PR #3148, opened 2021-01-16
Fix #3146: Fix cyclic imports

### LINKED ISSUE #3146: Fix cyclic import dependency in challenges/aws_utils.py
### Description

Currently, we can't import `Challenge` model or any method from `challenges/utils.py` in [challenges/aws_utils.py](https://github.com/Cloud-CV/EvalAI/blob/master/apps/challenges/aws_utils.py) because `aws_utils.py` is being imported in `challenges/model.py`. If we try to import any util methods or models from `challenges/utils.py` or `challenges/models.py` in `challenges/aws_utils.py` it fails because of cyclic imports.

### Fix

Refactor `challenges/models.py` to fix cyclic imports

burnerlee:


apps/challenges/aws_utils.py
```diff
@@ -1195,3 +1200,20 @@ def challenge_approval_callback(sender, instance, field_name, **kwargs):
                             challenge.id, failures[0]["message"]
                         )
                     )
+
+
+signals.post_save.connect(
```
Ram81, line 1205: @burnerlee we can't move models post save callback to `aws_utils`. As the name of the file suggests this file is only for code related to aws services util methods. Please move `post_save` callbacks to `challenges.models.py`

apps/challenges/aws_utils.py
```diff
@@ -1195,3 +1200,20 @@ def challenge_approval_callback(sender, instance, field_name, **kwargs):
...
+    model_field_name(field_name="test_annotation")(
+        restart_workers_signal_callback
+    ),
+    sender=ChallengePhase,
+    weak=False,
+)
+
+signals.post_save.connect(
+    model_field_name(field_name="evaluation_script")(
```
Ram81, line 1214: same here

apps/challenges/models.py
```diff
@@ -185,18 +178,10 @@ def is_active(self):
         return False
 
 
-signals.post_save.connect(
-    model_field_name(field_name="evaluation_script")(
```
Ram81, line 189: Instead of removing `post_save` callbacks from `models.py` you could create the `restart_workers` callback method in `models.py` and import `aws_utils. restart_workers_signal_callback` only in the callback method.

apps/challenges/models.py
```diff
@@ -329,15 +314,6 @@ def save(self, *args, **kwargs):
         return challenge_phase_instance
 
 
-signals.post_save.connect(
-    model_field_name(field_name="test_annotation")(
-        restart_workers_signal_callback
```
Ram81, line 334: same here

Ram81 (changes requested):

burnerlee:
@Ram81, made the requested changes. Please review

apps/challenges/models.py
```diff
@@ -185,18 +179,25 @@ def is_active(self):
...
+    signals.post_save.connect(
+        model_field_name(field_name=field_name)(
+            aws.restart_workers_signal_callback
+        ),
+        sender=sender,
+        weak=False,
+    )
+
+
+post_save_connect("evaluation_script", Challenge)
```
Ram81, line 194: @burnerlee can we move `post_save_connect("test_annotation", ChallengePhase)` from line 333 to line 195.

Ram81 (changes requested):
@burnerlee can you also share how you tested this change?

apps/challenges/models.py
burnerlee: `ChallengePhase` model is defined later in the file. Hence to move this callback up, I would also need to move that model class upwards in the file accordingly

burnerlee:
> @burnerlee can you also share how you tested this change?
I tried imported models from models.py in aws_utils.py and built and run the code. It was showing errors as expected in the beginning, but then no errors were returned after the fix was made. I was able to import anything from models.py in aws_utils then

apps/challenges/models.py
Ram81: @burnerlee you can move the `post_save_connect` callback from line 194 to line 332 then. It would be good to have all the post save callbacks at same place

Ram81:
>I tried imported models from models.py in aws_utils.py and built and run the code. It was showing errors as expected in the beginning, but then no errors were returned after the fix was made. I was able to import anything from models.py in aws_utils then

Can you also try making changes to `test_annotations` and `evaluation_script` fields from django admin and try using the save option there to see if the callbacks are working as expected or throwing errors.

burnerlee:
> > I tried imported models from models.py in aws_utils.py and built and run the code. It was showing errors as expected in the beginning, but then no errors were returned after the fix was made. I was able to import anything from models.py in aws_utils then
> 
> Can you also try making changes to `test_annotations` and `evaluation_script` fields from django admin and try using the save option there to see if the callbacks are working as expected or throwing errors.

Yeah sure! Do these fields belong to the challenge model?

burnerlee:
@Ram81 forgot to tag

burnerlee:
Moved the callbacks together. Also tested by updating the fields through the Django admin and the callbacks work fine (tested by consoling something in the callback aws_utils function)

Ram81:
LGTM 👍  @RishabhJain2018 can you please take a look.

burnerlee:
@RishabhJain2018 ping for review

RishabhJain2018 (approved):

PR #3149, opened 2021-01-17
Automate code upload cluster arn and subnet setup

Ram81:
### Description

This PR -

- [x] Adds celery task for cluster and nodegroup arn creation
- [x] Adds celery task for VPC, Internet Gateway, Route Table, Security group, and Subnet creation

### DIFF
--- a/apps/challenges/aws_utils.py
+++ b/apps/challenges/aws_utils.py
@@ -1287,6 +1287,15 @@ def create_eks_cluster_subnets(challenge):
         waiter = client.get_waiter("subnet_available")
         waiter.wait(SubnetIds=subnet_ids)
 
+        # Creating managed node group needs subnets to auto assign ip v4
+        for subnet_id in subnet_ids:
+            response = client.modify_subnet_attribute(
+                MapPublicIpOnLaunch={
+                    "Value": True,
+                },
+                SubnetId=subnet_id,
+            )
+
         # Associate route table with subnets
         response = client.associate_route_table(
             RouteTableId=route_table_id,
--- a/apps/challenges/aws_utils.py
+++ b/apps/challenges/aws_utils.py
@@ -1164,8 +1164,8 @@ def create_eks_arn_roles(challenge):
     waiter = client.get_waiter("role_exists")
     waiter.wait(RoleName=node_group_role_name)
 
-    node_group_policies = settings.EKS_NODE_GROUP_POLICIES
-    for policy_arn in node_group_policies:
+    task_execution_policies = settings.EKS_NODE_GROUP_POLICIES
+    for policy_arn in task_execution_policies:
         try:
             # Attach AWS managed EKS worker node policy to the role
             response = client.attach_role_poilcy(
@@ -1368,6 +1368,15 @@ def create_eks_cluster(challenge):
                     challenge=challenge_obj, name=cluster_name
                 )
             )
+        except Exception as e:
+            logger.exception(e)
+            return
+        challenge_aws_keys = get_aws_credentials_for_challenge(
+            challenge_obj.pk
+        )
+        client = get_boto3_client("eks", challenge_aws_keys)
+        vpc_dict = get_code_upload_setup_meta_for_challenge(challenge_obj.pk)
+        try:
             response = client.create_cluster(
                 name=cluster_name,
                 version="1.15",
--- a/apps/challenges/aws_utils.py
+++ b/apps/challenges/aws_utils.py
@@ -1315,22 +1315,6 @@ def create_eks_cluster_subnets(challenge):
         )
         security_group_id = response["GroupId"]
 
-        response = client.attach_security_group_ingress(
-            GroupId=security_group_id,
-            IpPermissions=[
-                {
-                    "FromPort": 22,
-                    "IpProtocol": "tcp",
-                    "IpRanges": [
-                        {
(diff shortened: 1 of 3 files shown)


RishabhJain2018:
@Ram81 Please resolve the conflicts here.

apps/challenges/aws_utils.py
```diff
@@ -1109,6 +1110,251 @@ def create_eks_nodegroup(challenge, cluster_name):
...
+    """
+    from .models import ChallengeEvaluationCluster
+    from .serializers import ChallengeEvaluationClusterSerializer
+    from .utils import get_aws_credentials_for_challenge
+
+    for obj in serializers.deserialize("json", challenge):
+        challenge_obj = obj.object
+    challenge_aws_keys = get_aws_credentials_for_challenge(challenge_obj.pk)
+    client = get_boto3_client("iam", challenge_aws_keys)
+    eks_role_name = "evalai-code-upload-eks-role"
```
RishabhJain2018, line 1129: Can we please attach challenge pk along with `staging`, `production` keyword to it so that it is unique for every challenge and we know exactly which challenge it is evaluating.

apps/challenges/aws_utils.py
```diff
@@ -1109,6 +1110,251 @@ def create_eks_nodegroup(challenge, cluster_name):
...
+            PolicyArn=settings.EKS_CLUSTER_POLICY,
+            AssumeRolePolicyDocument=json.dumps(
+                settings.EKS_CLUSTER_TRUST_RELATION
+            ),
+        )
+    except ClientError as e:
+        logger.exception(e)
+        return response
+
+    node_group_role_name = "evalai-code-upload-nodegroup-role"
```
RishabhJain2018, line 1156: Can we please attach challenge pk along with `staging`, `production` keyword to it so that it is unique for every challenge and we know exactly which challenge it is evaluating.

apps/challenges/aws_utils.py
```diff
@@ -1109,6 +1110,251 @@ def create_eks_nodegroup(challenge, cluster_name):
...
+    from .serializers import ChallengeEvaluationClusterSerializer
+    from .utils import get_aws_credentials_for_challenge
+
+    for obj in serializers.deserialize("json", challenge):
+        challenge_obj = obj.object
+    challenge_aws_keys = get_aws_credentials_for_challenge(challenge_obj.pk)
+    client = get_boto3_client("ec2", challenge_aws_keys)
+    vpc_ids = []
+    try:
+        response = client.create_vpc(CidrBlock="100.68.0.0/16")
```
RishabhJain2018, line 1243: Please add a TODO here to replace this hardcoded value in future.

apps/challenges/aws_utils.py
```diff
@@ -1109,6 +1110,251 @@ def create_eks_nodegroup(challenge, cluster_name):
...
+        client.create_route(
+            DestinationCidrBlock="0.0.0.0/0",
+            GatewayId=internet_gateway_id,
+            RouteTableId=route_table_id,
+        )
+
+        # Create subnets
+        subnet_ids = []
+        response = client.create_subnet(
+            CidrBlock="100.68.0.0/20",
```
RishabhJain2018, line 1272: Please add a TODO here to replace this hardcoded value in future.

apps/challenges/aws_utils.py
```diff
@@ -1109,6 +1110,251 @@ def create_eks_nodegroup(challenge, cluster_name):
...
+            AvailabilityZone="us-east-1a",
```
RishabhJain2018, line 1273: What if `us-east-1a` doesn't support it? Is it not possible to pass a list of availability zones?

apps/challenges/aws_utils.py
```diff
@@ -1109,6 +1110,251 @@ def create_eks_nodegroup(challenge, cluster_name):
...
+            VpcId=vpc_ids[0],
+        )
+        subnet_1_id = response["Subnet"]["SubnetId"]
+        subnet_ids.append(subnet_1_id)
+
+        response = client.create_subnet(
+            CidrBlock="100.68.32.0/20",
```
RishabhJain2018, line 1280: Same as above.

apps/challenges/aws_utils.py
```diff
@@ -1109,6 +1110,251 @@ def create_eks_nodegroup(challenge, cluster_name):
...
+            AvailabilityZone="us-east-1b",
```
RishabhJain2018, line 1281: Same as above.

RishabhJain2018 (changes requested):

apps/challenges/aws_utils.py
Ram81: We can only pass 1 availability zone. We could've allowed boto3 to auto assign availability zone but for EKS we need subnets in two availability zone. We can accept these from challenge config same as CIDRs

RishabhJain2018: okay, let's add a TODO above to replace this as well.

Ram81: Added it

apps/challenges/aws_utils.py
RishabhJain2018: @Ram81 Seems like you missed here.

RishabhJain2018 (changes requested):

RishabhJain2018 (approved):

PR #3150, opened 2021-01-17
Fix #2994: Handle exception when submission is not found

### LINKED ISSUE #2994: Exception when model not found in `get_model_object` utils
### Description

Currently, the `get_model_object` util raises an exception when a model object with queried primary key is not found. This exception is not being handled in partial submission evaluation API. See [this](https://sentry.io/organizations/cloudcv/issues/1883775184/?project=152529&referrer=slack) error for an example

burnerlee:
Handle exception when `get_model_object` throws an error in `update_partially_evaluated_submission` API when submission is not found for given submission_id

### DIFF
--- a/apps/jobs/views.py
+++ b/apps/jobs/views.py
@@ -19,6 +19,7 @@ from django.core.files.uploadedfile import SimpleUploadedFile
 from django.db import transaction, IntegrityError
 from django.db.models import Count
 from django.utils import timezone
+from rest_framework.exceptions import NotFound
 
 from rest_framework_expiring_authtoken.authentication import (
     ExpiringTokenAuthentication,
@@ -646,11 +647,11 @@ def leaderboard(request, challenge_phase_split_id):
                                 ),
                                 "submission__is_baseline": openapi.Schema(
                                     type=openapi.TYPE_BOOLEAN,
-                                    description="Boolean to decide if submission is baseline"
+                                    description="Boolean to decide if submission is baseline",
                                 ),
                                 "submission__is_public": openapi.Schema(
                                     type=openapi.TYPE_BOOLEAN,
-                                    description="Boolean to decide if submission is public"
+                                    description="Boolean to decide if submission is public",
                                 ),
                                 "challenge_phase_split": openapi.Schema(
                                     type=openapi.TYPE_STRING,
@@ -665,7 +666,7 @@ def leaderboard(request, challenge_phase_split_id):
                                 ),
                                 "error": openapi.Schema(
                                     type=openapi.TYPE_STRING,
-                                    description="Error returned for the result"
+                                    description="Error returned for the result",
                                 ),
                                 "leaderboard__schema": openapi.Schema(
                                     type=openapi.TYPE_OBJECT,
@@ -682,7 +683,7 @@ def leaderboard(request, challenge_phase_split_id):
                                             type=openapi.TYPE_STRING,
                                             description="Default ordering label for the leaderboard schema",
                                         ),
-                                    }
+                                    },
                                 ),
                                 "submission__submitted_at": openapi.Schema(
                                     type=openapi.TYPE_STRING,
@@ -1151,8 +1152,8 @@ def update_submission(request, challenge_pk):
                         response_data, status=status.HTTP_400_BAD_REQUEST
                     )
 
-                leaderboard_metrics = challenge_phase_split.leaderboard.schema.get(
-                    "labels"
+                leaderboard_metrics = (
+                    challenge_phase_split.leaderboard.schema.get("labels")
                 )
                 missing_metrics = []
                 malformed_metrics = []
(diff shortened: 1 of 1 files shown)


apps/jobs/views.py
```diff
@@ -1440,8 +1441,15 @@ def update_partially_evaluated_submission(request, challenge_pk):
         stderr_content = request.data.get("stderr", "")
         submission_result = request.data.get("result", "")
         metadata = request.data.get("metadata", "")
-        submission = get_submission_model(submission_pk)
-
+        try:
+            submission = get_submission_model(submission_pk)
+        except NotFound:
+            response_data = {
+                "error": "Submission does not exist with submission_id {}".format(
```
Ram81, line 1448: Change `Submission does not exist with submission_id {}` to `Submission with submission_pk {} does not exist`

apps/jobs/views.py
```diff
@@ -1589,7 +1597,15 @@ def update_partially_evaluated_submission(request, challenge_pk):
         submission_pk = request.data.get("submission")
         submission_status = request.data.get("submission_status", "").lower()
         job_name = request.data.get("job_name", "").lower()
-        submission = get_submission_model(submission_pk)
+        try:
+            submission = get_submission_model(submission_pk)
+        except NotFound:
+            response_data = {
+                "error": "Submission does not exist with submission_id {}".format(
```
Ram81, line 1604: Change `Submission does not exist with submission_id {}` to `Submission with submission_pk {} does not exist`

Ram81 (changes requested):

burnerlee:
@Ram81 made the requested change

apps/jobs/views.py
```diff
@@ -1440,8 +1441,15 @@ def update_partially_evaluated_submission(request, challenge_pk):
         stderr_content = request.data.get("stderr", "")
         submission_result = request.data.get("result", "")
         metadata = request.data.get("metadata", "")
-        submission = get_submission_model(submission_pk)
-
+        try:
+            submission = get_submission_model(submission_pk)
+        except NotFound:
```
Ram81, line 1446: @burnerlee instead of `NotFound` use `Submission.DoesNotExist`. You can see other APIs in `jobs/views.py` to see an example.

apps/jobs/views.py
```diff
@@ -1589,7 +1597,15 @@ def update_partially_evaluated_submission(request, challenge_pk):
         submission_pk = request.data.get("submission")
         submission_status = request.data.get("submission_status", "").lower()
         job_name = request.data.get("job_name", "").lower()
-        submission = get_submission_model(submission_pk)
+        try:
+            submission = get_submission_model(submission_pk)
+        except NotFound:
```
Ram81, line 1602: same here

Ram81 (changes requested):

apps/jobs/views.py
burnerlee: @Ram81, I had tried this earlier. But the `get_model_object` raises an exception of the type `NotFound` and was not caught by when `Except Submission.DoesNotExist` was used

Ram81 (approved):

apps/jobs/views.py
RishabhJain2018: @burnerlee @Ram81 Why are we using `get_submission_model` in a `try/catch` block? IIRC, `get_submission_model` is derived from `get_model_object` and it automatically catches the exception here -- https://github.com/Cloud-CV/EvalAI/blob/5698781b99e3fcea100430a2ef85322c0f0545db/apps/base/utils.py#L70.
Am I missing something?

apps/jobs/views.py
RishabhJain2018: @burnerlee @Ram81 Why are we using `get_submission_model` in a `try/catch` block? IIRC, `get_submission_model` is derived from `get_model_object` and it automatically catches the exception here -- https://github.com/Cloud-CV/EvalAI/blob/5698781b99e3fcea100430a2ef85322c0f0545db/apps/base/utils.py#L70.
Am I missing something?

RishabhJain2018 (changes requested):

burnerlee:
@RishabhJain2018 yes, it catches the exception, but therefore it sends an error response of 404 status to the frontend. This PR is to catch it and send a 400 Bad Request Error instead

RishabhJain2018:
>@RishabhJain2018 yes, it catches the exception, but therefore it sends an error response of 404 status to the frontend. This PR is to catch it and send a 400 Bad Request Error instead

I think showing `400` isn't correct when the submission ID doesn't exist. It should be 404.
@Ram81 Could you please check this?

Ram81:
@RishabhJain2018 In almost all of the EvalAI APIs we throw a `400` bad request when we don't find an entry in db. We can start using `404` from now.

burnerlee:
@Ram81 what action is required here? It already throws a 404 error without this PR

RishabhJain2018:
Yes, `404` would be good.

Thanks, @burnerlee for the effort but we will need to close this PR.

PR #3173, opened 2021-01-25
Adding Challenge Type Chip with CSS

tanishqsandhu:
**Adding Challenge Type Chip with CSS**

For the Challenge Type work I have added the frontend chip with proper CSS- color can be changed as needed. Let me know if any changes need to be made.

Will transition to backend work now. See screenshot below:

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,3 +96,109 @@ EvalAI is currently maintained by [Rishabh Jain](https://rishabhjain.xyz/) and [
 ## Contribution guidelines
 
 If you are interested in contributing to EvalAI, follow our [contribution guidelines](https://github.com/Cloud-CV/EvalAI/blob/master/.github/CONTRIBUTING.md).
+
+
+## Contributors
+
+[//]: contributor-faces
+<a href="https://github.com/RishabhJain2018"><img src="https://avatars.githubusercontent.com/u/12206047?v=4" title="RishabhJain2018" width="50" height="50"></a>
+<a href="https://github.com/deshraj"><img src="https://avatars.githubusercontent.com/u/2945708?v=4" title="deshraj" width="50" height="50"></a>
+<a href="https://github.com/Ram81"><img src="https://avatars.githubusercontent.com/u/16323427?v=4" title="Ram81" width="50" height="50"></a>
+<a href="https://github.com/taranjeet"><img src="https://avatars.githubusercontent.com/u/4302268?v=4" title="taranjeet" width="50" height="50"></a>
+<a href="https://github.com/Sanji515"><img src="https://avatars.githubusercontent.com/u/32524438?v=4" title="Sanji515" width="50" height="50"></a>
+<a href="https://github.com/aka-jain"><img src="https://avatars.githubusercontent.com/u/11537940?v=4" title="aka-jain" width="50" height="50"></a>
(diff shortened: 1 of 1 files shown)


frontend/src/views/web/challenge-list.html
```diff
@@ -9,7 +9,8 @@
                         <img class="bg-img" ng-src="{{challenge.image}}">
                         <span class=" ev-card-title fs-14"><span><img ng-src="{{challenge.image}}" ></span>{{challenge.title}}</span>
                     </div>
-                    <div class="card-content">
+                    <div class="chip">test chip</div>
```
Ram81, line 12: @tanishqsandhu these chips need to be fetched from backend. Now that you have frontend ready you can add a new field in challenge model and pass it to frontend from get challenge api. Next step should be showing the title of the chip from data returned by the API

Ram81 (changes requested):

tanishqsandhu:
image

The CHallenge Type field is now added and being passed to front end.

apps/challenges/models.py
```diff
@@ -42,6 +42,9 @@ def __init__(self, *args, **kwargs):
     terms_and_conditions = models.TextField(null=True, blank=True)
     submission_guidelines = models.TextField(null=True, blank=True)
     evaluation_details = models.TextField(null=True, blank=True)
+    challenge_type = ArrayField(
```
RishabhJain2018, line 45: Can we please change the name to `tags` and also add `null=True`

RishabhJain2018: Also, please change the max_length to 200.

RishabhJain2018 (changes requested):
Can you please make changes to the existing API's which fetches the data from this field and you can display it on UI.

tanishqsandhu (commented):
Hey, had a quick question- since migration 80&81 cancel each other out, do I leave it or do I delete?

Also here is my migrations :
image

RishabhJain2018:
> Hey, had a quick question- since migration 80&81 cancel each other out, do I leave it or do I delete?
> 
> Also here is my migrations :
> image

HI @tanishqsandhu , You will need to remove both the migrations and create migrations again for `tags` field such that it is in sync with the previous migrations.

tanishqsandhu:
> > Hey, had a quick question- since migration 80&81 cancel each other out, do I leave it or do I delete?
> > Also here is my migrations :
> > image
> 
> Hi @tanishqsandhu , You will need to remove both the migrations and create migrations again for `tags` field such that it is in sync with the previous migrations.

Hey @RishabhJain2018 , removed migrations and redid tags migration. Does it look fine now?

Ram81 (changes requested):
@tanishqsandhu can you fix the failing travis build. The unit tests we have for challenge serializer are failing as a new field was added to the serializer. You have to go through unit tests once to fix the failing tests. To run them locally you can use following command
```
docker-compose run -e DJANGO_SETTINGS_MODULE=settings.test django pytest tests/unit/challenges/test_views.py
```

frontend/src/views/web/challenge-list.html
RishabhJain2018: Hey @tanishqsandhu, Can you incorporate this comment. Here are the steps that you will have to follow --
1. In the challenge controller, check if you are getting this `tags` field from the backend.
2. if yes, this `test chip` should be replaced by that.
3. if the field's value is none or blank, then it shouldn't appear.

RishabhJain2018 (changes requested):

PR #3179, opened 2021-01-26
Allow only team creator to update team details

Ram81:
### Description

This PR - 

- [x] Adds check in participant team details API to allow details update only by team creator.

### DIFF
--- a/apps/participants/utils.py
+++ b/apps/participants/utils.py
@@ -28,7 +28,7 @@ def get_participant_teams_for_user(user):
 
 
 def is_user_creator_of_participant_team(user, participant_team):
-    """Returns boolean if user is creator of Participant Team ids"""
+    """Returns boolean if a user is the creator of participant team"""
     return participant_team.created_by.pk == user.pk
 
 
--- a/apps/participants/utils.py
+++ b/apps/participants/utils.py
@@ -27,6 +27,11 @@ def get_participant_teams_for_user(user):
     return Participant.objects.filter(user=user).values_list("team", flat=True)
 
 
+def is_user_creator_of_participant_team(user, participant_team):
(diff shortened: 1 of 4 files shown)


apps/participants/views.py
```diff
@@ -137,6 +138,18 @@ def participant_team_detail(request, pk):
     elif request.method in ["PUT", "PATCH"]:
 
         if request.method == "PATCH":
+            if not is_user_creator_of_participant_team(
+                request.user, participant_team
+            ):
+                response_data = {
+                    "error": [
```
deshraj, line 145: Value of `error` key should be string instead of a list, no?

apps/participants/utils.py
```diff
@@ -27,6 +27,11 @@ def get_participant_teams_for_user(user):
     return Participant.objects.filter(user=user).values_list("team", flat=True)
 
 
+def is_user_creator_of_participant_team(user, participant_team):
+    """Returns boolean if user is creator of Participant Team ids"""
```
RishabhJain2018, line 31: `Returns boolean if user is creator of Participant Team ids` -> `Returns boolean if a user is the creator of participant team`

RishabhJain2018 (changes requested):

apps/participants/views.py
Ram81: On the frontend side we expect array of errors and show only error at 0th index. I think it is because we only used to return serializer errors from this API

RishabhJain2018: okay, but shouldn't we make a change on the frontend checking if `error` is present then we are displaying the string?
Sending the value as a list doesn't look right to me.

RishabhJain2018 (approved):

PR #3186, opened 2021-01-28
Fix #3146: Fix cyclic imports issue in aws_utils

### LINKED ISSUE #3146: Fix cyclic import dependency in challenges/aws_utils.py

burnerlee:
Allow the developer to import all models at the top of the file challenges/aws_utils.py. Since the callback is being called once on initialisation, there were not many workarounds and the post save function's definition had to be moved to a new file to resolve the conflict.


burnerlee:
@Ram81 ping for review

apps/challenges/aws_utils.py
```diff
@@ -20,8 +20,9 @@
 from base.utils import get_boto3_client, send_email
 from evalai.celery import app
 
-logger = logging.getLogger(__name__)
+# from .models import ChallengeEvaluationCluster,Challenge
```
Ram81, line 23: @burnerlee can you uncomment this and move all imports from methods `setup_eks_cluster`, `create_subnets_and_arns` `create_cluster`, etc here.

apps/challenges/aws_utils.py
```diff
@@ -1147,8 +1148,7 @@ def setup_eks_cluster(challenge):
     try:
         # Attach AWS managed EKS cluster policy to the role
         response = client.attach_role_policy(
-            RoleName=eks_role_name,
-            PolicyArn=settings.EKS_CLUSTER_POLICY,
+            RoleName=eks_role_name, PolicyArn=settings.EKS_CLUSTER_POLICY,
```
Ram81, line 1151: @burnerlee were this changes done by `pre-commit`? If not please revert

Ram81 (changes requested):

burnerlee:
@Ram81 I just realized I missed something and it would continue to throw errors in some cases now too. The post_save_connect function is always called when the modules are initialized and thus once called, it would need the aws_utils.py, after which the circular import error is bound to happen. I think the best possible solution to this is if we could move the post_save_connect function calls to aws_utils.py. The function is still defined in the models.py file

burnerlee:
Tested this code, doesn't throw errors and the callbacks are working fine too

apps/challenges/aws_utils.py
```diff
@@ -1118,7 +1122,6 @@ def setup_eks_cluster(challenge):
     Arguments:
         instance {<class 'django.db.models.query.QuerySet'>} -- instance of the model calling the post hook
     """
-    from .models import ChallengeEvaluationCluster
     from .serializers import ChallengeEvaluationClusterSerializer
```
Ram81, line 1125: @burnerlee please move serializer and utils imports to the start of the file and remove imports from the methods.

Ram81 (changes requested):

burnerlee:
@Ram81 move the imports at the starting of the file

RishabhJain2018:
@Ram81 Can you please review this?

apps/challenges/models.py
```diff
@@ -315,21 +313,16 @@ def save(self, *args, **kwargs):
         return challenge_phase_instance
 
 
-def post_save_connect(field_name, sender):
-    import challenges.aws_utils as aws
+def post_save_connect(field_name, sender, callback):
     signals.post_save.connect(
         model_field_name(field_name=field_name)(
-            aws.restart_workers_signal_callback
```
Ram81, line 322: @burnerlee why are we removing restart workers callback from here and not removing the eks cluster setup callback from [here](https://github.com/Cloud-CV/EvalAI/blob/master/apps/challenges/models.py#L194)

Ram81 (changes requested):

apps/challenges/models.py
burnerlee: @Ram81 this is because the functions like `create_eks_cluster_for_challenge`are called only when post_save signal is received rather than being called during the start of evalai program. Whereas the callback function in the first case is called when the program startup which causes an error of cyclic import as all modules are initialized in the starting. So these functions import these modules inside them which actually imports the module when these functions are called. 
Whereas if I do the same for the callback function, it would import the module at the start of the execution of the program only as it is called at the start of the program which causes the same problem which is created due to importing the modules at the top of the file.

burnerlee:
@Ram81 ping

apps/challenges/aws_utils.py
```diff
@@ -1025,6 +1023,12 @@ def restart_workers_signal_callback(sender, instance, field_name, **kwargs):
                     )
 
 
+post_save_connect("evaluation_script", Challenge,
```
Ram81, line 1026: @burnerlee moving post save callback register logic to `aws_utils` doesn't make sense. We need a cleaner way to fix the cyclic imports. It's fine if we have to refactor some part of the code but let's keep callbacks in `models.py`.

Ram81 (changes requested):

PR #3196, opened 2021-02-03
Backend: Fix bug for not parsing "#" in all_submissions_download

RishabhJain2018:
This PR --
- [x] "#" isn't parsed by writer.writerow(), hence it is replaced by "-" in the `method description` and `method name` fields.

### DIFF
--- a/apps/challenges/views.py
+++ b/apps/challenges/views.py
@@ -1793,7 +1793,6 @@ def download_all_submissions(
                 )
                 # Issue: "#" isn't parsed by writer.writerow(), hence it is replaced by "-"
                 # TODO: Find a better way to solve the above issue.
-                conversion = set("#")
                 for submission in submissions.data:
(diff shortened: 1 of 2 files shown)


apps/challenges/views.py
```diff
@@ -1827,8 +1828,8 @@ def download_all_submissions(
                             submission["created_at"],
                             submission["submission_result_file"],
                             submission["submission_metadata_file"],
-                            submission["method_name"],
-                            submission["method_description"],
+                            "".join("-" if c in conversion else c for c in submission["method_name"]),
```
Ram81, line 1831: @RishabhJain2018 why can't we use `.replace` method?

RishabhJain2018: yes, make sense. I was testing with a bunch of characters and finally, it reduced to one.

PR #3224, opened 2021-02-05
Allow only team participant to make participate API call

Ram81:
### Description

This PR -

- [x] Adds check to allow only team participants to make the participate API call.


RishabhJain2018 (changes requested):
@Ram81 Can you please fix the Travis build.

RishabhJain2018 (approved):

PR #3233, opened 2021-02-09
Add check to override submissions visibility for private phase

Ram81:
### Description

This PR -

- [x] Adds check to override submission visibility passed from cli when phase is private


apps/jobs/views.py
```diff
@@ -336,6 +336,10 @@ def challenge_submission(request, challenge_id, challenge_phase_id):
                     if submission_serializer.is_valid():
                         submission_serializer.save()
 
+        # Override submission visibility if leaderboard_public = False for a challenge phase
+        if not challenge_phase.leaderboard_public:
+            request.data["is_public"] = False
```
RishabhJain2018, line 341: Let's not hardcode this but pick from the `challenge_phase` object.

RishabhJain2018 (changes requested):

RishabhJain2018:
@Ram81 Can you please update this PR?

Ram81:
@RishabhJain2018 done

apps/jobs/views.py
```diff
@@ -336,6 +336,10 @@ def challenge_submission(request, challenge_id, challenge_phase_id):
                     if submission_serializer.is_valid():
                         submission_serializer.save()
 
+        # Override submission visibility if leaderboard_public = False for a challenge phase
+        if not challenge_phase.leaderboard_public:
+            request.data["is_public"] = challenge_phase.leaderboard_public
```
RishabhJain2018, line 341: It should be `challenge_phase.leaderboard_public` -> `challenge_phase.is_submission_public`, no?

RishabhJain2018 (changes requested):

apps/jobs/views.py
Ram81: is it possible for hosts to set `leaderboard_public=False` and `is_submission_public=True` ? both will be same right?

RishabhJain2018: Yes, it is possible for them to set these things and they both are different fields.

RishabhJain2018 (approved):

PR #3250, opened 2021-02-16
Fix #3249: Added JWT as Auth Token

### LINKED ISSUE #3249: Backend: Submit page on EvalAI for code upload challenges shows old auth token
**Issue**: Since we recently changed to JWT auth token, the submit page for code upload challenges on EvalAI still shows the old token.
image

**Fix**:
- [x] Replace the current auth token with the  new token.

ShauryaAg:
Fixes #3249 

After fix:
image

### DIFF
--- a/frontend/src/css/modules/submission.scss
+++ b/frontend/src/css/modules/submission.scss
@@ -45,16 +45,22 @@ li {
 
 code {
 	font-family: "Roboto";
+	display: inline-block;
 	font-size: 14px;
 	font-weight: 300;
 	background:#eff0f1; border:#eff0f1 dashed 1px; padding:6px;
 	border-radius: 10px;
+	max-width: 90%;
+	white-space: nowrap;
+	overflow: hidden;
+	text-overflow: ellipsis;
 }
 
 .code-btn {
 	height: 100% !important;
 	padding: 0 0.5rem !important;
 	line-height: 200% !important;
+	margin-top: -25px;
 }
 
 .dashed-link {
--- a/frontend/src/js/controllers/authCtrl.js
+++ b/frontend/src/js/controllers/authCtrl.js
@@ -186,6 +186,38 @@
             }
         };
 
+        // Function to fetch and set refreshJWT 
+        vm.setRefreshJWT = function () {
+            var parameters = {};
+            parameters.url = 'accounts/user/get_auth_token';
+            parameters.method = 'GET';
+            parameters.token = utilities.getData('userKey');
+            parameters.callback = {
+                onSuccess: function (response) {
+                    if (response.status == 200) {
+                        utilities.storeData('refreshJWT', response.data.token);
+                    } else {
+                        alert("Could not fetch Auth Token");
+                    }
+                },
+                onError: function (response) {
+                    if (response.status == 400) {
+                        vm.isFormError = true;
+                        var non_field_errors;
+                        try {
+                            non_field_errors = typeof (response.data.non_field_errors) !== 'undefined' ? true : false;
+                            if (non_field_errors) {
+                                vm.FormError = response.data.non_field_errors[0];
+                            }
+                        } catch (error) {
+                            $rootScope.notify("error", error);
+                        }
+                    }
+                }
+            };
+            utilities.sendRequest(parameters, "header");
+        };
+
         // Function to login
         vm.userLogin = function(loginFormValid) {
             if (loginFormValid) {
@@ -202,6 +234,7 @@
                     onSuccess: function(response) {
                         if (response.status == 200) {
                             utilities.storeData('userKey', response.data.token);
+                            vm.setRefreshJWT();
                             if ($rootScope.previousState) {
                                 $state.go($rootScope.previousState);
                                 vm.stopLoader();
--- a/frontend/src/js/controllers/challengeCtrl.js
+++ b/frontend/src/js/controllers/challengeCtrl.js
@@ -82,6 +82,8 @@
         var userKey = utilities.getData('userKey');
         vm.authToken = userKey;
 
+        vm.refreshJWT = utilities.getData('refreshJWT');
+
         vm.subErrors = {};
 
         vm.isChallengeLeaderboardPrivate = false;
--- a/frontend/src/views/web/challenge/submission.html
+++ b/frontend/src/views/web/challenge/submission.html
@@ -46,10 +46,10 @@
                         </button>
                     </li>
                     <li>Add your EvalAI account token to evalai-cli<br>
-                        <code>$ evalai set_token {{challenge.authToken}}</code>
+                        <code>$ evalai set_token {{challenge.refreshJWT}}</code>
                         <button class="btn ev-btn-dark waves-effect waves-dark grad-btn code-btn"
                             ng-click="challenge.showConfirmation('Command copied to clipboard')" ngclipboard
-                            data-clipboard-text="evalai set_token {{challenge.authToken}}">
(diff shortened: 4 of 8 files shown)


ShauryaAg:
@RishabhJain2018 @Ram81 I have made the necessary changes to set JWT as the Auth Token, but I still have a doubt about a piece of code.

https://github.com/Cloud-CV/EvalAI/blob/adde7be63e25af834d65ef2625103040122f561e/settings/common.py#L355

Here, you are using `RefreshToken` to prove authentication, but I think it will much more sense to use `AccessToken` instead.
Should I change this too?

RishabhJain2018 (commented):
@ShauryaAg Is this PR cpmplete?

ShauryaAg:
@RishabhJain2018 The PR is complete, and it does the necessary function of using `JWT` instead of `ExpiringAuthToken`. But please take a look at changing `RefreshToken` to `AccessToken` as I have mentioned above.

settings/common.py
```diff
@@ -348,7 +350,7 @@
     "ALGORITHM": "HS256",
     "AUDIENCE": None,
     "ISSUER": None,
-    "AUTH_HEADER_TYPES": ("Bearer",),
```
RishabhJain2018, line 351: Why are we changing this?

settings/common.py
```diff
@@ -156,6 +156,8 @@
 
 SITE_ID = 1
 
+REST_USE_JWT = True
```
RishabhJain2018, line 159: Why do we need to add this?

evalai/urls.py
```diff
@@ -21,7 +21,7 @@
 from django.conf.urls.static import static
 
 from allauth.account.views import ConfirmEmailView
-from rest_framework_expiring_authtoken.views import obtain_expiring_auth_token
+from rest_framework_simplejwt.views import token_obtain_pair
```
RishabhJain2018, line 24: Why are we changing this?

evalai/urls.py
```diff
@@ -49,8 +49,8 @@
     url(r"^api/admin/", admin.site.urls),
     url(
         r"^api/auth/login",
-        obtain_expiring_auth_token,
-        name="obtain_expiring_auth_token",
+        token_obtain_pair,
```
RishabhJain2018, line 52: Why are we changing this?

RishabhJain2018 (changes requested):

RishabhJain2018:
> >@RishabhJain2018 @Ram81 I have made the necessary changes to set JWT as the Auth Token, but I still have a doubt about a piece of code.
> 
> EvalAI/settings/common.py
> 
> Line 355 in adde7be
> 
>  "AUTH_TOKEN_CLASSES": ("rest_framework_simplejwt.tokens.RefreshToken",), 
> Here, we are using RefreshToken to prove authentication, but I think it will make much more sense to use AccessToken instead.
> Should I change this too?

We are using `RefreshToken` because we want a way to blacklist some of the tokens. IIRC, it isn't possible with `Access Token`.
cc: @Ram81

settings/common.py
ShauryaAg: This is changed so that `rest_auth` can verify the user after login. If we don't make this change, we'll need to make changes in the frontend to set default `Authorization` headers as `Bearer <Token>` (currently, it sets default `Authorization` header as `Token <Token>`, [here](https://github.com/Cloud-CV/EvalAI/blob/adde7be63e25af834d65ef2625103040122f561e/frontend_v2/src/app/services/api.service.ts#L26))

settings/common.py
ShauryaAg: This changed is need for `rest_auth` to verify user through `JWT`.

evalai/urls.py
ShauryaAg: Previously, we were using `obtain_expiring_auth_token` which returns a user `Token` (**not** `JWT`), `token_obtain_pair` method returns the `Access Token JWT` and `Refresh Token JWT`.

evalai/urls.py
ShauryaAg: This is to replace `obtain_expiring_auth_token` with `token_obtain_pair` for login.

Ram81: @ShauryaAg we can't use this. If you try login/logout multiple times the JWT token will keep changing. We are using JWT tokens mainly for API and CLI api calls. The expiring token is used for frontend codebase. For this change you just have to add an API call for `get_auth_token` [API](https://github.com/Cloud-CV/EvalAI/blob/master/apps/accounts/views.py#L43) and use the token returned by this API to show on `Submit` page.

Ram81:
Yes, that's correct

settings/common.py
```diff
@@ -348,7 +350,7 @@
     "ALGORITHM": "HS256",
     "AUDIENCE": None,
     "ISSUER": None,
-    "AUTH_HEADER_TYPES": ("Bearer",),
+    "AUTH_HEADER_TYPES": ("Token",),
```
Ram81, line 353: @ShauryaAg we are not switching to `AccessToken`. We are using `RefreshToken` because we want to support token blacklisting. Can you revert this change.

Ram81 (changes requested):

settings/common.py
ShauryaAg: @Ram81 I did not change `RefreshToken` to `AccessToken`, it was only something I suggest as I was curious.

However, changing `Bearer` to `Token` is still needed, as `rest_auth` verifies and returns using request headers with `Authorization: Token <Token>` as set in the frontend.
https://github.com/Cloud-CV/EvalAI/blob/adde7be63e25af834d65ef2625103040122f561e/frontend_v2/src/app/services/api.service.ts#L26

Here, we are setting default headers for `Authorization` as `Token <Token>`. If we don't change `Bearer` to `Token` in `AUTH_HEADER_TYPES`, then we will need to set `Authorization` header as `Bearer <Token>` in the frontend.

evalai/urls.py
ShauryaAg: Currently, the `authtoken` is set in the local storage of the browser during login. Should I just make changes in the frontend to call that `get_auth_token` method everytime instead of setting the `authtoken` in local storage as before?

settings/common.py
Ram81: We don't want to switch to Bearer token on frontend side. To show the `RefreshToken` on the submit page just make an API call to `get_auth_token` API and use the response to show the token on UI. No need to change the authorization headers on frontend local storage

evalai/urls.py
Ram81: Yes, or you could make API call once and store the bearer token as a separate variable in local storage.

ShauryaAg:
I have made the changes in Frontend as well as Frontend_V2 to set `refreshJWT` in local storage during login. Please review.

ShauryaAg:
- Frontend after change: 
image

- Frontend_V2 after change:
image

RishabhJain2018 (changes requested):
Hey @ShauryaAg, On the UI, can we please show the token up to say 20 characters and then show `...` after it. It currently looks a bit ugly on UI.

Also, please make sure that when it is copied then the complete token is copied.

ShauryaAg:
Frontend After Fix:
image

Frontend_V2 After Fix:
image

RishabhJain2018 (changes requested):
LGTM, code-wise.
@Ram81 Can you please check on your local once if this works?

Ram81:
@RishabhJain2018 it works fine. Verified locally.

RishabhJain2018 (approved):

PR #3260, opened 2021-02-22
Fix error handling response in eks cluster creation celery task

Ram81:
### Description

We were returning `response` from eks cluster creation celery task on exception even when it is not being used and variable is unaccessible in exception block. This PR fixes that.


apps/challenges/aws_utils.py
```diff
@@ -1101,9 +1101,10 @@ def create_eks_nodegroup(challenge, cluster_name):
             amiType="AL2_x86_64_GPU",
             nodeRole=cluster_meta["EKS_NODEGROUP_ROLE_ARN"],
         )
+        logger.info("Nodegroup create: {}".format(response))
     except ClientError as e:
         logger.exception(e)
-        return response
+        return
```
RishabhJain2018, line 1107: image
Can we please use these exceptions to handle it?

RishabhJain2018 (changes requested):

apps/challenges/aws_utils.py
Ram81: ClientError handles it right? we are logging exception to sentry. The issue was we were accessing `response` variable in exception block where it is not initialized

RishabhJain2018 (approved):

PR #3274, opened 2021-03-01
Add support to attach EFS to code upload worker job pods

Ram81:
### Description

This PR adds - 

- [x] Create EFS and security groups to restrict access to EFS for a specific EKS cluster.
- [x] Mount EFS as a persistent volume on code upload worker pod that will only be accessible to the environment image container in the job.


apps/challenges/models.py
```diff
@@ -568,6 +568,16 @@ class ChallengeEvaluationCluster(TimeStampedModel):
         max_length=512, null=True, blank=True
     )
     route_table_id = models.CharField(max_length=512, null=True, blank=True)
+    efs_security_group_id = models.CharField(
```
RishabhJain2018, line 571: I think these will go Challenge Phase model as one environment corresponds to one `phase`. What do you think?

scripts/workers/code_upload_worker_utils/persistent_volume.yaml
```diff
@@ -0,0 +1,15 @@
...
+  capacity:
+    storage: 5Gi
+  volumeMode: Filesystem
+  accessModes:
+    - ReadWriteMany
+  persistentVolumeReclaimPolicy: Retain
+  storageClassName: efs-sc
+  csi:
+    driver: efs.csi.aws.com
+    volumeHandle: {{EFS_ID}}
```
RishabhJain2018, line 15: Please add a new line here.

scripts/workers/code_upload_worker_utils/persistent_volume_claim.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: PersistentVolumeClaim
+metadata:
+  name: efs-claim
+spec:
+  accessModes:
+    - ReadWriteMany
+  storageClassName: efs-sc
+  resources:
+    requests:
+      storage: 5Gi
```
RishabhJain2018, line 11: Please add a new line here.

RishabhJain2018, line 11: Also, please change this storage to 500 GB

scripts/workers/code_upload_worker_utils/install_dependencies.sh
```diff
@@ -31,6 +31,12 @@ echo "### Kubectl Installed"
 curl https://raw.githubusercontent.com/aws-samples/amazon-cloudwatch-container-insights/latest/k8s-deployment-manifest-templates/deployment-mode/daemonset/container-insights-monitoring/quickstart/cwagent-fluentd-quickstart.yaml | sed "s/{{cluster_name}}/$CLUSTER_NAME/;s/{{region_name}}/$AWS_DEFAULT_REGION/" | kubectl apply -f -
 echo "### Container Insights Installed"
 
+# Setup EFS as persistent volume
```
RishabhJain2018, line 34: At the moment, this will get set up for every cluster, right?

scripts/workers/code_upload_worker_utils/persistent_volume_storage_class.yaml
```diff
@@ -0,0 +1,5 @@
+kind: StorageClass
+apiVersion: storage.k8s.io/v1
+metadata:
+  name: efs-sc
+provisioner: efs.csi.aws.com
```
RishabhJain2018, line 5: Please add a new line here.

scripts/workers/code_upload_worker_utils/install_dependencies.sh
RishabhJain2018: We should add a check in the database for a challenge phase if it requires the persistent volume or not.

RishabhJain2018 (changes requested):

scripts/workers/code_upload_worker_utils/persistent_volume_claim.yaml
Ram81: The size mentioned here won't affect how much data you could store on EFS. As EFS doesn't have any upper limit we don't have to worry about it. We have to mention some size here because kubernetes needs it. We can keep this value as it is

scripts/workers/code_upload_worker_utils/install_dependencies.sh
Ram81: Makes sense. I will add that check

apps/challenges/models.py
Ram81: We can instead maintain a dict of EFS meta data in `ChallengeEvaluationCluster` model. Right now meta of all components is in the `ChallengeEvaluationCluster` model

scripts/workers/code_upload_worker_utils/install_dependencies.sh
Ram81: But this will complicate things in bash script. We have to somehow pass a list of tuples Efs id, pvc name, pv name and use some logic in code upload worker to get mapping of which pvc and pv to mount on which pod

apps/challenges/models.py
RishabhJain2018: yes, but `efs_id` will go into the challenge_phase, right?

Ram81: Yeah, we can store it in `ChallengePhase` model

apps/challenges/aws_utils.py
```diff
@@ -430,8 +441,15 @@ def register_task_def_by_challenge_pk(client, queue_name, challenge):
...
         if challenge.is_docker_based:
-            from .models import ChallengeEvaluationCluster
+            from .models import ChallengeEvaluationCluster, ChallengePhase
 
+            challenge_phases = ChallengePhase.objects.filter(
+                challenge=challenge.pk
+            )
+            efs_id = ""
+            if challenge_phases.count() > 0:
+                efs_id = challenge_phases[0].efs_id
```
RishabhJain2018, line 451: Why are we taking only the `0`th index?

apps/challenges/aws_utils.py
```diff
@@ -1242,13 +1262,17 @@ def create_eks_cluster_subnets(challenge):
...
+    from .serializers import (
+        ChallengeEvaluationClusterSerializer,
+        ChallengePhaseCreateSerializer,
+    )
     from .utils import get_aws_credentials_for_challenge
 
     for obj in serializers.deserialize("json", challenge):
         challenge_obj = obj.object
     challenge_aws_keys = get_aws_credentials_for_challenge(challenge_obj.pk)
+    environment_suffix = "{}-{}".format(challenge_obj.pk, settings.ENVIRONMENT)
```
RishabhJain2018, line 1275: Let's also add the phase pk so that we know which phase does it correspond to.

apps/challenges/models.py
```diff
@@ -167,6 +167,8 @@ def __init__(self, *args, **kwargs):
     desired_worker_instance = models.IntegerField(
         null=True, blank=True, default=1
     )
+    # Enable/Disable EFS creation for the challenge
+    enable_efs = models.BooleanField(default=False)
```
RishabhJain2018, line 171: We should also move this to the challenge phase, so suppose if there is a case where one phase requires the `EFS` and other doesn't we would end up creating `EFS` for both the phases. What do you think?

RishabhJain2018 (changes requested):

apps/challenges/aws_utils.py
Ram81: For now, we want to just set up 1 EFS per challenge right?

apps/challenges/models.py
Ram81: We can move it to the challenge phase but if EFS is not being used there won't be any cost associated with it https://aws.amazon.com/efs/pricing/.

Ram81: If we are moving enable flag to the phase model we should move all other meta data like security_group, mount_targets, etc. But I wanted to avoid storing cluster related details in phase model

apps/challenges/aws_utils.py
Ram81: We can't add this unless we switch to 1 EFS per phase in this PR itself

Ram81:
@RishabhJain2018 you can review this now.

apps/challenges/aws_utils.py
```diff
@@ -1327,6 +1336,54 @@ def create_eks_cluster_subnets(challenge):
         )
         security_group_id = response["GroupId"]
 
+        response = client.create_security_group(
+            GroupName="evalai-code-upload-challenge-efs-{}".format(
+                environment_suffix
+            ),
+            Description="EKs nodegroup EFS",
```
RishabhJain2018, line 1343: `EKs` -> `EKS`

RishabhJain2018 (changes requested):
@Ram81 Minor comment, rest looks good!

Ram81:
@RishabhJain2018 this PR is ready

RishabhJain2018 (approved):

PR #3299, opened 2021-03-10
show private submission toggle on leaderboard even if no submissions are available

### LINKED ISSUE #3296: Host can't see private submissions toggle not available when no public submissions are available on leaderboard
When the host goes to the leaderboard, he is not able to see the private submissions at first. And if no public submissions are available, then it directly shows that no results are available. The toggle to include/exclude private submissions does not appear in this case. Therefore, if for a particular challenge all the submissions are private, then the host would see no submissions in the leaderboard, and nor would he get a toggle to include private submissions in his view.
Therefore, the toggle to include/exclude private submissions also be available if `leaderboard.length === 0`

burnerlee:
resolves #3296


burnerlee:
@Ram81 up for review

Ram81:
@burnerlee can you fix the failing build for the PR

burnerlee:
@Ram81, the Travis build seems to fail due to some other error. I had run all the tests before pushing, and all of them passes. Also, the Travis logs show that it fails to due failure in step -> docker-compose build. But it builds without any error

burnerlee:
@Ram81 @RishabhJain2018 rebased on master

Issue #3301, opened 2021-03-11
Show fargate worker status on manage tab


Ram81:
### Description

Using the manage tab challenge hosts can start/stop/restart the challenge workers but they don't get any feedback from EvalAI when worker started/stopped/restarted successfully other than notification when action was executed. As fargate worker takes a few minutes to complete start/stop/restart hosts don't know the status of the challenge worker. We want to show status of fargate worker for each challenge in the manage tab. This status will be updated every 30 seconds (poll duration can be discussed), we will need a polling setup to poll the worker status using boto3 APIs.

### Expected deliverables

1. Polling setup to fetch challenge worker status using boto3 APIs
2. APIs on the backend to fetch worker status
3. Frontend changes to show status on the manage tab

burnerlee:
@Ram81 could I please work on the issue?

Ram81:
@burnerlee  sure, go ahead

burnerlee:
@Ram81 how do I interact with the containers in the development environment for testing?

burnerlee:
also @Ram81, I think this is the function I would need to fetch the details. https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/eks.html#EKS.Client.describe_fargate_profile. Could you please confirm it once 😅 
![Screenshot from 2021-03-11 20-54-01](https://user-images.githubusercontent.com/55936223/110810579-efda1f00-82ab-11eb-8a84-b21dc2927d71.png)

burnerlee:
I am planning to call it repeatedly after an interval of poll-time similar to what is done to fetch logs

Ram81:
>@Ram81 how do I interact with the containers in the development environment for testing? It says that I can't perform actions on container in the development mode

@burnerlee you won't be able to test fargate setup locally. You can look at boto3 APIs and code it up. Once you have a PR ready I can help you with testing.

>also @Ram81, I think this is the function I would need to fetch the details. https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/eks.html#EKS.Client.describe_fargate_profile. Could you please confirm it once 😅

[...]

PR #3305, opened 2021-03-11
Frontend: Disable the start/stop/restart buttons avoiding multiple api calls.

savish28:
This is in reference to the suggestion discussed in https://github.com/Cloud-CV/EvalAI/pull/2918#issuecomment-677794344.

Relevent/Related Issues - #2918, #3301.

Currently, the restart/start/stop buttons on the Manage Worker are active at all times which results in multiple API calls of restarting and starting.

Now, while the event of restarting or starting is in progress these buttons are disabled.

restart_start_disable_eval

### DIFF
--- a/frontend/src/js/controllers/challengeCtrl.js
+++ b/frontend/src/js/controllers/challengeCtrl.js
@@ -10,6 +10,7 @@
 
     function ChallengeCtrl(utilities, loaderService, $scope, $state, $http, $stateParams, $rootScope, $interval, $mdDialog, moment, $location, $anchorScroll, $timeout) {
         var vm = this;
+        vm.disableWorkerButtons = false;
         vm.areSubmissionsFailing = false;
         vm.getAllEntriesTestOption = "Include private submissions";
         vm.showPrivateIds = [];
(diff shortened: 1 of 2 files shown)


savish28:
@RishabhJain2018. Could you please review the change.

savish28:
@RishabhJain2018 Ping.

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -150,11 +151,13 @@
         // API call to manage the worker from UI.
         // Response data will be like: {action: "Success" or "Failure", error: <String to include only if action is Failure.>}
         vm.manageWorker = function(action){
+            if (action == "start" || action == "restart") vm.disableWorkerButtons = true;
```
Ram81, line 154: @savish28 why are we not checking for stop action?

Ram81 (changes requested):

frontend/src/js/controllers/challengeCtrl.js
savish28: @Ram81 because the motivation behind this PR is this suggestion https://github.com/Cloud-CV/EvalAI/pull/2918#issuecomment-677794344.

savish28: We can add the stop action as well if you suggest. I thought that Starting and Restarting are heavy operations compared to stopping.

Ram81: @savish28 I think it was not clear in that issue but the idea is to disable the start/stop/restart buttons based on status of the workers that will be fetched using boto3 API calls. Idea is to fetch status of all running workers for a challenge and use the status of those workers to disable these action buttons. So this PR will need changes which are dependent on this #3308 PR. Once it is merged you can make the required changes.

Ram81:
Closing the PR as this is not the expected solution for the issue

PR #3308, opened 2021-03-13
fix #3301: show worker status of manage challenge tab

### LINKED ISSUE #3301: Show fargate worker status on manage tab

burnerlee:


burnerlee:
@Ram81 PR up for review. I'll need your help in testing

burnerlee:
@Ram81 ping

apps/challenges/aws_utils.py
```diff
@@ -1067,6 +1067,38 @@ def get_logs_from_cloudwatch(
...
+    client = get_boto3_client("ecs", aws_keys)
+    worker_status = {}
+    if settings.DEBUG:
+        worker_status = {
+            "error": False,
+            "details": "Not available for development environment"
+        }
+    else:
+        try:
+            response = client.describe_tasks(
```
Ram81, line 1083: @burnerlee I don't think this is the correct API. This will give the status of task definition right? for each task definition there are containers/instances running. What we need is the status of those containers/instances. So the output of API should be list of `task_ids` (i.e. container ids) and `status` (status will be one of `RUNNING`, `PENDING`, etc). Can you please check the ECS APIs to see if there is an API that we can use. Let me know if you need help

Ram81 (changes requested):

apps/challenges/aws_utils.py
burnerlee: @Ram81 in this API also returns the data of each container running for each task definition including the status for each container. So do I need to show the status of each container on the frontend?

Ram81: Yes, currently we only spawn 1 container per challenge but we want this feature to work for multiple containers too.

apps/challenges/aws_utils.py
```diff
@@ -1067,6 +1067,38 @@ def get_logs_from_cloudwatch(
...
+                tasks=[
+                    task_def_arn
```
Ram81, line 1085: @burnerlee this API won't work directly for us. We store task definition for each service in our challenge model. To get the list of tasks for which we want the status you have to first store `serviceName` in our model, this can be done while creating the fargate service for each challenge. You can look at boto3 API call in `start_workers` and make required changes to store service name in our models. 

Once you have the service name you can use `client.list_tasks` to list the tasks that are active in  the service and fetch status for each task from that response using `client.describe_task` API.

Ram81 (changes requested):
@burnerlee let me know if you have any doubts

Ram81:
@burnerlee any updates on the requested changes?

burnerlee:
Sorry for the delay.. I'll make this change

burnerlee:
@Ram81 if multiple tasks are deployed for a single challenge, then the field `task_def_arn` would become of no use for a challenge as now we would have multiple of them and we can store the `service_name` instead only?

burnerlee:
@Ram81 changed the data fetching API. Please check if the data being returned by the backend is correct as per the requirement now.... Once verified, I'll write the code for frontend template to render this

Alabhya268:
@burnerlee @Ram81, Can this not be done with EC2 as client? There is a method available that returns worker status [describe_instance_status](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2.html#EC2.Client.describe_instance_status).
I'm still learning about this topic so please don't mind if it's wrong.

burnerlee:
Hey @Alabhya268 actually what we have in the database for our challenge is a task def and servicename, so we need to find status of containers inside a task for each task, which can be more easily done by the api used in the PR due to the type of parameters it accepts and responses it provides 😅.  Thanks for your suggestion ❤️
Also @Alabhya268 you asked me something regarding django migrations. I had replied in the thread but the issue was closed after an hour I replied. Just checking if you didn't miss it😅. Otherwise I'll share the details again

burnerlee:
> @Ram81 changed the data fetching API. Please check if the data being returned by the backend is correct as per the requirement now.... Once verified, I'll write the code for frontend template to render this

@Ram81 ping

Alabhya268:
> Hey @Alabhya268 actually what we have in the database for our challenge is a task def and servicename, so we need to find status of containers inside a task for each task, which can be more easily done by the api used in the PR due to the type of parameters it accepts and responses it provides . Thanks for your suggestion 
> Also @Alabhya268 you asked me something regarding django migrations. I had replied in the thread but the issue was closed after an hour I replied. Just checking if you didn't miss it. Otherwise I'll share the details again

Thanks for clearing my doubts also thanks for suggestion I have already seen it.

Ram81 (changes requested):
@burnerlee can you fix the failing build

apps/challenges/aws_utils.py
```diff
@@ -531,6 +531,8 @@ def create_service_by_challenge_pk(client, challenge, client_token):
 
     queue_name = challenge.queue
     service_name = "{}_service".format(queue_name)
+    challenge.service_name = service_name
```
Ram81, line 534: @burnerlee this looks incorrect to me. The service name will be returned by the `start_workers` boto3 API call. Service name is a combination of queue name, challenge pk and a random string. Can you share how you verified that the service name is `{queue_name}_service` ?

Ram81 (changes requested):

apps/challenges/aws_utils.py
Ram81: @burnerlee I think it wasn't clear in my message. I was asking from where did you verify that `_service` is appended at the end of the service name. If we are using it while creating a service we should create a utils method which takes challenge model object and returns service name so that there is no bug somewhere in the codebase

Ram81: As we might use service name at multiple places it would be good to have a single method which can be used to fetch service name

burnerlee: oh i'll just mention that line here

apps/challenges/aws_utils.py
```diff
@@ -531,6 +531,8 @@ def create_service_by_challenge_pk(client, challenge, client_token):
 
     queue_name = challenge.queue
     service_name = "{}_service".format(queue_name)
```
burnerlee, line 533: @Ram81 we get the service name from here and then pass this in the `create_service` boto api call

burnerlee:
> @burnerlee can you fix the failing build

yeah the failing build is due to some conflict in django migrations after i rebased it on master. On it

apps/challenges/aws_utils.py
burnerlee: @Ram81 yeah makes sense. I can add a utils function for this.

burnerlee: I have fixed the build fail and saved the service name from api response for now after the latest change. If you say, I would revert to the previous method and make a utils function as well :smile:

apps/challenges/aws_utils.py
```diff
@@ -1067,6 +1068,51 @@ def get_logs_from_cloudwatch(
...
+    client = get_boto3_client("ecs", aws_keys)
+    tasks_status = {}
+    if settings.DEBUG:
+        tasks_status = {
+            "error": False,
+            "details": "Not available for development environment"
+        }
+    else:
+        try:
+            tasks_response = client.list_tasks(
```
Ram81, line 1084: @burnerlee let's split this into two utils methods. First one will list tasks for a service and second method will take list of task arns as input and return response of `describe_tasks`

apps/challenges/models.py
```diff
@@ -131,6 +131,10 @@ def __init__(self, *args, **kwargs):
     task_def_arn = models.CharField(
         null=True, blank=True, max_length=2048, default=""
     )
+    # The service name for the challenge which contains all the tasks
```
Ram81, line 134: Change `The service name for the challenge which contains all the tasks` -> `Challenge evaluation worker service name`

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -203,6 +203,35 @@
...
+                parameters.data = {};
+                parameters.callback = {
+                    onSuccess: function(response) {
+                        var details = response.data;
+                        vm.workerStatus = details.tasks_status;
+                    },
+                    onError: function(response) {
+                        var error = response.data.error;
+                        if (error == undefined){
+                            $rootScope.notify("error", "There was an error.");
```
Ram81, line 220: @burnerlee please make the error message more clear. For ex: `An error occurred when fetching worker status` or `Worker fetch status failed`

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -203,6 +203,35 @@
...
+                        }
+                        else {
+                            $rootScope.notify("error", "There was an error: " + error);
```
Ram81, line 223: same here

Ram81, line 223: change to `} else {`

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -203,6 +203,35 @@
             $interval.cancel(vm.logs_poller);
         };
 
+         // Get the status of workers continuously
+         vm.startFetchingWorkerStatus = function() {
```
Ram81, line 207: Change method name to `fetchWorkerStatus`

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -203,6 +203,35 @@
             $interval.cancel(vm.logs_poller);
         };
 
+         // Get the status of workers continuously
```
Ram81, line 206: Change to `Poll evaluation worker status at interval x`

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -203,6 +203,35 @@
...
+                        else {
+                            $rootScope.notify("error", "There was an error: " + error);
+                        }
+                    }
+                };
+                utilities.sendRequest(parameters);
+            }, 5000);
+        };
+
+        vm.stopFetchingWorkerStatus = function(){
```
Ram81, line 231: change to `stopWorkerStatusFetchPoller`

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -2597,6 +2626,7 @@
             vm.stopFetchingSubmissions();
             vm.stopLeaderboard();
             vm.stopLoadingLogs();
+            vm.stopFetchingSubmissions();
```
Ram81, line 2629: @burnerlee this looks like an incorrect method call? Why do we have to stop fetching submissions?

frontend/src/views/web/challenge/manage.html
```diff
@@ -12,6 +12,11 @@ <h5 class="w-300">Submission worker actions</h5>
                 <button class="btn ev-btn-dark waves-effect waves-dark grad-btn grad-btn-dark fs-14" ng-click="challenge.manageWorker('restart');"><i class="fa fa-refresh"></i> Restart worker</button>
             </div>
         </div>
+        <div class="row margin-bottom-cancel">
+            <div class="col s12">
+                <h5 class="w-300">Submission worker status - {{challenge.workerStatus?challenge.workerStatus:"Loading..."}}</h5>
```
Ram81, line 17: @burnerlee can we should color coded status here. 

1. RUNNING - Green
2. PENDING - Orange
3. STOPPED - RED

Ram81 (changes requested):

burnerlee:
@Ram81 made all the  backend changes. I need to discuss some points before finalising the code for frontend. 
For each challenge, we can have multiple tasks, and for each task we can have multiple containers. Currently, our backend api returns the status for each container of each tasks. So let us say we have 2 tasks running for a challenge, and each task have 5 containers. So our backend api would give us 10 status for each of container along with its containerArn and taskArn. What would be the format to display the status of each container?
Also, I would need your help to test this code if the logs are received flawlessly 😄

PR #3310, opened 2021-03-14
Backend: Add LetterCase-Independent filename validation in Challenge yml validation

savish28:
Current Scenario:- 
- Images with not lowercase 'JPG', 'PNG', 'JPeg' etc are ignored.
- The HTML template is only considered ending with '.html' but '.HTML' are ignored and are treated as error.
- Challenge Config yml, yaml are also considered only when the extension is completely lowercase and other cases gives invalid config.

The changes proposed solves the above issues.

### DIFF
--- a/apps/challenges/challenge_config_utils.py
+++ b/apps/challenges/challenge_config_utils.py
@@ -52,9 +52,9 @@ def get_yaml_files_from_challenge_config(zip_ref):
     yaml_file_name = None
     extracted_folder_name = None
     for name in zip_ref.namelist():
-        if (name.endswith(".yaml") or name.endswith(".yml")) and (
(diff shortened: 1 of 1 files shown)


savish28:
@RishabhJain2018 @Ram81 Could you please review the changes.

savish28:
@Ram81 @RishabhJain2018 Could you please review the changes.

savish28:
@Ram81 Ping, The GitHub challenge upload API validation ignores files that do not have a lowercase extension. Like HTML, PNG etc files. Could you please have a look.

Ram81:
@savish28 we don't need this feature. This seems like unnecessary validation. Closing the PR

PR #3316, opened 2021-03-15
[Backend] Fix #3313: Add start evaluation worker function call

### LINKED ISSUE #3313: [Backend] Add start evaluation worker function call in challenge creation using github API
### Description

When a challenge is created using github based challenge creation pipeline, follow the steps mentioned [here](https://github.com/Cloud-CV/EvalAI-starters#create-challenge-using-github) to use the set up, we don't automatically start evaluation workers for the challenge. We should make a `start_worker` celery task call when a new challenge is created using github based set up. [Here's](https://github.com/Cloud-CV/EvalAI/blob/master/apps/challenges/views.py#L1472) the reference to how we are doing it for challenge creation using zip file.

### Deliverable

- [x] Backend API changes to trigger `start_workers` celery task when challenge is created using github based setup.

afif-fahim:
Fixes #3313

* Added start evaluation worker function call when creating challenge using create_or_update_github_challenge API

### DIFF
--- a/apps/challenges/views.py
+++ b/apps/challenges/views.py
@@ -3229,6 +3229,35 @@ def create_or_update_github_challenge(request, challenge_host_team_pk):
                         }
                         send_slack_notification(message=message)
 
+                    template_data = get_challenge_template_data(zip_config.challenge)
+                    if (
+                        not challenge.is_docker_based
+                        and challenge.inform_hosts
+                        and not challenge.remote_evaluation
+                    ):
+                        try:
+                            response = start_workers([zip_config.challenge])
+                            count, failures = response["count"], response["failures"]
+                            logging.info(
+                                "Total worker start count is {} and failures are: {}".format(
+                                    count, failures
+                                )
+                            )
+                            if count:
+                                logging.info(
+                                    "{} workers started successfully".format(count)
+                                )
+                                template_id = settings.SENDGRID_SETTINGS.get(
+                                    "TEMPLATES"
+                                ).get("WORKER_START_EMAIL")
+                                send_emails(emails, template_id, template_data)
+                        except Exception:
+                            logger.exception(
+                                "Failed to start workers for challenge {}".format(
+                                    zip_config.challenge.pk
+                                )
+                            )
+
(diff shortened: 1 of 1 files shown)


apps/challenges/views.py
```diff
@@ -3229,6 +3229,35 @@ def create_or_update_github_challenge(request, challenge_host_team_pk):
                         }
                         send_slack_notification(message=message)
 
+                    template_data = get_challenge_template_data(zip_config.challenge)
+                    if (
+                        not challenge.is_docker_based
+                        and challenge.inform_hosts
+                        and not challenge.remote_evaluation
```
Ram81, line 3236: @afif-fahim we need to add a check to start worker only if this API was called for the first time. You can use the value of the number of workers in challenge model to figure out if we have already started the worker for a challenge or not. If value is not null then we don't need to call the `start_worker` method

Ram81 (changes requested):

apps/challenges/views.py
afif-fahim: @Ram81 can you review again please?

apps/challenges/views.py
```diff
@@ -3229,6 +3229,36 @@ def create_or_update_github_challenge(request, challenge_host_team_pk):
                         }
                         send_slack_notification(message=message)
 
+                    template_data = get_challenge_template_data(zip_config.challenge)
+                    if (
+                        not challenge.is_docker_based
+                        and challenge.inform_hosts
```
Ram81, line 3235: @afif-fahim this check is only neede when sending emails. if `inform_hosts` is false we will start the workers but we will not send any email to update the challenge host. This check should go on line `3251`

Ram81 (changes requested):

apps/challenges/views.py
afif-fahim: @Ram81 made the change and updated the PR.

Ram81 (approved):
LGTM

afif-fahim:
@RishabhJain2018 can you merge this PR please?

PR #3317, opened 2021-03-15
Add execution time column in leaderboard

### LINKED ISSUE #3315: Add execution time of submissions on the leaderboard
**Issue:**
We want to display the execution time of the submissions for a specific phase split leaderboard.

**Tasks:**
- [ ] Add a boolean field `show_execution_time` with `default=False` in the ChallengePhaseSplit model.
- [ ] Add the field in the `ChallengePhaseSplitSerializer`.
- [ ] While filtering the query set in the `calculate_distinct_sorted_leaderboard_data` function, use the above field to annotate the query set with the `execution_time` field.
- [ ] On the UI in the `leaderboard.html`, add a new column with name `Execution time (sec.)` for the execution time before `Submitted at` and use `challenge.selectedPhaseSplit.show_execution_time` to show and hide that column with the values.

Please feel free to ask if you have any doubts/questions.

**Note:** This is a high _very priority issue_, please pick it up only when a PR can be opened within 24 hours.

burnerlee:
Add a field named show_execution_time in challengephasesplit
model. If this field is set as true, then show execution
time of the submission in the leaderboard

resolves #3315


burnerlee:
@RishabhJain2018 @Ram81 PR up for review

apps/jobs/utils.py
```diff
@@ -334,6 +334,9 @@ def calculate_distinct_sorted_leaderboard_data(
 
     all_banned_participant_team = []
     for leaderboard_item in leaderboard_data:
+        if challenge_phase_split.show_execution_time:
```
RishabhJain2018, line 337: Hey @burnerlee, Can we please add this `if` a little above i.e. before this line `leaderboard_data = leaderboard_data.annotate` and create the annotated list based on this flag. Here you are again querying the database `n` times for `n` entries in the leaderboard. This will make things slow when there is a large number of entries on the leaderboard.

frontend/src/views/web/challenge/leaderboard.html
```diff
@@ -98,6 +98,9 @@ <h5 class="w-300">Leaderboard</h5>
                                     </span>
                                 </a>
                             </td>
+                            <td ng-if="challenge.selectedPhaseSplit.show_execution_time" data-field="execution_time">
+                                    <span class="fs-20 w-300">Execution Time (sec)</span>
```
RishabhJain2018, line 102: `Execution Time (sec)` -> `Execution time (sec)`

RishabhJain2018 (changes requested):
Thanks, @burnerlee for quickly opening the PR.

Please add the sorting as well on the leaderboard.

RishabhJain2018:
@burnerlee Just wanted to check-in to ask if there are any blockers?

burnerlee:
> @burnerlee Just wanted to check-in to ask if there are any blockers?

no, was working on another project. I'll just update the PR, would be done in an hour

RishabhJain2018:
awesome, thanks!

burnerlee:
@RishabhJain2018 how do I fetch value of submission__execution_time during the annotate step? execution_time is a field defined with @property tag and I am not able to fetch it if I put "submission__execution_time" in values field

RishabhJain2018:
You will need to calculate it from `started_at` and `completed_at` fields something like this -- https://stackoverflow.com/a/52189603. Let me know if this helps.

burnerlee:
@RishabhJain2018 , thanks for responding. In the meantime, I implemented another solution, would save the extra queries. Please review this. If this seems incorrect, I'll change it accordingly

RishabhJain2018:
I looked at the solution, it seems inefficient to me because, for every API call, it will `pop` the elements from the query list. Could you please try the above solution once.

burnerlee:
> I looked at the solution, it seems inefficient to me because, for every API call, it will `pop` the elements from the query list. Could you please try the above solution once.

yeah sure

burnerlee:
@RishabhJain2018 made the change

RishabhJain2018 (approved):

RishabhJain2018:
LGTM, Thanks for adding this quickly!

Issue #3312, opened 2021-03-14
[Backend]Add boolean check in github challenge update API to avoid overwriting annotations uploaded using cli


Ram81:
### Description

Currently, when challenge hosts update the challenge using github based challenge workflow the test annotations get overwrite after each update. We also allow hosts to update annotations using EvalAI cli. When a challenge host updates the annotation using CLI we should not overwrite the annotations passed from github based challenge update API.

### Deliverable

- [ ] Add `annotations_uploaded_using_cli` field in challenge phase model
- [ ] Set `annotations_uploaded_using_cli` to `true` when annotations are uploaded using CLI
- [ ] Add check in update github based challenge API to look up whether annotations were uploaded using CLI before overwriting annotations

burnerlee:
@Ram81 can I please work on this issue?

GrimmeryGonsa:
@Ram81 is this issue still free for picking?

burnerlee:
@Ram81 ping, may I start working on the issue if required

xavierfigueroav:
@Ram81 Can I work on this issue?

PR #3322, opened 2021-03-17
Backend: Fix#3312 Overwriting annotations uploaded using cli by github api.

savish28:
Fix: #3312 .

Previous Flow:
- Github API overwrites the complete Challenge Instance and related things like phases etc.

Current Flow:
- If test annotations have been uploaded by cli in past(`annotations_uploaded_using_cli` is True) then partial updates by GitHub i.e everything except test annotations.
-  If test annotations have never been uploaded by cli in past(`annotations_uploaded_using_cli` is False) then complete update by Github Api as previous flow.

Completed the following tasks as suggested
- [x] Add `annotations_uploaded_using_cli` field in challenge phase model
- [x] Set `annotations_uploaded_using_cli` to true when annotations are uploaded using CLI
- [x] Add check in update github based challenge API to look up whether annotations were uploaded using CLI before overwriting annotations


savish28:
@Ram81 @RishabhJain2018 Could you please review the changes.

apps/challenges/views.py
```diff
@@ -3320,7 +3323,10 @@ def create_or_update_github_challenge(request, challenge_host_team_pk):
                 challenge_phase = ChallengePhase.objects.filter(
                     challenge__pk=challenge.pk, config_id=data["id"]
                 ).first()
-                if challenge_test_annotation_file:
+                if (
+                    challenge_test_annotation_file
+                    and challenge_phase.annotations_uploaded_using_cli is False
```
Ram81, line 3328: @savish28 this check will stop updating the complete challenge phase model object after annotations are uploaded using cli once right? We only want to stop overwriting annotations

apps/challenges/views.py
```diff
@@ -3330,6 +3336,19 @@ def create_or_update_github_challenge(request, challenge_host_team_pk):
                             "config_id": data["config_id"],
                         },
                     )
+                elif (
+                    challenge_test_annotation_file
+                    and challenge_phase.annotations_uploaded_using_cli
```
Ram81, line 3341: Same here. Did you test updating Phase meta fields after uploading annotations from cli?

Ram81 (changes requested):

apps/challenges/views.py
savish28: Hey, this check would update the whole challenge phase model because annotations_uploaded_using_cli is False meaning that cli has not been used to update annotations. So it is a complete update, but when annotations_uploaded_using_cli is True we want a partial update and that is done in elif statement that would stop overwriting annotations.

apps/challenges/views.py
savish28: Since this a partial update it would update all the data given except the test annotation file preventing the overwriting.

Ram81: @savish28 the update in this `elif` block just has `partial=True` where are we excluding annotations from being updated?

Ram81: Can you share how you tested this change?

Ram81:
@savish28 another part of this task is to make changes to set `annotations_uploaded_using_cli` to `True` when cli command is used to make the update. You also have to make changes in https://github.com/Cloud-CV/evalai-cli/ repo

savish28 (commented):
> @savish28 another part of this task is to make changes to set `annotations_uploaded_using_cli` to `True` when cli command is used to make the update. You also have to make changes in https://github.com/Cloud-CV/evalai-cli/ repo

Agreed, I have made the relevant changes here and opened a pull request https://github.com/Cloud-CV/evalai-cli/pull/322.

apps/challenges/views.py
savish28: final_overwrite_annotatie
@Ram81 This is the basic example in which shows the following
- `annotations_uploaded_using_cli` False then description and test annotation both are overwritten
- `annotations_uploaded_using_cli` True then description is changed but test annotation is not overwritten as desired.

savish28:
@Ram81 Ping.

apps/challenges/views.py
Ram81: @savish28 can we change `challenge_phase.annotations_uploaded_using_cli is False` to `not challenge_phase.annotations_uploaded_using_cli`

Ram81 (changes requested):
@savish28 minor change. Rest of the PR LGTM 👍

savish28:
@Ram81 Fixed the requested changes.

Ram81 (approved):
LGTM. @RishabhJain2018 can you take a look at the PR once

RishabhJain2018 (approved):

PR #3330, opened 2021-03-21
Add separate staging and production frontend v2 build

Ram81:
### Description

Pass separate configuration for staging and production build for frontend v2

### DIFF
--- a/docker/prod/nodejs_v2/Dockerfile
+++ b/docker/prod/nodejs_v2/Dockerfile
@@ -1,5 +1,6 @@
 FROM node:12.2.0 as node
 
+ARG NODE_ENV
 RUN npm install -g @angular/cli@7.3.9
 
 WORKDIR /code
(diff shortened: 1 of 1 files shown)


docker/prod/nodejs_v2/Dockerfile
```diff
@@ -7,7 +7,7 @@ COPY frontend_v2/package.json frontend_v2/yarn.lock /code/
 RUN yarn install
 ENV PATH="/code/node_modules/.bin:$PATH" 
 COPY frontend_v2/ /code
-RUN ng build --prod
+RUN ng build --prod --configuration=${NODE_ENV}
```
RishabhJain2018, line 10: Can we please define this environment variable first before using it?

RishabhJain2018 (changes requested):

docker/prod/nodejs_v2/Dockerfile
Ram81: It's already there https://github.com/Cloud-CV/EvalAI/blob/master/docker-compose-staging.yml#L69

RishabhJain2018: No, you will have to add it like this -- https://github.com/Cloud-CV/EvalAI/blob/c3898a570c5b515a2a2f2f80aa2fa2fec61a9cd0/docker/prod/nodejs/Dockerfile#L3

RishabhJain2018 (approved):

Issue #3337, opened 2021-03-24
[Frontend V2] Show execution time and submitted at for each entry on My submissions page


Ram81:
### Description

Currently, we show `Execution time` and `Submitted at` for each submission on the `My submission` tab once use clicks on the submission entry.  Here's how it looks currently

![Screen Shot 2021-03-24 at 11 52 40 AM](https://user-images.githubusercontent.com/16323427/112264854-7cff7980-8c97-11eb-87da-f68fb56936b6.png)

For each entry in the `My submissions` table we should move `Execution time` and `Submitted at` out of collapsible div and show it for each row

savish28:
@Ram81 Could I please work on this?

Ram81:
@savish28 sure, go ahead

PR #3354, opened 2021-03-24
Frontend_V2: Show execution time and submitted Time at My Submission row(#3337)

savish28:
Issue Fix: #3337
For each entry in the `My submissions` table moved `Execution time` and `Submitted at` out of collapsible div and showed it for each row

Before:
Screenshot from 2021-03-24 15-05-48

After:
Screenshot from 2021-03-24 15-03-07


savish28:
@Ram81 ping. PR ready for review.

frontend_v2/src/app/components/challenge/challengesubmissions/challengesubmissions.component.ts
```diff
@@ -151,8 +151,8 @@ export class ChallengesubmissionsComponent implements OnInit, AfterViewInit {
    */
   apiCall: any;
 
-  columnsToDisplay = ['s_no', 'status', 'submission_result_file', 'stdout_file', 'stderr_file'];
-  columnsHeadings = ['S.No.', 'Status', 'Result File', 'Stdout File', 'Stderr File', ];
+  columnsToDisplay = ['s_no', 'status', 'execution_time', 'submitted_at', 'submission_result_file', 'stdout_file', 'stderr_file'];
+  columnsHeadings = ['S.No.', 'Status', 'Execution Time', 'Submitted At', 'Result File', 'Stdout File', 'Stderr File'];
```
Ram81, line 155: @savish28 let's move `Submitted At` to the end of the table.

Ram81 (changes requested):

frontend_v2/src/app/components/challenge/challengesubmissions/challengesubmissions.component.ts
savish28: @Ram81 Done.

Ram81:
@savish28 can you share updated screenshot of the page.

savish28:
> @savish28 can you share updated screenshot of the page.

Sure, Here it is.
Screenshot from 2021-03-24 20-06-08

Ram81:
@savish26 3rd row shows `None sec` can we please fix it. If execution time is none we should show `N/A`

savish28:
> @savish26 3rd row shows `None sec` can we please fix it. If execution time is none we should show `N/A`

Yup, Now it looks like, (Attaching the screenshot).

Screenshot from 2021-03-25 20-35-39

Ram81:
@savish28 the execution time for failed submissions should also be there. Can you please confirm why the value is being populated as `None`, as we calculate execution time using `submitted_at` and `completed_at` date fields of a submission it should not be `None`.

savish28:
> @savish28 the execution time for failed submissions should also be there. Can you please confirm why the value is being populated as `None`, as we calculate execution time using `submitted_at` and `completed_at` date fields of a submission it should not be `None`.

Yes, I was also looking for the same thing. I am unsure of how the thing appeared on the first place.  But to re-create the problem  I had set the value of `submitted_at` and `completed_at` to empty from admin.

So as you suggested it would be good to safeguard this by showing N/A instead of None sec.

Ram81:
>@Ram81, I was also looking for the same thing. I am unsure of how the thing appeared on the first place. But to re-create the problem I had set the value of submitted_at and completed_at to empty from admin.

This case never happens. Isn't the data in the screenshot from the seed script? or did you make new submissions?

savish28:
@Ram81 I made a new submission. The seed data doesn't contain any None case. When in seed data I changed the start time to empty it then showed None sec.

Ram81 (approved):
LGTM 👍

PR #3365, opened 2021-03-27
Revert frontend v2 config build

Ram81:
### Description

Revert failing build for frontend v2. Logs are [here](https://travis-ci.com/github/Cloud-CV/EvalAI/builds/221383035#L6305)

### DIFF
--- a/docker/prod/nodejs_v2/Dockerfile
+++ b/docker/prod/nodejs_v2/Dockerfile
@@ -8,7 +8,7 @@ COPY frontend_v2/package.json frontend_v2/yarn.lock /code/
 RUN yarn install
 ENV PATH="/code/node_modules/.bin:$PATH" 
 COPY frontend_v2/ /code
-RUN ng build --prod --configuration=${NODE_ENV}
+RUN ng build --prod
 
 FROM nginx:1.13.12-alpine
 ARG NODE_ENV


docker/prod/nodejs_v2/Dockerfile
```diff
@@ -8,7 +8,7 @@ COPY frontend_v2/package.json frontend_v2/yarn.lock /code/
 RUN yarn install
 ENV PATH="/code/node_modules/.bin:$PATH" 
 COPY frontend_v2/ /code
-RUN ng build --prod --configuration=${NODE_ENV}
+RUN ng build --prod
```
RishabhJain2018, line 11: Isn't this the issue?

https://stackoverflow.com/questions/50333003/could-not-find-module-angular-devkit-build-angular

RishabhJain2018 (changes requested):

docker/prod/nodejs_v2/Dockerfile
Ram81: Yeah but all packages should be installed in `yarn install` right? 

Anyway, I thought we should revert this change and get the code upload worker image pushed so that habitat challenge worker can be set up. Currently, it is not even picking up new submissions and running submissions will all fail because of cilium policy being deployed on every worker restart.

Issue #3376, opened 2021-03-29
[FrontendV2] Incorrect success message on page access


ankitpal1029:
![image](https://user-images.githubusercontent.com/49815751/112798200-355b6200-908a-11eb-97a2-fbd7fbf7aa28.png)

i have set up the docker image but every time i visit a new route i get a success message with an X right next to it .This happens when logged in as a new user who's not authenticated their email id.

OS: Linux (PopOs 20.10)
Browser: Brave browser

niharsalunke:
@Ram81 may I work on this?

niharsalunke:
@ankitpal1029 are you working on this already?

Ram81:
@niharsalunke please complete #3346 before starting on another issue

niharsalunke:
Sure @Ram81. Working on it.

ankitpal1029:
@Ram81 it's to be noted that this error occurs only with a newly created user and not the default 3 users already provided

ankitpal1029:
> @ankitpal1029 are you working on this already?

@niharsalunke yes i'm working on it currently.

ankitpal1029:
@Ram81  i think this is an issue of email verification, for context frontend version 1 doesn't allow the user to access the challenges till they have verified their email and instead shows them this page,
![image](https://user-images.githubusercontent.com/49815751/112803767-46f43800-9091-11eb-9551-83a1cd67cf2c.png)
note this is after selecting a challenge and clicking on it.
Maybe we should implement something of this sort for v2 as well, what are your views on this?

ankitpal1029:
@Ram81 i think i have figured out the cause of the issue , 
on recieving a 403 error the global.service is trying to access err.error['error'] whereas
![image](https://user-images.githubusercontent.com/49815751/112862127-12a16b80-90d3-11eb-9c9c-287205b343f9.png)
 it recieves an error with field detail:
![image](https://user-images.githubusercontent.com/49815751/112862029-fd2c4180-90d2-11eb-8fd7-07c2c4681c51.png)
fixing that gives us the appropriate error message: 
![image](https://user-images.githubusercontent.com/49815751/112862276-3c5a9280-90d3-11eb-8985-f96382a3f633.png)

ankitpal1029:
@Ram81 can you take a look at my pr

ankitpal1029:
@Ram81 can you take a look at the pr

AaqibAhamed:
PR Raised

[...]

PR #3379, opened 2021-03-29
[FrontendV2] Incorrect success message on page access (#3376)

ankitpal1029:
Fixed the issue of incorrect success message appearing on page access (when the user's mail is not verified)

Issue was in the access of err.error['error'] rather than err.error['detail'] that the backend api was providing.

after fixing the issue an error message of this form appears (which is what we desire):
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -66,4 +66,4 @@ ssl/
 *.iws
 out/
 .idea_modules/
-
+Session.vim
--- a/frontend_v2/src/app/services/global.service.ts
+++ b/frontend_v2/src/app/services/global.service.ts
@@ -439,7 +439,7 @@ export class GlobalService {
     if (err.status === 401) {
       this.checkTokenValidity(err, toast);
     } else if (err.status === 403 && toast) {
-      this.showToast('error', err.error['error'], 5);
+      this.showToast('error', err.error['detail'], 5);
     } else if (err.status === 404 && toast) {
       this.showToast('error', err.error['detail'], 5);
     } else if (err.status === 406 && toast) {


ankitpal1029:
@Ram81  can you review this , let me know if this is the approach we are going to take or just block out these routes entirely for unverified users

.gitignore
```diff
@@ -66,4 +66,4 @@ ssl/
 *.iws
 out/
 .idea_modules/
-
+Session.vim
```
Ram81, line 69: @ankitpal1029 please remove this

frontend_v2/src/app/services/global.service.ts
```diff
@@ -439,7 +439,7 @@ export class GlobalService {
     if (err.status === 401) {
       this.checkTokenValidity(err, toast);
     } else if (err.status === 403 && toast) {
-      this.showToast('error', err.error['error'], 5);
+      this.showToast('error', err.error['detail'], 5);
```
Ram81, line 442: @ankitpal1029 we should check if `error` key exists in the error response. If it does, then we show that in the toast. If not, then we look for `detail` key and show that

Ram81 (changes requested):
@ankitpal1029 can you share a screenshot of the error message after the change

Ram81:
@ankitpal1029 we should use the same approach we use in frontend. We don't allow challenge page access if the email is not verified. You can look at frontend v1 to check exact behavior for all pages

ankitpal1029:
@Ram81 that works , i'll replicate that behaviour and make a new pr

.gitignore
ankitpal1029: Session.vim is autogenerated when using neovim much like .vscode

PR #3382, opened 2021-03-31
fix #3350: fix multiple simultaneous request to fetch leaderboard data and stop polling on tab change

### LINKED ISSUE #3350: [Frontend V2] Stop leaderboard polling when user switches to another tab or page
### Description

Currently, after a user visits the leaderboard page in frontend v2 the polling for updating leaderboard is not stopped. Because of this issue user experience is bad and it slows down UI access until you refresh the page.

burnerlee:
currently on frontend_v2, multiple request are made on the leaderboard page every 5 seconds to fetch the leaderboard data. This is because when user selects one phasesplit, then he is sent to `leaderboard/phase_split_id`.. But the other parts of `phaseSplitSelected()` keep on executing and start polling. Also, polling is started by the '/leaderboard/80` so multiple requests are made on each poll. Also, we change page from `leaderboard/phase_split_id`, only the polling of this page stops and not the one which started on `/leaderboard`. Hence, this keeps on polling data unnecessarily making the UI slow
resolves #3350

### DIFF
--- a/frontend_v2/src/app/components/challenge/challengeleaderboard/challengeleaderboard.component.ts
+++ b/frontend_v2/src/app/components/challenge/challengeleaderboard/challengeleaderboard.component.ts
@@ -335,7 +335,7 @@ export class ChallengeleaderboardComponent implements OnInit, AfterViewInit, OnD
         SELF.router.navigate(['../../' + phaseSplit['id']], { relativeTo: this.route });
       }
       SELF.selectedPhaseSplit = phaseSplit;
-      if (SELF.selectedPhaseSplit) {
+      if (SELF.selectedPhaseSplit && !SELF.router.url.endsWith('leaderboard')) {
         SELF.fetchLeaderboard(SELF.selectedPhaseSplit['id']);
         SELF.showLeaderboardByLatest = SELF.selectedPhaseSplit.show_leaderboard_by_latest_submission;
         SELF.sortLeaderboardTextOption = SELF.showLeaderboardByLatest ? 'Sort by best' : 'Sort by latest';


burnerlee:
before the fix:
#3350error
after fixing:
#3350output

burnerlee:
@Ram81 please review

frontend_v2/src/app/components/challenge/challengeleaderboard/challengeleaderboard.component.ts
```diff
@@ -335,7 +335,7 @@ export class ChallengeleaderboardComponent implements OnInit, AfterViewInit, OnD
         SELF.router.navigate(['../../' + phaseSplit['id']], { relativeTo: this.route });
       }
       SELF.selectedPhaseSplit = phaseSplit;
-      if (SELF.selectedPhaseSplit) {
+      if (SELF.selectedPhaseSplit && !SELF.router.url.endsWith('leaderboard')) {
```
Ram81, line 338: @burnerlee this check is to make sure the leaderboard is fetched only when the user goes to the leaderboard page right? I don't understand why it is `!` of routes ending with `leaderboard`. Can you please clarify

Ram81 (changes requested):

frontend_v2/src/app/components/challenge/challengeleaderboard/challengeleaderboard.component.ts
burnerlee: @Ram81 yeah sure..  What happens currently is that initially I go to the `/leaderboard` page. When I select a phase split, the function `phaseSplitSelected` is called. In this function, we start a poller to start fetching the leaderboard data (let's call this poller1) and then we redirect ourselves to `/leaderboard/phaseSplitId` where the same page is rendered and `phaseSplitSelected` is called again. Therefore this starts another poller (poller2). Now, if you see the first GIF attached you would see that there are 2 requests polled every 5 seconds currently before the fix. This is due to both the pollers. And then when we leave this page (`/leaderboard/phaseSplitId`) the poller for this page (poller2) is stopped. But poller1 is never stopped and it keeps on fetching leaderboard data. So I have added a condition that a poller is only started when we are not on the leaderboard page., so that only poller2 in the above case is started. Hence, in the second GIF attached, you can see that only 1 request is made for each poll and also it stops as soon as we exit that page

Ram81: @burnerlee then I think the check should be `url.contains('leaderboard')` and doesn't end with it. Did you also check if the polling is not started on any other tabs?

burnerlee: @Ram81, actually we want the poller to start when url is something like `localhost:4200/challenge/20/leaderboard/80` and to not start when it is `localhost:4200/challenge/20/leaderboard`. So I feel endsWith would work here. 
Yes, I checked it. Actually for the poller to start there must be 2 conditions. A phaseSplit has to be selected in the leaderboard page and the url of the page at that time must not end with `leaderboard`

ShauryaAg: @burnerlee Wouldn't it make more sense to check if the current tab is in focus, rather than checking the URL?
Something like [this](https://stackoverflow.com/questions/1760250/how-to-tell-if-browser-tab-is-active)

burnerlee: @shauryaag actually the tab is active in both the conditions 😅 maybe you can try to reproduce the issue by running in your pc.. you'll get an insight on what's causing the multiple requests

ShauryaAg:
@burnerlee After some testing, the poller does not start at `/leaderboard` endpoint, it only starts after we go to `/leaderboard/<phaseSplitId>`

The multiple simultaneous request are because `selectPhaseSplit` method from `selectphase` component is getting called twice: [here](https://github.com/Cloud-CV/EvalAI/blob/master/frontend_v2/src/app/components/challenge/challengeleaderboard/challengeleaderboard.component.ts#L307) and [here](https://github.com/Cloud-CV/EvalAI/blob/master/frontend_v2/src/app/components/utility/selectphase/selectphase.component.html#L76).

I tried removing the `selectPhaseSplit` method call from `challengeleaderboard.component.ts`, that fixes the multiple requests issue, but then the page is not rendered with desired phaseSplit leaderboard on the first click, (it is rendered after the same phase split is selected again), which could be a state management issue which I am unable to debug at the moment.

Some help would be appreciated.

burnerlee:
@ShauryaAg hey! Yes I am getting what you're trying to do. Seems right to me too. I think the the phaseselect function trigger is not added on rendering of the leaderboard page, and did you remove my condition in case you're working on this branch?

ShauryaAg:
@burnerlee I was working on a new branch without the changes that you made previously. Please let me know if find a fix, maybe we work on this issue together.

burnerlee:
Oh I see.. @ShauryaAg can we please work on this tomorrow evening? Sadly my laptop has gone for servicing due to window blue screen error.. I'll get it back tomorrow most probably. We would have sorted it now only otherwise 😅

ShauryaAg:
Sure @burnerlee.

ShauryaAg:
@burnerlee @Ram81 I have created another PR that implements the changes I mentioned above, as I believe that is a more robust solution to the issue.

burnerlee:
@Ram81 ping

burnerlee:
@ShauryaAg why aren't we directly checking the url, rather than checking if tab is active or not 😅 and then making more changes

ShauryaAg:
@burnerlee say if the URL were to change in the future, we'd also need to change this condition or if multiple tabs were open with the same URL or if some other site was open where the URL also ends with `/leaderboard`.
Putting a check to see if the tab is active or not seems like a better option to me.

burnerlee:
@ShauryaAg actually this condition is only checked when a phase is selected on this page. So there won't be a problem anywhere else. And also, if you see the code just above the change I made, there is the same condition placed for checking something else. So I feel it won't be a problem. [For your reference]: https://github.com/Cloud-CV/EvalAI/blob/6a872385c01d3be150b3486c83eedcc91c7b47b4/frontend_v2/src/app/components/challenge/challengeleaderboard/challengeleaderboard.component.ts#L330

ShauryaAg:
@burnerlee I checked the changes you made, it doesn't stop polling when the user changes the tab. As you mentioned: it only checks when a phase is selected.
The condition to stop polling when the user changes the tab should be placed right before the API call is made.

burnerlee:
Can you please share a video showing that polling is not stopped. I tested it. It wad working fine for me

ShauryaAg:
@burnerlee Sure.
As you can see I changed the tab when there were only 2 requests, waited for about 10 secs, went back and there were 2 more requests (total 4). Those 2 requests should not have been made while I was on the other tab.

https://user-images.githubusercontent.com/31778302/113517817-ea68af80-959f-11eb-8579-4059753658b8.mp4

Ram81:
@burnerlee is this PR ready for review?

burnerlee:
Yes yes, I was waiting for the review.. you can test it.. works fine for me😀

Ram81 (changes requested):
@burnerlee this change doesn't fix the issue. I am seeing leaderboard API being called 4 times every 5 seconds. Steps I did:

1. Go to leaderboard page switching multiple tabs
2. Go to participated challenges
3. Go back to the same challenge page
4. Go to submit page
5. Make a submission
6. Go to the leaderboard page and check network console

ShauryaAg:
@Ram81 #3392 does fix the issue. Can you please check it out?

PR #3384, opened 2021-03-31
Frontend_V2:Incorrect success message on page access(3376)

### LINKED ISSUE #3376: [FrontendV2] Incorrect success message on page access

ankitpal1029:
Fixes #3376 
The incorrect error message behaviour was due to the user having their email not verified, 
this was fixed by redirecting the user to the permission-denied route upon recieving a status code of 403 
from the API,

Behaviour after commit (for unverified users):

https://user-images.githubusercontent.com/49815751/113205526-0280bc00-928c-11eb-80a9-f93037aed253.mp4

### DIFF
--- a/frontend_v2/src/app/services/global.service.ts
+++ b/frontend_v2/src/app/services/global.service.ts
@@ -1,5 +1,6 @@
 import { Injectable, Output, EventEmitter } from '@angular/core';
 import { BehaviorSubject } from 'rxjs';
+import { Router } from '@angular/router';
 
 @Injectable()
 export class GlobalService {
@@ -84,7 +85,7 @@ export class GlobalService {
   /**
    * constructor
    */
-  constructor() {}
+  constructor(private router: Router) {}
 
   /**
    * Update Scrolled State.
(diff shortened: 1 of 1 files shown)


ankitpal1029:
@Ram81 please review the pr

ankitpal1029:
@Ram81 ping to review

frontend_v2/src/app/services/global.service.ts
```diff
@@ -439,7 +440,9 @@ export class GlobalService {
     if (err.status === 401) {
       this.checkTokenValidity(err, toast);
     } else if (err.status === 403 && toast) {
-      this.showToast('error', err.error['error'], 5);
+      console.log(err.error);
```
Shekharrajak, line 443: It is better to remove the console log before commit that may slow down the browser tab and I think we do not want to show the internal logs publicly.

PR #3422, opened 2021-05-18
Fix 1800: Load challenge data in the process on demand

### LINKED ISSUE #1800: Load challenge data in the process when challenge is created

barrywhart:
Resolves #1800

### DIFF
--- a/scripts/workers/submission_worker.py
+++ b/scripts/workers/submission_worker.py
@@ -385,6 +385,10 @@ def run_submission(
 
     submission_output = None
     phase_id = challenge_phase.id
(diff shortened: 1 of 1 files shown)


RishabhJain2018:
Closing this since it is a very old PR.

PR #3424, opened 2021-05-22
Fix team filter on All submissions page

gautamjajoo:
**Problem**: When we filter the team name on all submission page, the data is different than expected. It is because whichever api call is taking the most time that is being displayed on the page.

**Fix**:

1. Fixed it by checking if the api call for data displayed is same as entered.
2. The pagination of filter submissions is also fixed. Since the pagination number is taken from the URL and when we filter the team name, the URL changes as well.

https://user-images.githubusercontent.com/24366008/119237574-2b6e5080-bb5b-11eb-8bbb-5d5a674dc281.mp4

@RishabhJain2018 @Ram81 @Kajol-Kumari


Kajol-Kumari (changes requested):
[Issue] Tho we are able to show consistent data to the users but we aren't avoiding the redundant api calls which we could have achieved using `switchMap` @gautamjajoo 

\cc @Ram81

frontend_v2/src/app/components/challenge/challengeviewallsubmissions/challengeviewallsubmissions.component.ts
```diff
@@ -253,55 +259,69 @@ export class ChallengeviewallsubmissionsComponent implements OnInit, AfterViewIn
...
-          SELF.paginationDetails.currentPage = 1;
-        } else {
-          SELF.paginationDetails.isNext = '';
-          SELF.paginationDetails.currentPage = Math.ceil(data.next.split('page=')[1] - 1);
-        }
-        if (data.previous === null) {
-          SELF.paginationDetails.isPrev = 'disabled';
-        } else {
-          SELF.paginationDetails.isPrev = '';
+        if(name == SELF.filterSubmissionsQuery) {
```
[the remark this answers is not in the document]
gautamjajoo: Actually, the data we get is of the latest API call, therefore using this condition it ignores the previous API calls. 
Since, we would be using `switchMap` to reduce unnecessary API calls, this would not be used then.

gautamjajoo:
@Ram81 @Kajol-Kumari Reverted back the changes. We are now ignoring the previous API calls. Also, added it to the `participant-team` name filter.

frontend_v2/src/app/components/challenge/challengeviewallsubmissions/challengeviewallsubmissions.component.ts
```diff
@@ -130,6 +130,12 @@ export class ChallengeviewallsubmissionsComponent implements OnInit, AfterViewIn
    */
   fieldsToGetExport: any = [];
 
+  /**
+   *Check whether team name is filtered
+   */
+  isTeamFiltered: boolean = true;
```
Ram81, line 136: @gautamjajoo the logic is not clear to me. Is this flag used to check if the query matches what we passed as parameter to filter API?

Ram81 (changes requested):
@Kajol-Kumari could you test this locally once

frontend_v2/src/app/components/challenge/challengeviewallsubmissions/challengeviewallsubmissions.component.ts
gautamjajoo: So, basically the pagination number is taken from the URL and when we filter the team name, the URL changes as well, hence we use this flag to check if there is a filter added then we change our URL to a different one.

Kajol-Kumari (approved):
LGTM 👍

Ram81 (approved):
LGTM 👍

PR #3430, opened 2021-05-25
Leaderboard data is now updated when user clicks on refresh leaderboard 

gautamjajoo:
**Issue:** Leaderboard data is not updated when page is outdated and user clicks on refresh leaderboard option

**Fix:** The function was working fine, the label was actually coverred with text hence increased the `z-index` fixes the issue.

https://user-images.githubusercontent.com/24366008/119462107-039b0a80-bd5e-11eb-9323-06a1c9fee656.mp4

@Ram81 @Kajol-Kumari


frontend_v2/src/styles/base.scss
```diff
@@ -346,6 +346,7 @@ mat-chip {
 
 .update-page {
   display: block;
+  z-index: 1;
```
Kajol-Kumari, line 349: Please don't make this change in the base class, instead you should override the same class under the leaderboard component.

gautamjajoo: We are using this class only for the refresh label and we are using this class in two components where we would be facing this problem, that's why I did the changes in `base` class.

Kajol-Kumari: Okay, so if we require this change at both places, it's totally fine to change it in here.

Kajol-Kumari (approved):
LGTM 👍

Ram81 (approved):
LGTM

PR #3429, opened 2021-05-24
Add: Static code upload challenge Model changes and Message Queue Changes 

savish28:
Context: Static Code upload challenge Addition
Overview: Model changes + Message Queue Changes
Deliverables:
- [x]  Add is_static_dataset_docker_based_challenge field in challenge model.
- [x]  Add submission_input_file field in submission model.
- [x] Add is_static_code_upload_submission flag in SQS queue message for submissions made to static code upload challenge.
- [x] Worker Changes w.r.t Message body.

@KhalidRmb 
cc: @Ram81 
This PR had changes in various files so, for abstraction would create a new PR for Django admin actions.

PR #3436, opened 2021-05-29
Add: Static data set code upload challenge service and task defination with Start/stop/delete worker Action

savish28:
Context: Static Code upload challenge Addition
Overview: Addition worker in Django admin pipeline.
Depends on: #3429 
Deliverables:
- [x] Add Django admin actions to start/delete evaluation cluster for code upload challenge.
- [x] Add Task definition for static dataset code upload challenge. 

File Changed: aws_utils.py, others are part of depends on PR.

@KhalidRmb
cc: @Ram81


Ram81:
@savish28 this PR is incomplete. For the task

>  Add Django admin actions to start/delete evaluation cluster for code upload challenge.

We have to write the functions to delete the EKS cluster, EFS, EKS nodegroup, subnets and VPC and create a django admin action for complete deletion of cluster.

What your PR does is allow us to start/stop/delete just the fargate worker for all challenges. The above task will have become a large PR so you can work on it on a separate branch and open a PR later.

apps/challenges/aws_utils.py
```diff
@@ -305,6 +305,214 @@
 }}
 """
 
+task_definition_static_code_upload_worker = """
```
Ram81, line 308: @savish28 why do we need a new task definition for `static_code_upload_worker`? The task definition will be same as `code_upload_worker` task definition. If you need additional environment variables we can add it to `code_upload_worker` task def.

apps/challenges/aws_utils.py
```diff
@@ -452,23 +661,45 @@ def register_task_def_by_challenge_pk(client, queue_name, challenge):
...
-                CPU=worker_cpu_cores,
-                MEMORY=worker_memory,
-                log_group_name=log_group_name,
-                EVALAI_DNS=EVALAI_DNS,
-                EFS_ID=efs_id,
-                **COMMON_SETTINGS_DICT,
-                **challenge_aws_keys,
-            )
+            if challenge.is_static_dataset_docker_based_challenge:
+                definition = task_definition_static_code_upload_worker.format(
```
Ram81, line 665: If we are using `code_upload_worker` task def we won't need a separate config here. The cluster creation setup for static code upload challenge is exactly same as code upload worker.

Ram81 (changes requested):

apps/challenges/aws_utils.py
savish28: @Ram81 It is not just the change in env variables but also the container definitions. The static code has two containers and code upload has only one so needed to create a new defn

apps/challenges/aws_utils.py
savish28: @Ram81 Yeah, `containerDefinitions` parameter is a list and contains the container info. If we wish to use same defn for both code upload and static code upload challenge, we might need to contain {containerDefinitions} as a variable in the code upload task defn, currently, it is provided as-is. So in that case for static code upload, we would create containerDefinitions with two containers and in case of code upload pass only one container. This might be a workaround but limits the flexibility.

Ram81: I see, if that is the case then we can have separate task def for both challenge types

Ram81 (approved):
LGTM 👍

PR #3439, opened 2021-05-29
[Minor Fix] remove submission card when challenge not active

gautamjajoo:
**Issue:** 

Screenshot from 2021-05-30 02-01-01

**Fix:** Removed the unecessary submission card when challenge is not active


Ram81 (approved):
LGTM 👍

PR #3442, opened 2021-05-30
Show error message when downloading submissions that has > 5k results

gautamjajoo:
When downloading results for a phase that has > 5k results, an error is shown to use the `get_all_submissions API` to download the results. 

Also, removed the code where we are fetching the number of submissions of the participant Team on `all-submissions` page since we do not need it here.

@Ram81 @Kajol-Kumari 

Screenshot from 2021-05-31 00-39-39

### DIFF
--- a/frontend_v2/src/app/components/challenge/challengeviewallsubmissions/challengeviewallsubmissions.component.ts
+++ b/frontend_v2/src/app/components/challenge/challengeviewallsubmissions/challengeviewallsubmissions.component.ts
@@ -238,7 +238,6 @@ export class ChallengeviewallsubmissionsComponent implements OnInit, AfterViewIn
       SELF.submissionCount = 0;
       if (SELF.challenge['id'] && phase['id']) {
         SELF.fetchSubmissions(SELF.challenge['id'], phase['id']);
-        SELF.fetchSubmissionCounts(this.challenge['id'], phase['id']);
       }
     };
   }
@@ -263,6 +262,7 @@ export class ChallengeviewallsubmissionsComponent implements OnInit, AfterViewIn
     SELF.apiService.getUrl(API_PATH).subscribe(
       (data) => {
         SELF.submissions = data['results'];
+        this.submissionCount = data.count;
         let index = 0;
         SELF.submissions.forEach((submission) => {
           submission['s_no'] = index + 1;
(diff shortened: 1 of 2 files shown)


frontend_v2/src/app/components/challenge/challengeviewallsubmissions/challengeviewallsubmissions.component.ts
```diff
@@ -263,6 +262,7 @@ export class ChallengeviewallsubmissionsComponent implements OnInit, AfterViewIn
     SELF.apiService.getUrl(API_PATH).subscribe(
       (data) => {
         SELF.submissions = data['results'];
+        this.submissionCount = data.count;
```
Kajol-Kumari, line 265: Please use `SELF` instead of `this` here

frontend_v2/src/app/components/challenge/challengeviewallsubmissions/challengeviewallsubmissions.component.ts
```diff
@@ -335,17 +335,22 @@ export class ChallengeviewallsubmissionsComponent implements OnInit, AfterViewIn
...
+        else {
+          SELF.apiService.getUrl(API_PATH, false).subscribe(
+            (data) => {
+              SELF.windowService.downloadFile(data, 'all_submissions.csv');
+            },
+            (err) => {
+              SELF.globalService.handleApiError(err);
+            },
+            () => {
+              this.logger.info('Download complete.', SELF.challenge['id'], SELF.selectedPhase['id']);
```
Kajol-Kumari, line 350: Please use `SELF` instead of `this`

frontend_v2/src/app/services/endpoints.service.ts
```diff
@@ -309,7 +309,7 @@ ${phase}/submission?participant_team__team_name=${participantTeamName}`;
   }
 
   /**
-   * Challenge Submission Counts
+   * Challenge Submission Counts of the participant Team
```
Kajol-Kumari, line 312: Can we please revert this change as it's required for this PR

Kajol-Kumari (approved):
Minor changes required, rest Looks good to me 👍

frontend_v2/src/app/services/endpoints.service.ts
gautamjajoo: We were using this url in the submissions page, but this was confusing whether it was for all submissions or submissions for a particular team that's why I have modified it to make it more descriptive.

Ram81:
@gautamjajoo we should show this in a card on the `All submissions` page instead of a toast. You can highlight the card after the user clicks on the `Download` button. We will show the card on the page even if the download button is not clicked as a note if the number of submissions is > 5k. Let me know if you have doubts.

gautamjajoo:
@Ram81 The card would be like the `Submission Guidelines` one above the `Make Submission` which is on the `submit` page

gautamjajoo:
**Results > 5k**


**Results < 5k**

Ram81 (changes requested):
@gautamjajoo can you fix the message card style. It doesn't look consistent with other cards that we show to hosts on the challenge page.

gautamjajoo:
@Ram81 Updated card:

Screenshot from 2021-06-04 22-24-30

Ram81 (changes requested):
@gautamjajoo can you increase the top, bottom and left padding for the message. For style reference you can create a new challenge on frontend v1 and go to challenge overview page to see how we show instructions currently.

gautamjajoo:
@Ram81 @Kajol-Kumari 
Screenshot from 2021-06-09 14-00-16

Ram81 (approved):
LGTM

RishabhJain2018 (approved):

PR #3446, opened 2021-05-31
integrate prometheus-django library and add django as scrape target for prometheus

burnerlee:
- Integrate `django-prometheus` library on Django container. Export backend database metrics on prometheus endpoint. Metrics are available as at `/prometheus/metrics`. 
- Adds a custom database engine to add monitoring for django databases. Custom database engine was designed because we have used postgres_psycopg2 as our backend engine. Prometheus built in library for django only supports postgres. Therefore, a new custom engine was built by extending the features of prometheus exporters by a wrapper class on the base class of the currently used engine.
- Adds django as a scrape target to prometheus. Now prometheus would scrap data from `/prometheus/metrics` from django and these can be plotted in grafana. A screenshot of plot in grafana of Leaderboard get api for 2XX responses in the last 5 minutes has been attached
counter-graph


burnerlee:
prometheus-django

settings/dev.py
```diff
@@ -12,6 +12,7 @@
 DATABASES = {
     "default": {
         "ENGINE": "django.db.backends.postgresql_psycopg2",
+        "ENGINE": "django_prometheus.db.backends.postgresql",
```
Ram81, line 15: Why are we adding multiple `ENGINE` keys in the `DATABASES` config? This will break EvalAI DB access

settings/prod.py
```diff
@@ -21,6 +21,7 @@
 DATABASES = {
     "default": {
         "ENGINE": "django.db.backends.postgresql_psycopg2",
+        "ENGINE": "django_prometheus.db.backends.postgresql",
```
Ram81, line 24: Why are we adding multiple `ENGINE` keys in the `DATABASES` config? This will break EvalAI DB access

settings/test.py
```diff
@@ -5,6 +5,7 @@
 DATABASES = {
     "default": {
         "ENGINE": "django.db.backends.postgresql_psycopg2",
+        "ENGINE": "django_prometheus.db.backends.postgresql",
```
Ram81, line 8: Why are we adding multiple `ENGINE` keys in the `DATABASES` config? This will break EvalAI DB access

Ram81 (changes requested):

Ram81:
@burnerlee can you fix the failing build for this PR.

settings/dev.py
```diff
@@ -11,7 +11,7 @@
 
 DATABASES = {
     "default": {
-        "ENGINE": "django.db.backends.postgresql_psycopg2",
+        "ENGINE": "settings.custom_databases.postgres_psycopg2",
```
Ram81, line 14: @burnerlee why do we need a custom DB wrapper?

burnerlee: @Ram81 we need `django_prometheus.db.backends.postgresql_psycopg2` as our backend engine to support monitoring of database through prometheus. However django-prometheus library provides only `django_prometheus.db.backends.postgresql` as a built in engine in library. Thus to support our used engine in evalAI, i had to define a custom backend which has the properties of both `postgresql_psycopg2` and `prometheus`

Ram81:
@burnerlee scraping from django server at frequency of `30s` would create high traffic on EvalAI backend server. We need to setup push setup for API metrics instead of pull setup. Pull setup would create overhead on backend server which would require us to scale our infrastructure.

monitoring/prometheus/prometheus.yml
```diff
@@ -6,3 +6,9 @@ scrape_configs:
   - job_name: 'prometheus' 
     static_configs: 
       - targets: ['localhost:9090']
+
+  - job_name: 'django-prometheus'
+    metrics_path: '/prometheus/metrics'
+    static_configs:
+      - targets: ['django:8000']
```
Ram81, line 13: Fetch this from env file. We will use evalai endpoint as target not django container port. This won't work in case we set up prometheus on a separate container.

Ram81 (changes requested):

monitoring/prometheus/prometheus.yml
Ram81, line 13: @burnerlee why is this still `django:8000`? Can't we pull this from `docker.env`? If not, you'll have to create `prometheus_staging.yaml` and use that.

Ram81 (changes requested):
@burnerlee it's fine to use container names for dev but for staging setup we need to make sure things are setup using url configs. Monitoring setup might not be on same instance as backend server

Ram81:
@burnerlee can you fix this?

burnerlee:
@Ram81 made a new prometheus file for staging environment. Made a change to use this file in this PR.. please review this as well https://github.com/Cloud-CV/EvalAI/pull/3447

docker-compose.yml
```diff
@@ -69,7 +69,7 @@ services:
     image: prom/prometheus:latest
     user: "1000"
     volumes:
-      - ./monitoring/prometheus/prometheus.yml:/etc/prometheus/prometheus.yml
+      - ./monitoring/prometheus/prometheus_dev.yml:/etc/prometheus/prometheus.yml
```
Ram81, line 72: Rename `prometheus_dev.yaml` to `prometheus.yaml`

monitoring/prometheus/prometheus_staging.yml
```diff
@@ -0,0 +1,14 @@
+global:
+  scrape_interval: 30s
+  external_labels:
+    monitor: 'evalai-monitor'
+scrape_configs:
```
Ram81, line 5: Add newline above

monitoring/prometheus/prometheus_staging.yml
```diff
@@ -0,0 +1,14 @@
...
+  - job_name: 'prometheus' 
+    static_configs: 
+      - targets: ['https://staging.eval.ai:9090']
```
Ram81, line 8: Do you expect prometheus to also have `https://staging.eval.ai` as the domain name. This doesn't make sense. If this is the prometheus endpoint then you can use `localhost` here as only the backend servers will be on different instance

Ram81 (changes requested):
@burnerlee how is the staging docker launch picking up the `prometheus_staging.yaml`? Please add those changes as well

burnerlee:
@Ram81 I mentioned it in the above comment😅 I have made change to take that file in the staging environment in PR 3447

monitoring/prometheus/prometheus_staging.yml
burnerlee: @Ram81 I mentioned it in the above commentsweat_smile I have made change to take that file in the staging environment in PR 3447

Ram81:
@burnerlee why? the previous PR will break without changes from this PR right? If you are changing config names you should modify the usage in same PR. It's not a good practice merge code that will break.

burnerlee:
> @burnerlee how is the staging docker launch picking up the `prometheus_staging.yaml`? Please add those changes as well

@Ram81 I mentioned it in the above comment sweat_smile I have made change to take that file in the staging environment in PR 3447

burnerlee:
> @burnerlee why? the previous PR will break without changes from this PR right? If you are changing config names you should modify the usage in same PR. It's not a good practice merge code that will break.

Actually this PR does not contain the code where I define what which prometheus file has to be used in staging environment, and those changes are made in PR #3447, which is not merged in master. So maybe i should introduce the prometheus_staging.yml file in that PR instead of that PR... or other wise I'll have to rebase this branch on top of branch of that PR

monitoring/prometheus/prometheus_staging.yml
```diff
@@ -0,0 +1,15 @@
...
+
+scrape_configs:
+  - job_name: 'prometheus' 
+    static_configs: 
+      - targets: ['localhost:9090']
+
+  - job_name: 'django-prometheus'
+    metrics_path: '/prometheus/metrics'
+    static_configs:
+      - targets: ['https://staging.eval.ai:8000']
```
Ram81, line 14: @burnerlee this won't work. Our django server is behind a nginx which handles redirection. See the nginx config [here](https://github.com/Cloud-CV/EvalAI/blob/master/docker/prod/nodejs/nginx_staging.conf). You can't access django server using port on`https` endpoint

monitoring/prometheus/prometheus_staging.yml
```diff
@@ -0,0 +1,15 @@
...
+
```
Ram81, line 15: please add newline at the end of the file. Last line seems to have extra space.

Ram81 (changes requested):

Ram81:
@burnerlee sure, push the config file name change in #3447

burnerlee:
> @burnerlee sure, push the config file name change in #3447

yeah, did that change there.. please review.. I'll fix this PR in the meanwhile

Ram81:
I think you missed to push the commit

burnerlee:
> I think you missed to push the commit

https://github.com/Cloud-CV/EvalAI/pull/3447/commits/301c4c8232718e7381a85d9a1f993d94deeeab2b

Ram81:
Yes

burnerlee:
@Ram81 removed the staging prometheus file from here.. I'll add that file in #3447

monitoring/prometheus/prometheus_staging.yml
burnerlee: I'll take a while.. I have no experience in reading this type of file.. just giving it a read

burnerlee:
@Ram81 i'll add the prometheus_staging file in #3447 as soon i understand the config file....currently on it... then I think this PR needs no more change?

Ram81:
Yes, this PR is dev only setup for prometheus and grafana

Ram81 (approved):
LGTM

evalai/urls.py
```diff
@@ -92,6 +92,7 @@
         schema_view.with_ui("redoc", cache_timeout=0),
         name="schema-redoc",
     ),
+    url(r"^api/prometheus/", include("django_prometheus.urls")),
```
Ram81, line 95: Let's make this path `api/metrics`

monitoring/prometheus/prometheus.yml
```diff
@@ -2,7 +2,13 @@ global:
...
   external_labels:
     monitor: 'evalai-monitor'
+
 scrape_configs:
   - job_name: 'prometheus' 
     static_configs: 
       - targets: ['localhost:9090']
+
+  - job_name: 'django-prometheus'
+    metrics_path: '/api/prometheus/metrics'
```
Ram81, line 12: Change this path as well

Ram81 (changes requested):

evalai/urls.py
```diff
@@ -92,6 +92,7 @@
         schema_view.with_ui("redoc", cache_timeout=0),
         name="schema-redoc",
     ),
+    url(r"^api/", include("django_prometheus.urls")),
```
Ram81, line 95: @burnerlee I thought this would become `api/metrics`. Including prometheus urls at `api/` isn't a good idea. You can revert to `api/prometheus`

monitoring/prometheus/prometheus.yml
```diff
@@ -2,7 +2,13 @@ global:
...
   external_labels:
     monitor: 'evalai-monitor'
+
 scrape_configs:
   - job_name: 'prometheus' 
     static_configs: 
       - targets: ['localhost:9090']
+
+  - job_name: 'django-prometheus'
+    metrics_path: '/api/metrics'
```
Ram81, line 12: Update the path here too

Ram81 (changes requested):
Minor change, rest of it LGTM

evalai/urls.py
RishabhJain2018: @burnerlee @Ram81 I might be missing context here but what is the issue with using `api/metrics`

Ram81: @RishabhJain2018 I just wanted all prometheus under single root api path

RishabhJain2018: I see..okay. Sounds good!

docker/dev/docker.env
```diff
@@ -35,3 +35,5 @@ SERVICE_DNS=localhost
 
 GF_SECURITY_ADMIN_USER=admin    
 GF_SECURITY_ADMIN_PASSWORD=password
+
+prometheus_multiproc_dir=/code/monitoring/prometheus/django
```
Ram81, line 39: @burnerlee what is this for?

burnerlee: @Ram81 for aggregating the metrics of the 4 workers running for our django instance.. there were 2 options: 
1. to create a separate route for each worker and scrape metrics from each of them but it won't be really good.
2. The second was to aggregate metrics on the application level using a common registry for all the workers, which is maintained by django-prometheus by itself. For this it creates some temporary files in this directory and we have to set the environment variable for this. this env variable has to be set in lowercase only so we won't be able to make it uppercase.. however i have made a PR on this library to add support for upper case as well

Ram81: @burnerlee have you checked how much data it stores? We don't want disk to become a bottleneck for metrics

Ram81 (approved):
LGTM 👍

Ram81:
@RishabhJain2018 can you take a look.

RishabhJain2018 (approved):

PR #3449, opened 2021-06-03
Fix the JS error when loading all challenges on the challenge page

gautamjajoo:
**Fix:** Use `Object.keys()` to get the length of the object and store them. Use `setTimeout` for defering the code([Ref](https://blog.angular-university.io/angular-debugging/))

@Ram81 @Kajol-Kumari


Ram81:
@Kajol-Kumari can you test the changes locally once.

frontend_v2/src/app/components/publiclists/challengelist/challengelist.component.ts
```diff
@@ -226,15 +241,24 @@ export class ChallengelistComponent implements OnInit {
     this.routerPublic = this.router;
   }
 
+  /**
+   * After view is initialized.
+   */
+   ngAfterViewInit() {}
```
Kajol-Kumari, line 247: This won't be required here so can we please remove it?

frontend_v2/src/app/components/publiclists/challengelist/challengelist.component.ts
```diff
@@ -226,15 +241,24 @@ export class ChallengelistComponent implements OnInit {
...
+
+ 
   /**
    * Listener for page scroll event
    */
   @HostListener('window:scroll', [])
   onWindowScroll(): void {
+    setTimeout(() => {
```
Kajol-Kumari, line 255: We won't require to use `setTimeout` anywhere as the error was coming only because the variable `ongoingChallenges` was getting used in the template even before its initialising.

frontend_v2/src/app/components/publiclists/challengelist/challengelist.component.ts
```diff
@@ -377,14 +401,18 @@ export class ChallengelistComponent implements OnInit {
    */
   fetchChallengesFromApi(path, callback = null) {
     const SELF = this;
+    setTimeout(() => {
```
Kajol-Kumari, line 404: please remove it as it won't be required

Kajol-Kumari (changes requested):
@gautamjajoo The reason of getting the error is that the variable `ongoingChallenges` is getting used in the template before its initialisation. So, we can fix this by just initialising it at [here](https://github.com/Cloud-CV/EvalAI/blob/master/frontend_v2/src/app/components/publiclists/challengelist/challengelist.component.ts#L41) i.e `ongoingChallenges = []` so we won't require extra three variables to store the count separately.

Also, I can see similar kind of error in `participated` and `hosted` challenges page. Please fix them as well.

gautamjajoo:
@Kajol-Kumari Updated the PR

frontend_v2/src/app/components/publiclists/challengelist/challengelist.component.html
```diff
@@ -3,7 +3,7 @@
   <div *ngIf="routerPublic.url === allChallengesRoutePathCommon" [@.disabled]="true">
     <h4 class="fw-light">All challenges</h4>
     <mat-tab-group>
-      <mat-tab label="Ongoing ({{ ongoingChallenges.length }})">
+      <mat-tab label="Ongoing ({{ ongoingChallenges.length }})" id="ongoing-challenges">
```
Kajol-Kumari, line 6: can we please revert this change as its not required

Kajol-Kumari (approved):
LGTM  👍

frontend_v2/src/app/components/publiclists/challengelist/challengelist.component.html
gautamjajoo: Actually we are getting an error as the id [here](https://github.com/Cloud-CV/EvalAI/blob/39bd56239d2eef5a1b4f16ab0eb44be7b5fbdc07/frontend_v2/src/app/components/publiclists/challengelist/challengelist.component.ts#L234) is not assigned anywhere.

Ram81 (approved):
LGTM

PR #3452, opened 2021-06-03
Refactor code upload worker to spawn single pod for static code upload challenge

savish28:
Context: Static Code Upload Challenge

Deliverables:
- [x] Refactor code upload worker to spawn single pod for static code upload challenge.
- [x] Refactor submission worker to disregard static dataset code upload job submission message.
- [x] Add volume mount setup in job pods to mount test dataset
- [x] Add volume mount setup in job pods to mount scripts for submission.
- [x] Add pre-stop hook to submit the job output csv/json to EvalAI.
- [x] Export stdout, stderr logs from jobs to share with participants
- [x] Create monitor_submission.sh
- [x] Update make_submission.sh to create completed.txt/failed.txt
- [x] Create Sidecar container
- [x] Refactor pre-stop hook code to sidecar container. 

@KhalidRmb 
cc: @Ram81


scripts/workers/code_upload_submission_worker.py
```diff
@@ -119,6 +119,69 @@ def create_job_object(message, environment_image):
...
+
+    volume_mount_list = get_volume_mount_list("/dataset")
+    lifecycle_config = client.V1Lifecycle(
+        pre_stop=client.V1Handler(
+            _exec=client.V1ExecAction(
+                command=["/bin/sh", "-c", "sleep 10"]  # Submit File to evalai
+            )
+        )
+    )
+    # Configureate Pod agent container
```
KhalidRmb, line 149: @savish28 Please fix the spelling

scripts/workers/code_upload_submission_worker.py
```diff
@@ -119,6 +119,69 @@ def create_job_object(message, environment_image):
...
+    EVALAI_API_SERVER_ENV = client.V1EnvVar(
+        name="EVALAI_API_SERVER", value=EVALAI_API_SERVER
+    )
+    # Used by agent to create submission file by phase_pk and selecting dataset location
+    MESSAGE_BODY_ENV = client.V1EnvVar(name="BODY", value=json.dumps(message))
+    submission_pk = message["submission_pk"]
+    image = message["submitted_image_uri"]
+
+    volume_mount_list = get_volume_mount_list("/dataset")
+    lifecycle_config = client.V1Lifecycle(
```
Ram81, line 142: please change this to `life_cycle_config`

scripts/workers/code_upload_submission_worker.py
```diff
@@ -119,6 +119,69 @@ def create_job_object(message, environment_image):
...
+        pre_stop=client.V1Handler(
+            _exec=client.V1ExecAction(
+                command=["/bin/sh", "-c", "sleep 10"]  # Submit File to evalai
```
Ram81, line 145: @savish28 is this incomplete?

scripts/workers/code_upload_submission_worker.py
```diff
@@ -119,6 +119,69 @@ def create_job_object(message, environment_image):
...
+            )
+        )
+    )
+    # Configureate Pod agent container
+    agent_container = client.V1Container(
+        name="agent",
```
Ram81, line 151: Please rename the container to `submission_container` and name of the container will be `submission`

scripts/workers/code_upload_submission_worker.py
```diff
@@ -119,6 +119,69 @@ def create_job_object(message, environment_image):
...
+        image=image,
+        env=[PYTHONUNBUFFERED_ENV, EVALAI_API_SERVER_ENV, MESSAGE_BODY_ENV],
```
Ram81, line 153: environment variable to make submission API call will also be part of this list

scripts/workers/code_upload_submission_worker.py
```diff
@@ -119,6 +119,69 @@ def create_job_object(message, environment_image):
...
+        spec=client.V1PodSpec(
+            containers=[agent_container],
+            restart_policy="Never",
+            volumes=volume_list,
+        ),
+    )
+    # Create the specification of deployment
+    spec = client.V1JobSpec(backoff_limit=1, template=template)
+    # Instantiate the job object
+    job = client.V1Job(
```
Ram81, line 174: We can create a new method to get the `Job` object. Parameters of this method will be `submission_pk` with needed specs

scripts/workers/code_upload_submission_worker.py
```diff
@@ -163,8 +226,11 @@ def process_submission_callback(api_instance, body, challenge_phase, evalai):
     """
     try:
         logger.info("[x] Received submission message %s" % body)
-        environment_image = challenge_phase.get("environment_image")
-        job = create_job_object(body, environment_image)
+        if body.get("is_static_dataset_code_upload_submission"):
+            job = create_single_pod_job_object(body)
```
Ram81, line 230: Let's rename the method to `create_static_code_upload_submission_job_object`

Ram81 (changes requested):

scripts/workers/code_upload_submission_worker.py
savish28: @Ram81 Initially I thought of creating the pipeline first in this PR and creating the curl request in another PR, but now updating the changes in this PR.

scripts/workers/code_upload_submission_worker.py
```diff
@@ -67,6 +67,56 @@ def get_submission_meta_update_curl(submission_pk):
     return curl_request
 
 
+def get_static_code_upload_submission_file_curl(
+    challenge_pk, phase_pk, submission_pk, submission_path
+):
+    url = "{}/api/jobs/challenge/{}/challenge_phase/{}/submission/{}".format(
```
Ram81, line 73: @savish28 let's move this to a script `make_submission.sh` and in pre-stop hook just do `sh /path/to/make_submission.sh`

scripts/workers/code_upload_submission_worker.py
```diff
@@ -127,14 +171,76 @@ def create_job_object(message, environment_image):
...
+    # Configure submission pod container pre-stop hook to submit file.
+    submission_path = "/submission"
+    SUBMISSION_PATH_ENV = client.V1EnvVar(
+        name="SUBMISSION_PATH", value=submission_path
+    )
+    # TODO: Mount an empty folder at submission_path or create post-start hook
+    submission_curl_request = get_static_code_upload_submission_file_curl(
+        challenge_pk, phase_pk, submission_pk, submission_path
+    )
+    pre_stop_exec_command = "mkdir -p {submission_path}; for file_path in `ls {submission_path} | tail -n 1`; do {curl_request}; done;".format(
```
Ram81, line 207: @savish28 `submission` folder should be present when initializing the container

Ram81, line 207: participant can save multiple files we don't need to check if there is only a single file. You can remove the `tail` command. We expect the participant to save submission file as `submission.csv`. If the file is not present we will fail the submission

Ram81 (changes requested):

scripts/workers/code_upload_submission_worker.py
savish28: Yeah, I thought, in case the participant deletes the submission folder(by mistake or something) xD our pre-stop hook method might fail, so added "mkdir -p". I guess if we are checking only submission.json or submission.csv I would refactor it and directly check if the file exists.

scripts/workers/code_upload_submission_worker.py
```diff
@@ -67,6 +111,77 @@ def get_submission_meta_update_curl(submission_pk):
     return curl_request
 
 
+def get_static_code_upload_no_submission_file_curl(
+    challenge_pk, phase_pk, submission_pk
+):
+    url = "{}/api/jobs/challenge/{}/update_submission/".format(
```
Ram81, line 117: @savish28 why are we making curl request from the worker? This is not the approach we discussed. The curl requests will be written in `make_submission.sh` we will only pass environment variables to the container. The curl request will use environment variable. Code upload worker shouldn't substitute curl params and share fixed curl with the submission job.

scripts/workers/code_upload_submission_worker.py
```diff
@@ -67,6 +111,77 @@ def get_submission_meta_update_curl(submission_pk):
...
+    url = "{}/api/jobs/challenge/{}/challenge_phase/{}/submission/{}".format(
+        EVALAI_API_SERVER, challenge_pk, phase_pk, submission_pk
+    )
+    curl_base_request = "curl --location --request PATCH '{}' --header 'Authorization: Bearer {}'".format(
+        url, AUTH_TOKEN
+    )
+    curl_request = (
+        curl_base_request
+        + ' -F "submission_input_file=@{}/$file_path"'.format(submission_path)
+    )
```
Ram81, line 147: please move this curl to `make_submission.sh`

scripts/workers/code_upload_submission_worker.py
```diff
@@ -127,14 +236,104 @@ def create_job_object(message, environment_image):
...
+    submission_curl_request = get_static_code_upload_submission_file_curl(
+        challenge_pk, phase_pk, submission_pk, submission_path
+    )
+    submission_file_not_present_curl_request = (
+        get_static_code_upload_no_submission_file_curl(
+            challenge_pk, phase_pk, submission_pk
+        )
+    )
+    pre_stop_exec_command = (
+        'sh /scripts/make_submission.sh {} "{}" "{}"'.format(
```
Ram81, line 299: Here you need to execute `sh /scripts/make_submission.sh` and all required parameters needed for `make_submission.sh` should be environment variables

scripts/workers/code_upload_worker_utils/make_submission.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+echo "File Submission Started."
+# Submission_folder $1
+# Sumbission curl request $2
+# File Not found curl request $3
+if [ -f "$1/submission.csv" ]
+then
+    file_path="submission.csv"
```
Ram81, line 8: Please add actual curl requests here.

scripts/workers/code_upload_worker_utils/make_submission.sh
```diff
@@ -0,0 +1,18 @@
...
+    eval $2
+elif [ -f "$1/submission.json" ]
+then
+  file_path="submission.json"
+  eval $2
+else
+    echo "Submission file not found."
+    eval $3
+fi
+echo "File Submission Ended"
```
Ram81, line 18: Add newline

Ram81 (changes requested):

scripts/workers/code_upload_submission_worker.py
```diff
@@ -127,14 +199,107 @@ def create_job_object(message, environment_image):
...
+    AUTH_TOKEN_ENV = client.V1EnvVar(name="AUTH_TOKEN", value=AUTH_TOKEN)
+    EVALAI_API_SERVER_ENV = client.V1EnvVar(
+        name="EVALAI_API_SERVER", value=EVALAI_API_SERVER
+    )
+    image = message["submitted_image_uri"]
+    # Get init container
+    init_container = get_init_container(submission_pk)
+    # Get dataset volume and volume mounts
+    volume_mount_list = get_volume_mount_list("/dataset", True)
+    volume_list = get_volume_list()
```
Ram81, line 237: @savish28 you can refactor this method to use `get_config_map_volume_object` for dataset volume as well

scripts/workers/code_upload_submission_worker.py
```diff
@@ -127,14 +199,107 @@ def create_job_object(message, environment_image):
...
+    # Create and Mount Script Volume
+    script_config_map_name = "script-config"
```
Ram81, line 239: rename it to `evalai-scripts-cm`

scripts/workers/code_upload_submission_worker.py
```diff
@@ -127,14 +199,107 @@ def create_job_object(message, environment_image):
...
+    create_script_config_map(script_config_map_name)
+    script_volume_name = "scripts-dir"
```
Ram81, line 241: rename to `evalai-scripts`

scripts/workers/code_upload_submission_worker.py
```diff
@@ -127,14 +199,107 @@ def create_job_object(message, environment_image):
...
+    script_volume = get_config_map_volume_object(
+        script_config_map_name, script_volume_name
+    )
+    volume_list.append(script_volume)
+    script_volume_mount = get_volume_mount_object(
+        "/scripts", script_volume_name, True
```
Ram81, line 247: let's change the mount path to `evalai_scripts`

scripts/workers/code_upload_worker_utils/make_submission.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh
+echo "File Submission Started."
+submission_url="$EVALAI_API_SERVER/api/jobs/challenge/$CHALLENGE_PK/challenge_phase/$PHASE_PK/submission/$SUBMISSION_PK"
+submission_curl_base_request="curl --location --request PATCH '$submission_url' --header 'Authorization: Bearer $AUTH_TOKEN'"
+if [ -f "$SUBMISSION_PATH/submission.csv" ]
```
Ram81, line 5: We need to take this extension as part of challenge config. For now we can use this script. Once we test everything we can switch to accepting submisison extension from challenge config

scripts/workers/code_upload_worker_utils/make_submission.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh
+echo "File Submission Started."
```
Ram81, line 2: Change this to `Submitting file to EvalAI...`

scripts/workers/code_upload_worker_utils/make_submission.sh
```diff
@@ -0,0 +1,28 @@
...
+    submission_data="{\
+      \"challenge_phase\": $PHASE_PK,\
+      \"submission\": $SUBMISSION_PK,\
+      \"stdout\": \"\",\
+      \"stderr\": \"submission.json/submission.csv not found.\",\
+      \"submission_status\": \"failed\",\
+      \"result\": \"[]\",\
+      \"metadata\": \"\"\
+    }"
+    curl_request="curl --location --request PUT '$url' -H 'Content-Type: application/json' --header 'Authorization: Bearer $AUTH_TOKEN' -d '$submission_data'"
```
Ram81, line 25: Add a message here `File submission failed` and exit the script after API call.

scripts/workers/code_upload_worker_utils/make_submission.sh
```diff
@@ -0,0 +1,28 @@
...
+fi
+eval $curl_request
+echo "\nFile Submission Ended"
```
Ram81, line 28: Add a message here `File submitted successfully`

Ram81 (changes requested):

scripts/workers/code_upload_worker_utils/make_submission.sh
Ram81, line 25: This submission also needs to be deleted from the queue after it is marked failed. Can you please make that change as well.

Ram81 (changes requested):

scripts/workers/code_upload_worker_utils/make_submission.sh
savish28: @Ram81, wouldn't the code upload worker do that? I saw there was a condition that deletes the method when the status is changed. 
Current flow: status change -> worker reads the message again -> check status and then deletes.
Is this suggested flow : change status and delete message directly?
Is the idea we also pass receipt_handle and then delete from message by sending it to evalai.

scripts/workers/code_upload_submission_worker.py
savish28: Done.

scripts/workers/code_upload_submission_worker.py
savish28: Done.

scripts/workers/code_upload_worker_utils/make_submission.sh
savish28: Done.

scripts/workers/code_upload_worker_utils/make_submission.sh
savish28: Done.

scripts/workers/code_upload_worker_utils/make_submission.sh
savish28: Done

scripts/workers/code_upload_submission_worker.py
savish28: Done

scripts/workers/code_upload_worker_utils/make_submission.sh
Ram81: Yes, the code upload worker will delete it we don't need to do it explicitly

scripts/workers/code_upload_submission_worker.py
```diff
@@ -57,6 +64,43 @@ def get_volume_list():
...
+    metadata = client.V1ObjectMeta(
+        name=config_map_name,
+    )
+    config_data = {}
+    for file_path in file_paths:
+        file_name = os.path.basename(file_path)
+        file_content = open(file_path, "r").read()
+        config_data[file_name] = file_content
+    # Instantiate the config_map object
+    config_map = client.V1ConfigMap(
```
KhalidRmb, line 90: @savish28 Please implement a try catch block here. Always a good practice when you're communicating with something external

scripts/workers/code_upload_submission_worker.py
```diff
@@ -57,6 +64,43 @@ def get_volume_list():
...
+        api_version="v1", kind="ConfigMap", data=config_data, metadata=metadata
+    )
+    return config_map
+
+
+def create_script_config_map(config_map_name):
+    submission_script_file_path = "code_upload_worker_utils/make_submission.sh"
+    script_config_map = create_config_map_object(
```
KhalidRmb, line 98: @savish28 Please add a try except block + logging message here. Isn't this communicating with something external?

savish28: @KhalidRmb this is creating the config map object locally and not connecting to external service at this stage, It connects with eks at a later stage where the actual job is created.

savish28: We can take the example of V1Volume as well, all these objects Volume, ConfigMap, VolumeMounts are created as-is, although they are all in a single try-except block at the creation stage where it connects with external client.

scripts/workers/code_upload_submission_worker.py
```diff
@@ -67,6 +115,40 @@ def get_submission_meta_update_curl(submission_pk):
...
+
+
+def get_init_container(submission_pk):
+    curl_request = get_submission_meta_update_curl(submission_pk)
+    # Configure init container
+    init_container = client.V1Container(
+        name="init-container",
+        image="busybox",
+        command="/bin/sh",
+        args=["-c", curl_request],
```
Ram81, line 147: @savish28 this would need to be rebased. See the changes in my recent [PR](https://github.com/Cloud-CV/EvalAI/pull/3497). `busybox` doesn't have `curl` so we have to use a ubuntu image and install curl explicitly. We will merge my PR today, you can rebase after that.

scripts/workers/code_upload_submission_worker.py
```diff
@@ -127,14 +203,121 @@ def create_job_object(message, environment_image):
...
         ),
-        spec=spec,
+        volume_mounts=volume_mount_list,
+    )
+    # Configure Pod sidecar container
+    sidecar_container = client.V1Container(
+        name="sidecar-container",
+        image="busybox",
+        command="/bin/sh",
+        args=["-c", "sh /evalai_scripts/monitor_submission.sh"],
```
Ram81, line 294: @savish28 this won't work in a `busybox` container. We have to use the ubuntu image and install curl explicitly. See my PR for reference. Also, the args might also not work. I had to pass the args as part of the command.

scripts/workers/code_upload_worker_utils/make_submission.sh
```diff
@@ -0,0 +1,19 @@
...
+then
+    file_path="submission.json"
+    curl_request="$submission_curl_base_request -F \"submission_input_file=@$SUBMISSION_PATH/$file_path\""
+else
+    exit 0
+fi
+echo "Submitting file to EvalAI..."
+eval $curl_request
+echo "\nFile submitted successfully"
+touch "$SUBMISSION_PATH/completed.txt"
```
Ram81, line 18: @savish28 instead of this you can use `echo $(date) > "$SUBMISSION_PATH/completed.txt"`. We don't need to create a file explicitly

scripts/workers/code_upload_worker_utils/monitor_submission.sh
```diff
@@ -0,0 +1,38 @@
...
+    echo "Checking for submission file after $time_elapsed secs.."
+    sh /evalai_scripts/make_submission.sh
+    if [ -f "$SUBMISSION_PATH/completed.txt" ]
+    then
+        echo "Submission Complete, Exiting."
+        exit 0
+    fi
+    if [ $time_elapsed -lt $SUBMISSION_TIME_LIMIT -a $(( $time_elapsed + $SUBMISSION_TIME_DELTA )) -gt $SUBMISSION_TIME_LIMIT ]
+    then
+        next_checkpoint=$SUBMISSION_TIME_LIMIT
```
Ram81, line 15: @savish28 please change the variable name to something like `time_delta` or `wait_interval`

scripts/workers/code_upload_worker_utils/make_submission.sh
```diff
@@ -0,0 +1,19 @@
...
+if [ -f "$SUBMISSION_PATH/submission.csv" ]
+then
+    file_path="submission.csv"
+    curl_request="$submission_curl_base_request -F \"submission_input_file=@$SUBMISSION_PATH/$file_path\""
+elif [ -f "$SUBMISSION_PATH/submission.json" ]
+then
+    file_path="submission.json"
+    curl_request="$submission_curl_base_request -F \"submission_input_file=@$SUBMISSION_PATH/$file_path\""
+else
+    exit 0
```
Ram81, line 13: Add a log statement here `Submission not complete`

scripts/workers/code_upload_worker_utils/monitor_submission.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/sh
+echo "Monitoring submission..."
+time_elapsed=0
+while [ $time_elapsed -le $SUBMISSION_TIME_LIMIT ]
+do
+    echo "Checking for submission file after $time_elapsed secs.."
+    sh /evalai_scripts/make_submission.sh
+    if [ -f "$SUBMISSION_PATH/completed.txt" ]
```
Ram81, line 8: Instead of this we could've just checked for `submission.csv/json` file in the monitor script. We woudn't need a `completed.txt` in that case. Any reason you preferred this setup?

scripts/workers/code_upload_worker_utils/monitor_submission.sh
```diff
@@ -0,0 +1,38 @@
...
+    \"submission\": $SUBMISSION_PK,\
+    \"stdout\": \"\",\
+    \"stderr\": \"submission.json/submission.csv not found.\",\
+    \"submission_status\": \"failed\",\
+    \"result\": \"[]\",\
+    \"metadata\": \"\"\
+}"
+curl_request="curl --location --request PUT '$url' -H 'Content-Type: application/json' --header 'Authorization: Bearer $AUTH_TOKEN' -d '$submission_data'"
+eval $curl_request
+echo "\nFile submission failed."
```
Ram81, line 37: This error is misleading right? This block is executed when submissions times out correct? If that is the case please make `stderr: "Execution time limit exceeded"`. 

We should handle error messages correctly here. If the submission containers exits before creating `submission.csv/json` show `submission.csv not found` otherwise if container is still running after n hours we should `Execution time limit exceeded`.

Ram81 (changes requested):

scripts/workers/code_upload_worker_utils/monitor_submission.sh
savish28: @Ram81 Yes, because I thought monitor_submission.sh should be independent of particulars of what file is submitted. `make_submission.sh` would take care of the particulars. Even if we modify that we want these types of files or maybe in future we want to submit multiple files all that logic would go into `make_submission.sh` and `monitor_submision.sh` would only check for `completed.txt`. In this way we don't have to repeat our logic again and again.

scripts/workers/code_upload_worker_utils/monitor_submission.sh
savish28: @Ram81 do you mean changing `SUBMISSION_TIME_DELTA` to `time_delta`? `next_checkpoint` would be the next time at which we would be going to check.

scripts/workers/code_upload_worker_utils/monitor_submission.sh
savish28: Yeah, we have a plan to check the health of submission docker from sidecar in near future, would add this error there. Changing this error to TLE error as suggested.

scripts/workers/code_upload_submission_worker.py
savish28: Okay, updated.

scripts/workers/code_upload_worker_utils/make_submission.sh
savish28: Updated

scripts/workers/code_upload_worker_utils/monitor_submission.sh
savish28: Updated.

scripts/workers/code_upload_worker_utils/make_submission.sh
savish28: Updated.

scripts/workers/code_upload_submission_worker.py
savish28: @Ram81, Sure would do the required changes as soon as the respective PR is merged.

scripts/workers/code_upload_worker_utils/monitor_submission.sh
Ram81: I meant change the name of `next_checkpoint` to something relevant to next time interval or similar

scripts/workers/code_upload_worker_utils/monitor_submission.sh
Ram81: Makes sense.

scripts/workers/code_upload_worker_utils/monitor_submission.sh
savish28: Updated.

scripts/workers/code_upload_submission_worker.py
```diff
@@ -368,7 +557,15 @@ def main():
     while True:
         message = evalai.get_message_from_sqs_queue()
         message_body = message.get("body")
+        message_body["submission_time_limit"] = challenge.submission_time_limit
```
Ram81, line 560: Why are we passing `submission_time_limit` in sqs message? This will be a fixed value for each challenge. Let's pass this as task_definition param to the worker. Let's not add too many details in sqs message

Ram81 (changes requested):

scripts/workers/code_upload_submission_worker.py
savish28: @Ram81, Since we are getting challenge object already in code upload worker, why is there a need to pass it from task_defn_worker? 
We could just set an env variable one time at the starting of the code upload worker when we are fetching the challenge details.
OR 
We could have a variable `submission_time_limit` passed explicitly to the function.

The submission_time_limit is not in the actual sqs message though it is in our local message_body variable but yeah we have that separate for good abstraction. The disadvantage is when we have some more info to pass we eventually have to pass it in some dict.
What do you suggest how we proceed?

Ram81: I see, let's create a new dict `submission_meta` and populate that in the same message body object for now. Adding new param to existing functions will need more changes.

savish28: Updated these changes.

scripts/workers/code_upload_submission_worker.py
```diff
@@ -56,6 +63,47 @@ def get_volume_list():
...
+        config_data[file_name] = file_content
+    # Instantiate the config_map object
+    config_map = client.V1ConfigMap(
+        api_version="v1", kind="ConfigMap", data=config_data, metadata=metadata
+    )
+    return config_map
+
+
+def create_script_config_map(config_map_name):
+    submission_script_file_path = "code_upload_worker_utils/make_submission.sh"
```
Ram81, line 96: This should be full path inside the docker container for it to work when reading the file. This should be `/code/scripts/workers/code_upload_worker_utils/make_submission.sh`

scripts/workers/code_upload_submission_worker.py
```diff
@@ -56,6 +63,47 @@ def get_volume_list():
...
+    monitor_submission_script_path = (
+        "code_upload_worker_utils/monitor_submission.sh"
```
Ram81, line 98: same here

scripts/workers/code_upload_worker_utils/make_submission.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
+submission_url="$EVALAI_API_SERVER/api/jobs/challenge/$CHALLENGE_PK/challenge_phase/$PHASE_PK/submission/$SUBMISSION_PK"
+submission_curl_base_request="curl --location --request PATCH '$submission_url' --header 'Authorization: Bearer $AUTH_TOKEN'"
```
Ram81, line 3: @savish28 have you verified these curl requests locally once?

tests/unit/challenges/test_views.py
```diff
@@ -1843,6 +1860,8 @@ def test_get_challenge_when_mode_is_host(self):
                     "team_url": self.challenge.creator.team_url,
                 },
                 "published": self.challenge.published,
+                "submission_time_limit": self.challenge.submission_time_limit,
+                "submission_time_limit": self.challenge.submission_time_limit,
```
Ram81, line 1864: Why is this repeated twice here?

Ram81 (changes requested):

scripts/workers/code_upload_worker_utils/make_submission.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
+submission_url="$EVALAI_API_SERVER/api/jobs/challenge/$CHALLENGE_PK/challenge_phase/$PHASE_PK/submission/$SUBMISSION_PK"
```
Ram81, line 2: @savish28 where are the respective API changes to update the `submission_input_file` and republishing the submission message to the queue? For this API there is not PATCH update available

Ram81 (changes requested):

tests/unit/challenges/test_views.py
savish28: Fixed

scripts/workers/code_upload_worker_utils/make_submission.sh
savish28: @Ram81 yes.

scripts/workers/code_upload_worker_utils/make_submission.sh
savish28: @Ram81 `change_submission_data_and_visibility` in jobs views.py is related to this, we have made changes to send message in this particular view in a previous PR that is merged.

scripts/workers/code_upload_submission_worker.py
savish28: Done

scripts/workers/code_upload_submission_worker.py
savish28: Updated

Ram81 (approved):
LGTM

RishabhJain2018 (approved):
LGTM :)

PR #3479, opened 2021-06-11
Add submission verification by hosts

Ram81:
### Description

This PR adds -

- [x] Option to mark submission `verified` from All submissions tab.
- [x] Show a badge on the leaderboard entry of the submission.

Screen Shot 2021-06-11 at 10 32 49 PM


apps/jobs/models.py
```diff
@@ -129,6 +129,7 @@ class Submission(TimeStampedModel):
     ignore_submission = models.BooleanField(default=False)
     # Store the values of meta attributes for the submission here.
     submission_metadata = JSONField(blank=True, null=True)
+    is_verified = models.BooleanField(default=False)
```
RishabhJain2018, line 132: `is_verified` -> `is_verified_by_host`. Also, please add comments on why this field is there.

apps/jobs/urls.py
```diff
@@ -110,6 +110,12 @@
         views.update_submission_started_at,
         name="update_submission_started_at",
     ),
+    url(
+        r"^challenge/(?P<challenge_pk>[0-9]+)/"
```
RishabhJain2018, line 114: `challenge/(?P<challenge_pk>[0-9]+)/"` -> `challenges/(?P<challenge_pk>[0-9]+)/"`

apps/jobs/views.py
```diff
@@ -2775,3 +2775,44 @@ def update_submission_started_at(request, submission_pk):
         return Response(response_data, status=status.HTTP_200_OK)
     else:
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
+
+
+@api_view(["PATCH"])
+@throttle_classes([UserRateThrottle])
+@permission_classes((permissions.IsAuthenticated, HasVerifiedEmail))
+@authentication_classes((JWTAuthentication, ExpiringTokenAuthentication))
+def update_submission_data(request, challenge_pk, submission_pk):
```
RishabhJain2018, line 2784: Why do we need a new API? Why can't we simply reuse the `update_submission` API?

RishabhJain2018 (changes requested):

apps/jobs/views.py
Ram81: `update_submission` expects `stdout`, `stderr`, and other submission metadata and updates all those fields. If we just pass status of submission it will clear all other details of a submission. And `PATCH` request is specifically set up for code upload worker it seems as it only allows changing status when submisison is `running`. We can make meta update optional in `PUT` for updating submisison status.

apps/jobs/urls.py
RishabhJain2018: @Ram81 I think you missed this comment.

apps/jobs/urls.py
```diff
@@ -110,6 +110,12 @@
...
+        r"submission/(?P<submission_pk>[0-9]+)/update_submission_meta/$",
```
RishabhJain2018, line 115: `r"submission/(?P<submission_pk>[0-9]+)/update_submission_meta/$"` -> `r"submissions/(?P<submission_pk>[0-9]+)/update_submission_meta/$"`

apps/jobs/views.py
```diff
@@ -2775,3 +2775,45 @@ def update_submission_started_at(request, submission_pk):
...
+@permission_classes((permissions.IsAuthenticated, HasVerifiedEmail))
+@authentication_classes((JWTAuthentication, ExpiringTokenAuthentication))
+def update_submission_meta(request, challenge_pk, submission_pk):
+    """
+    Common API Endpoint for updating the submission meta data for hosts and participants.
+    """
+
+    if is_user_a_host_of_challenge(request.user, challenge_pk):
+        try:
+            submission = Submission.objects.get(
```
RishabhJain2018, line 2791: Can we please use `get_submission_object` function from utils?

apps/jobs/views.py
```diff
@@ -2775,3 +2775,45 @@ def update_submission_started_at(request, submission_pk):
...
+        if serializer.is_valid():
+            serializer.save()
+            response_data = serializer.data
+            return Response(response_data, status=status.HTTP_200_OK)
+        else:
+            return Response(
+                serializer.errors, status=status.HTTP_400_BAD_REQUEST
+            )
+    else:
+        response_data = {
```
RishabhJain2018, line 2816: I think we also discussed adding the `submission_cancel` feature in this API only, no?

RishabhJain2018 (changes requested):

apps/jobs/views.py
Ram81: Let's make that change in a separate PR. I didn't want to merge them in a single one

RishabhJain2018 (approved):

PR #3493, opened 2021-06-20
Monitor: static code upload submission time limit

savish28:
Context: Static Code upload challenge.

Depends on: #3452

Deliverables:
- [x] Create monitor_submission.sh
- [x] Update make_submission.sh to create completed.txt/failed.txt
- [x] Create Sidecar container

This PR assumes that for all submissions, we have a default time limit of 1 day since no such field was available in the challenge phase model as of now. [Edit] - Added it in this PR.
We can have a separate PR for adding the timelimit field and integrating it with this. 

To review changes relevant to this PR commit "Monitor: static Code upload challenge" may be referred. All previous commits are from depending PR.

@KhalidRmb 
cc @Ram81


scripts/workers/code_upload_worker_utils/monitor_submission.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/sh
+echo "Monitoring submission..."
+time_elapsed=0
+time_delta=3600
```
KhalidRmb, line 4: @savish28  Can we have this as an environment variable or part of the configMap? Avg runtime may vary depending on the type of challenge 
What do you think @Ram81?

savish28: @KhalidRmb, yeah thought of that looked a good idea but the thing is it would make sense that if we keep it in challenge model somewhere.

KhalidRmb: Right.

savish28: Time delta is just for our time saving of the job running and doesn't directly deal with challenge i think. For eg if the timelimit is less that timedelta than it won't cost us much.

So that was the idea. But if you suggest i can put it in the challenge config.

KhalidRmb: It matters if say the submissions are processed in much less than an hour or just over a whole number of hours, and we spend a not so insignificant amount of time idle.

Submission time limit is static right? Isn't that set by default to 1 day?

Even if you put time_delta in the config, it should be taken from the challenge model only right?...you could set it as an environment variable like the submission time limit as well.

savish28: Yeah, in this case where submission are processed early on than time_delta and time limit is huge.
It makes sense to pass it as env variable. I would do that change.

I was thinking for eg If i am the host of the challenge i would not care at what interval the submission are monitored i guess. 😅. 

Would add as an env variable and add a field in challenge fot both time_delta and time_limit

KhalidRmb: The challenge host need not set it per se. You can leave default values as the ones right now, but the point is that it's configurable. 

I agree it's not absolutely necessary right now, just a neat way to do things if we decide for whatever reason in the future to have different kind of challenges have different time limits.

Thoughts @Ram81

savish28: @KhalidRmb added time_delta it in ENV, would add both fields in challenge model as well.

KhalidRmb: Env change is fine. Wait until Ram confirms for the challenge model change. Thanks!

savish28: In today's scrum ram and Rishabh agreed on adding submission_time_limit, so let's add that now and would add time_delta later. Cool?

KhalidRmb: Cool

savish28:
Closing this PR, merging this with #3452 and shifting the responsibility of pre-stop hook to sidecar container

PR #3497, opened 2021-06-24
Update Kubernetes DaemonSet API for cluster version > 1.16

Ram81:
### Description

- [x] Replace `ExtensionsV1beta1Api` with `AppsV1Api` after cluster update to version 1.16. Refer [this](https://stackoverflow.com/questions/59943526/no-matches-for-kind-daemonset-in-version-extensions-v1beta1)
- [x] Update `nvidia-device-plugin.yaml` to run as `apps/v1` API as the official version isn't supported for cluster version > 1.16
- [x] Replace init container image with ubuntu to execute curl successfully


scripts/workers/code_upload_submission_worker.py
```diff
@@ -95,8 +94,11 @@ def create_job_object(message, environment_image):
     init_container = client.V1Container(
         name="init-container",
         image="busybox",
```
savish28, line 96: Should `ubuntu:latest` come here instead of `busybox`?

Ram81: Fixed it. Thanks for pointing it out.

RishabhJain2018 (approved):

PR #3503, opened 2021-06-28
Reverse Proxy setup for prometheus and grafana

burnerlee:
Add configs to expose prometheus and grafana through endpoints configured by the reverse proxy. Prometheus is available on `ipaddr/prometheus` and Grafana on `ipaddr/grafana`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -66,6 +66,7 @@ services:
     - /code/node_modules
 
   prometheus:
+    hostname: prometheus
     image: prom/prometheus:latest
     user: "1000"
     volumes:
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -75,15 +76,20 @@ services:
...
       - '9090:9090'
   
   grafana:
+    hostname: grafana
     image: grafana/grafana:latest
     user: "1000"
     env_file:
       - docker/dev/docker.env
+    environment:
+      GF_SERVER_ROOT_URL: http://localhost:3000/grafana
```
Ram81, line 91: Fetch these values from `.env` file

docker-compose.yml
```diff
@@ -75,15 +76,20 @@ services:
...
+      GF_SERVER_SERVE_FROM_SUB_PATH: 'true'
```
Ram81, line 92: same here

Ram81 (changes requested):

burnerlee:
Closing this PR as currently we are planning to run Grafana & Prometheus on separate instance than Nodejs, so the its nginx config won't work

PR #3506, opened 2021-06-28
Bug Fix: static code upload worker challenge integrations

savish28:
Context: Static code upload challenge.

These changes are based on the testing of static code upload worker.

Deliverables:
- [x] Challenge object dict bug
- [x] messgae_body None 
- [x] Config Map creation
- [x] Change API endpoint from `change_submission_data_and_visibility` to `update_submission` for evaluation stage

@KhalidRmb 
cc: @Ram81

### DIFF
--- a/apps/jobs/views.py
+++ b/apps/jobs/views.py
@@ -1249,6 +1249,7 @@ def update_submission(request, challenge_pk):
     if request.method == "PATCH":
         submission_pk = request.data.get("submission")
         submission = get_submission_model(submission_pk)
+        # Submission of submission_input_file for static_dataset_code_upload challenge from submission to evaluation stage
         if (
             request.FILES.get("submission_input_file")
             and submission.challenge_phase.challenge.is_static_dataset_code_upload
--- a/apps/jobs/views.py
+++ b/apps/jobs/views.py
@@ -486,18 +486,6 @@ def change_submission_data_and_visibility(
     if serializer.is_valid():
         serializer.save()
         response_data = serializer.data
-        if (
-            request.FILES.get("submission_input_file")
-            and challenge.is_static_dataset_code_upload
-        ):
-            message = {
-                "challenge_pk": challenge_pk,
-                "phase_pk": challenge_phase_pk,
-                "submission_pk": submission_pk,
-                "is_static_dataset_code_upload_submission": False,
-            }
-            # publish message in the queue
-            publish_submission_message(message)
         return Response(response_data, status=status.HTTP_200_OK)
     else:
         return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
@@ -1260,9 +1248,37 @@ def update_submission(request, challenge_pk):
 
     if request.method == "PATCH":
         submission_pk = request.data.get("submission")
+        submission = get_submission_model(submission_pk)
+        if (
+            request.FILES.get("submission_input_file")
+            and submission.challenge_phase.challenge.is_static_dataset_code_upload
+        ):
+            serializer = SubmissionSerializer(
+                submission,
+                data=request.data,
+                context={
+                    "request": request,
+                },
+                partial=True,
+            )
+            if serializer.is_valid():
+                serializer.save()
+                message = {
+                    "challenge_pk": challenge_pk,
+                    "phase_pk": submission.challenge_phase.pk,
+                    "submission_pk": submission_pk,
+                    "is_static_dataset_code_upload_submission": False,
+                }
+                # publish message in the queue
+                publish_submission_message(message)
+                response_data = serializer.data
+                return Response(response_data, status=status.HTTP_200_OK)
+            else:
+                return Response(
+                    serializer.errors, status=status.HTTP_400_BAD_REQUEST
+                )
         submission_status = request.data.get("submission_status", "").lower()
         job_name = request.data.get("job_name", "").lower()
-        submission = get_submission_model(submission_pk)
         jobs = submission.job_name
         if job_name:
             jobs.append(job_name)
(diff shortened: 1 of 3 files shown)


scripts/workers/code_upload_submission_worker.py
```diff
@@ -557,31 +567,36 @@ def main():
         cluster_name, cluster_endpoint, challenge, evalai
     )
     install_gpu_drivers(api_instance)
+    api_instance = get_api_object(
```
Ram81, line 570: Why are we overriding `api_instance` here? It is already being used at multiple places

apps/jobs/views.py
```diff
@@ -1260,9 +1248,37 @@ def update_submission(request, challenge_pk):
 
     if request.method == "PATCH":
         submission_pk = request.data.get("submission")
+        submission = get_submission_model(submission_pk)
+        if (
```
Ram81, line 1252: @savish28 can you add a one line comment to describe what we are doing here

apps/jobs/views.py
```diff
@@ -1260,9 +1248,37 @@ def update_submission(request, challenge_pk):
...
+            request.FILES.get("submission_input_file")
```
Ram81, line 1253: Also, it would be good if you add a `if else` here. Update the `submission_input_file` if the challenge is static code upload challenge. Else if the challenge is code upload challenge do the other update. Check for the `request.FILES.get("submission_input_file")` should be inside if.

scripts/workers/code_upload_worker_utils/make_submission.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
-submission_url="$EVALAI_API_SERVER/api/jobs/challenge/$CHALLENGE_PK/challenge_phase/$PHASE_PK/submission/$SUBMISSION_PK"
+submission_url="$EVALAI_API_SERVER/api/jobs/challenge/$CHALLENGE_PK/update_submission/"
```
Ram81, line 2: Where is the submission pk for this API call?

scripts/workers/code_upload_worker_utils/make_submission.sh
```diff
@@ -1,5 +1,5 @@
...
 submission_curl_base_request="curl --location --request PATCH '$submission_url' --header 'Authorization: Bearer $AUTH_TOKEN'"
```
Ram81, line 3: The above API expects submission pk to be in request data. Please add that change

Ram81 (changes requested):

scripts/workers/code_upload_submission_worker.py
savish28: @Ram81, it was being overridden in the While before and the parameters were not changing, so just moved the code outside the While. 
Also, it is assigned firstly by `get_api_client` and then calls `install_gpu_drivers` and then assigned by `get_api_object` so followed the same thing. Just took api_instance out of the while loop.
We could change the name of the firstly used api_instance.

apps/jobs/views.py
savish28: the current flow is
if File_submitted + static_code_upload_challenge: Do our thing and return
"Prev flow"

Are you suggesting
if static_code_upload_challenge:
    if File_submitted: Do our thing
else:
     "Prev flow"

The reason for not following the second design is that we are using the Prev flow as well for the static_code_upload_challenge.

Ram81: I see, then above changes are fine

scripts/workers/code_upload_worker_utils/make_submission.sh
savish28: Updated.

scripts/workers/code_upload_worker_utils/make_submission.sh
savish28: @Ram81, Sure,
Added it in the code, also tested it locally by creating a challenge and submitting using this script, also tested the formation of new message.

apps/jobs/views.py
```diff
@@ -1260,9 +1248,38 @@ def update_submission(request, challenge_pk):
 
     if request.method == "PATCH":
         submission_pk = request.data.get("submission")
+        submission = get_submission_model(submission_pk)
+        # Submission of submission_input_file for static_dataset_code_upload challenge from submission to evaluation stage
```
Ram81, line 1252: Change to `Update submission_input_file for is_static_dataset_code_upload submission evaluation`

scripts/workers/code_upload_worker_utils/make_submission.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh
-submission_url="$EVALAI_API_SERVER/api/jobs/challenge/$CHALLENGE_PK/challenge_phase/$PHASE_PK/submission/$SUBMISSION_PK"
-submission_curl_base_request="curl --location --request PATCH '$submission_url' --header 'Authorization: Bearer $AUTH_TOKEN'"
+submission_url="$EVALAI_API_SERVER/api/jobs/challenge/$CHALLENGE_PK/update_submission/"
+submission_curl_base_request="curl --location --request PATCH '$submission_url' --header 'Authorization: Bearer $AUTH_TOKEN' -F 'submission=$SUBMISSION_PK'"
```
Ram81, line 3: @savish28 shouldn't this `-F 'submission=$SUBMISSION_PK'` be part of request body?  Have you tested by making a API call locally?

savish28: @Ram81 It is part of the request body itself, the F stands for Form Data. Yes, I have tested these changes locally and the file reflected on the submission model. Verified from django admin. We are eventually using reques.data and it contains submission and the submission_input_file

apps/jobs/views.py
savish28: Updated

Ram81 (approved):
LGTm 👍

RishabhJain2018 (approved):

PR #3507, opened 2021-06-29
Integrate alert manager to monitoring setup

burnerlee:
Add alert manager to monitoring setup. Define rules to trigger alerts when Prometheus scraping targets are down or API responses with 5XX cross the threshold. Add slack channel as the endpoint for issuing alerts


docker/prod/nodejs/nginx_staging.conf
```diff
@@ -14,6 +14,10 @@ upstream push_gateway {
   server push_gateway:9091 fail_timeout=0;
 }
 
+upstream alertmanager {
```
Ram81, line 17: @burnerlee Why do we need alert manager to be on the same instance?

docker/prod/nodejs/nginx_staging.conf
```diff
@@ -86,6 +90,13 @@ server {
     proxy_pass http://push_gateway;
   }
 
+  location /alerts {
```
Ram81, line 93: If its not needed on the same instance please remove the nginx config changes

docker/prod/nodejs_v2/nginx_staging.conf
```diff
@@ -14,6 +14,10 @@ upstream push_gateway {
   server push_gateway:9091 fail_timeout=0;
 }
 
+upstream alertmanager {
+  server alerts:9093 fail_timeout=0;
```
Ram81, line 18: same here

monitoring/prometheus/alertmanager_staging.yml
```diff
@@ -0,0 +1,37 @@
+global:
+  resolve_timeout: 10s
+  slack_api_url: 'https://hooks.slack.com/services/x/x/x'
```
Ram81, line 3: No need to set up a separate alertmanager config if both of them will have slack webhook url, channel set to dummy value. We can use a single `.yml` file and push it to s3 with populated values

scripts/deployment/deploy.sh
```diff
@@ -47,7 +47,7 @@ case $opt in
 					aws s3 cp s3://cloudcv-secrets/evalai/${env}/docker_${env}.env ./docker/prod/docker_${env}.env
 					docker-compose -f docker-compose-${env}.yml rm -s -v -f
 					docker-compose -f docker-compose-${env}.yml pull
-					docker-compose -f docker-compose-${env}.yml up -d --force-recreate --remove-orphans statsd-exporter django nodejs nodejs_v2 celery node_exporter push_gateway
+					docker-compose -f docker-compose-${env}.yml up -d --force-recreate --remove-orphans statsd-exporter django nodejs nodejs_v2 celery node_exporter push_gateway alertmanager
```
Ram81, line 50: Why do you need alert manager on same instance?

Ram81 (changes requested):

burnerlee:
we can wait for #3537 to get merged and then I would shift the alertmanager container configs to `docker-compose-monitoring.yml` Rest, the PR is finalised

docker/prod/nginx-ingress/nginx_staging.conf
```diff
@@ -54,4 +58,11 @@ server {
     proxy_set_header X-Forwarded-Proto $scheme;
     proxy_pass http://statsd_exporter;
   }
+  
+  location /alerts {
```
Ram81, line 62: this should be `alert_manager`

monitoring/alertmanager/templates/instances.tmpl
```diff
@@ -0,0 +1,18 @@
+{{ define "slack.instances.title" -}}
```
Ram81, line 1: @burnerlee what is the purpose of this file?

monitoring/prometheus/alertmanager.yml
```diff
@@ -0,0 +1,38 @@
...
+        title: '{{ .commonAnnotations.title }}'
+        text: '{{ .commonAnnotations.description }}'
+        send_resolved: false
+  - name: 'slack-apis-notifications'
+    slack_configs:
+      - channel: '#x'
+        title: '{{ template "slack.apis.title" . }}'
+        text: '{{ template "slack.apis.text" . }}'
+        send_resolved: false
+
```
Ram81, line 28: Please remove this newline

monitoring/prometheus/alertmanager.yml
```diff
@@ -0,0 +1,38 @@
...
+  - name: 'slack-instance-notifications'
+    slack_configs:
+      - channel: '#x'
+        title: '{{ template "slack.instances.title" . }}'
+        text: '{{ template "slack.instances.text" . }}'
+        send_resolved: false
+
+templates:
+- '/etc/alertmanager/templates/instances.tmpl'
+- '/etc/alertmanager/templates/apis.tmpl'
```
Ram81, line 38: Is this indented correctly?

monitoring/prometheus/alertmanager.yml
```diff
@@ -0,0 +1,38 @@
...
+      - channel: '#x'
+        title: '{{ .commonAnnotations.title }}'
+        text: '{{ .commonAnnotations.description }}'
+        send_resolved: false
+  - name: 'slack-apis-notifications'
+    slack_configs:
+      - channel: '#x'
+        title: '{{ template "slack.apis.title" . }}'
+        text: '{{ template "slack.apis.text" . }}'
+        send_resolved: false
```
Ram81, line 27: Why do we have multiple receivers? There should be one receiver channel, no?

monitoring/prometheus/prometheus_staging.yml
```diff
@@ -18,3 +21,9 @@ scrape_configs:
     metrics_path: '/node_exporter'
     static_configs:
       - targets: ['staging.eval.ai']
+
+alerting:
+  alertmanagers:
+  - path_prefix: '/alerts'
+    static_configs:
+    - targets: ['monitoring-staging.eval.ai']
```
Ram81, line 29: add newline

docker-compose-staging.yml
```diff
@@ -150,6 +151,22 @@ services:
       - prometheus
       - grafana
       - statsd-exporter
+      - alertmanager
     ports:
       - '80:80'
       - '443:443'
+    
+  alertmanager:
```
Ram81, line 159: This should be `alert-manager`

Ram81 (changes requested):

monitoring/alertmanager/templates/instances.tmpl
burnerlee: @Ram81 this is to define the template of the alert message that we are receiving on slack

burnerlee: We have 2 templates one for instance-down alerts and one for API Threshold exceeded alerts

monitoring/prometheus/alertmanager.yml
burnerlee: @Ram81 yeah works fine in the dev environment

monitoring/prometheus/alertmanager.yml
burnerlee: @Ram81 I have grouped all the possible alerts into two groups - Instance Alerts and API alerts. Both of these different templates due to different tags they have. For each alert group, we need to define receivers cause even though the channel is the same, we also provide the template configs here. Though, we can remove the default-receiver which I had kept to make the code a bit generic and use that receiver for any alerts that do not group under the api or instance groups. However, at the moment we are not emitting any such alerts. So, I added this as it may be needed sometime in the future

Ram81: @burnerlee please remove the default config. We only need to add the ones we are using

Ram81 (changes requested):
Minor change. Rest of it LGTM

monitoring/prometheus/alertmanager.yml
```diff
@@ -0,0 +1,31 @@
+global:
```
Ram81, line 1: Please rename the file to `alert_manager.yml`

Ram81 (changes requested):
minor change

Ram81 (approved):
LGTM

RishabhJain2018 (approved):

PR #3512, opened 2021-06-30
update travis auto deployment scripts and add configs for staging environment

burnerlee:
update staging and production deploy scripts to add Prometheus and Grafana services. updated commands:
`scripts/deployment/deploy.sh deploy-grafana`
`scripts/deployment/deploy.sh deploy-prometheus`
`scripts/deployment/deploy.sh deploy-statsd-exporter`
`scripts/deployment/deploy.sh auto-deploy`


docker/prod/nodejs/nginx_staging.conf
```diff
@@ -2,6 +2,10 @@ upstream django_app {
   server django:8000 fail_timeout=0;
 }
 
+upstream statsd_exporter {
```
Ram81, line 5: Why are we changing nginx config if the `statsd_exporter` is not needed on the same instance as EvalAI backend?

Ram81 (changes requested):

docker/prod/nodejs/nginx_staging.conf
burnerlee: @Ram81 actually it would be good if we can keep the statsd container along with node-exporter container (as django backend) on the same container as metrics are streamed to statsd for each API call from the backend

Ram81 (changes requested):
@burnerlee please rebase this PR

scripts/deployment/deploy.sh
```diff
@@ -135,6 +135,21 @@ case $opt in
...
+            echo "Deploying prometheus docker container..."
+            docker-compose -f docker-compose-${env}.yml up -d prometheus
+            echo "Completed deploy operation."
+            ;;
+        deploy-grafana)
+            echo "Deploying grafana docker container..."
+            docker-compose -f docker-compose-${env}.yml up -d grafana
+            echo "Completed deploy operation."
+            ;;
+        deploy-statsd-exporter)
```
Ram81, line 148: @burnerlee change this to `deploy-statsd`

Ram81 (changes requested):
@burnerlee minor change. Rest of it LGTM

Ram81 (approved):
LGTM 👍

PR #3517, opened 2021-07-03
Fix notification error and footer inconsistencies

gautamjajoo:
The PR fixes these two issues:
- [x] Notification not visible when under start button
- [x] The footer on home page is not consistent.

@Kajol-Kumari @Ram81


Ram81:
@gautamjajoo can post screenshots of the page before and after the fix

gautamjajoo:
**Before Fix:**
Screenshot from 2021-07-03 14-44-28
Screenshot from 2021-04-01 20-29-52

**After Fix:**
Screenshot from 2021-07-02 19-32-53

Ram81 (approved):

RishabhJain2018 (approved):

PR #3492, opened 2021-06-20
Refactor code upload challenge worker for submission stuck

savish28:
Context: In the Code upload challenge, if the job is not present on the cluster because of any external means (deleted etc). The submission message is stuck in the queue.

- [x] Refactor code upload challenge worker for submission stuck

This PR is in relation to #3452, would update this PR or 3452 based on whichever is merged first.

@KhalidRmb 
cc @Ram81

PR #3523, opened 2021-07-06
Cleanup submission: file not found before TLE

savish28:
Context: Monitoring static code upload challenge

Deliverables:
- [x] Cleanup submission code upload: file not found

Depends on: #3492 

@KhalidRmb 
cc: @Ram81


scripts/workers/code_upload_submission_worker.py
```diff
@@ -469,48 +509,81 @@ def update_failed_jobs_and_send_logs(
...
+                            logger.exception(
+                                "Exception while reading Job logs {}".format(e)
+                            )
+                    elif (
+                        container_name == "submission"
+                        and container_state.terminated.reason == "Completed"
+                        and container_state_map.get("sidecar-container")
+                        and container_state_map.get(
+                            "sidecar-container"
+                        ).terminated
```
Ram81, line 562: @savish28 let's move this check to a separate function. This is not readable

Ram81 (changes requested):
@savish28 could you describe the conditions in the change in PR description. It's not clear what is changed

scripts/workers/code_upload_submission_worker.py
savish28: @Ram81 Updated

scripts/workers/code_upload_submission_worker.py
```diff
@@ -518,6 +518,25 @@ def cleanup_submission(
         )
 
 
+def is_submission_container_executed(
```
Ram81, line 521: Please rename this method to `is_submission_evaluation_done`

scripts/workers/code_upload_submission_worker.py
```diff
@@ -567,6 +589,19 @@ def update_failed_jobs_and_send_logs(
                             logger.exception(
                                 "Exception while reading Job logs {}".format(e)
                             )
+                    elif is_submission_container_executed(
+                        container_name, container_state, container_state_map
+                    ):
```
Ram81, line 594: Also add a one line comment describing the check

Ram81 (changes requested):

scripts/workers/code_upload_submission_worker.py
```diff
@@ -567,6 +589,19 @@ def update_failed_jobs_and_send_logs(
                             logger.exception(
                                 "Exception while reading Job logs {}".format(e)
                             )
+                    elif is_submission_container_executed(
```
KhalidRmb, line 592: @savish28 Check if it's a submission container as a first condition here itself. You can remove it from inside the function

scripts/workers/code_upload_submission_worker.py
```diff
@@ -567,6 +589,19 @@ def update_failed_jobs_and_send_logs(
...
+                        container_name, container_state, container_state_map
+                    ):
+                        cleanup_submission(
```
KhalidRmb, line 595: @savish28 Refactoring: This call is repeated in this function 3 times with only one param being different. Just use a boolean to set cleanup and store the log/error. Then just call the function once at the end

scripts/workers/code_upload_submission_worker.py
savish28: @Ram81 Docstring is added in the function heading, should I add that here as well?

scripts/workers/code_upload_submission_worker.py
savish28: @KhalidRmb We were checking the conditions here only but for better readability, we shifted the checks to the function. Do we want the submission condition to be segregated from the function?

scripts/workers/code_upload_submission_worker.py
savish28: Good idea,sure.

scripts/workers/code_upload_submission_worker.py
KhalidRmb: Hmm...just the one check. Because why enter the function if it's not the submission container

scripts/workers/code_upload_submission_worker.py
savish28: Updated

scripts/workers/code_upload_submission_worker.py
savish28: Updated

scripts/workers/code_upload_submission_worker.py
savish28: Updated.

Ram81 (approved):
LGTM

RishabhJain2018 (approved):

PR #3526, opened 2021-07-09
clear metrics after pushing to pushgateway to avoid duplication of metrics

burnerlee:
once metrics are pushed to gateway for a particular submission id and queue, we want to clear that metric from the registry, so that it is not pushed again for the next submission

### DIFF
--- a/apps/jobs/sender.py
+++ b/apps/jobs/sender.py
@@ -11,7 +11,7 @@ from django.conf import settings
 from base.utils import send_slack_notification
 from challenges.models import Challenge
 from .utils import get_submission_model
-from .metrics import push_metrics_to_pushgateway, num_submissions_in_queue
+from .metrics import push_metrics_to_pushgateway, num_submissions_in_queue, num_processed_submissions
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


apps/jobs/sender.py
```diff
@@ -87,9 +87,13 @@ def publish_submission_message(message):
     # increase counter for submission pushed into queue
     submission_pk = message["submission_pk"]
     num_submissions_in_queue.labels(submission_pk, queue_name).inc()
+    num_processed_submissions.labels(submission_pk, queue_name).inc(0)
```
Ram81, line 90: Why are we doing this? Only submissions pending for evaluation are published. Incrementing metrics of processed submissions by `0` doesn't make sense. If we are doing this because of the inconsistency in the metric values currently then this is not the correct approach

apps/jobs/sender.py
```diff
@@ -87,9 +87,13 @@ def publish_submission_message(message):
...
     # push metrics to pushgateway as a submission is pushed into queue
     job_id = "submission_{}".format(submission_pk)
     push_metrics_to_pushgateway(job_id)
+    # clear metrics after they are pushed to avoid duplication on next push
+    num_submissions_in_queue.clear()
+    num_processed_submissions.clear()
```
Ram81, line 96: `num_processed_submissions` should be updated only by the evaluation worker

Ram81 (changes requested):

apps/jobs/sender.py
```diff
@@ -90,6 +90,8 @@ def publish_submission_message(message):
     # push metrics to pushgateway as a submission is pushed into queue
     job_id = "submission_{}".format(submission_pk)
     push_metrics_to_pushgateway(job_id)
+    # clear metrics after they are pushed to avoid duplication on next push
+    num_submissions_in_queue.clear()
```
Ram81, line 94: Why do we have to clear the metrics explicitly? I have used prometheus metrics we never had to clear the metrics.

Ram81 (changes requested):

Ram81 (commented):
Can we close this PR

burnerlee:
Closing this PR as we have shifted to Statsd-Exporter for queue monitoring

PR #3530, opened 2021-07-11
Aws: Add multi key support submission and code upload container

savish28:
Context: Static code upload challenge Fargate workers.

Deliverables:
- [x] Add multi AWS key support submission and code upload container.

This PR solves the bug of `AWS_STORAGE_BUCKET_NAME = ""` on fargate.
@KhalidRmb 
cc: @Ram81


apps/challenges/aws_utils.py
```diff
@@ -311,205 +311,212 @@
...
+    "cpu": "{CPU}",
+    "memory": "{MEMORY}",
+}}
+"""
 
-                {{
-                    "name": "AUTH_TOKEN",
-                    "value": "{auth_token}"
-                }},
+submission_container_defination = """
```
Ram81, line 325: Please fix the variable name

apps/challenges/aws_utils.py
```diff
@@ -311,205 +311,212 @@
...
+{{
```
Ram81, line 326: @savish28 Also, we can't have have these huge task definitions as variables in `aws_utils.py`. Please move these to a separate file.

apps/challenges/aws_utils.py
```diff
@@ -311,205 +311,212 @@
...
+        }},
+    }},
+}}
+"""
 
-                {{
-                    "name": "EVALAI_DNS",
-                    "value": "{EVALAI_DNS}"
-                }},
+code_upload_container_defination = """
```
Ram81, line 450: Fix the variable name

apps/challenges/aws_utils.py
```diff
@@ -668,25 +675,39 @@ def register_task_def_by_challenge_pk(client, queue_name, challenge):
...
-                    **challenge_aws_keys,
+                    **aws_keys,
+                )
+                definition = task_definition_static_code_upload_worker.format(
+                    queue_name=queue_name,
+                    code_upload_container=code_upload_container,
+                    submission_container=submission_container,
+                    CPU=worker_cpu_cores,
+                    MEMORY=worker_memory,
+                    **COMMON_SETTINGS_DICT,
```
Ram81, line 710: Please move this to a separate method. This piece of code is unreadable. Also, why do we need to substitute same variables in two task definitions can't we dedup it and make this code more readble?

Ram81 (changes requested):
@savish28 can we refactor the task definition creation code for static code upload challenge in the PR. Current changes are unreadable.

apps/challenges/aws_utils.py
savish28: @Ram81 Should we shift all task defn to a new file say `task_definations.py` and import all from there? or just these two container defns?

apps/challenges/aws_utils.py
savish28: Sure, we can move that to a different function, the common variables are `queue_name` and `COMMON_SETTINGS_DICT`, but still since `code_upload_container` and `submission_container_defination` are passed as a variable to `task_definition_static_code_upload_worker` so we need to format the all the varaibles of them before right? before passing

savish28: @Ram81, Should we test these changes first on fargate and if everything works we can maybe add the container definitions in  task_defination and task_definations_code_upload_worker as well and create three functions on `get_challenge_task_defn`, `get_code_upload_challenge_task_defn`, `get_static_dataset_code_upload_challenge_task_defn` in `task_definations.py` and eventually this peice of code can shift to `get_static_dataset_code_upload_challenge_task_defn` function.

apps/challenges/aws_utils.py
savish28: Updated

apps/challenges/aws_utils.py
savish28: Updated

apps/challenges/aws_utils.py
savish28: Updated

apps/challenges/aws_utils.py
```diff
@@ -16,6 +16,16 @@
     construct_and_send_worker_start_mail,
     construct_and_send_eks_cluster_creation_mail,
 )
+from .task_definations import (
```
Ram81, line 19: It should be `task_definitions` not `task_definations`. Please replace `definations` with `definition`

Ram81 (changes requested):
Please fix the spelling in file name as well

apps/challenges/aws_utils.py
savish28: Updated.

Ram81 (approved):

RishabhJain2018 (approved):

PR #3533, opened 2021-07-16
Update api_instacne for kubernetes token refresh

savish28:
Context: Code upload challenge worker.

Bug:
- [x] api_instance did not update so caused authentication error

@KhalidRmb 
cc: @Ram81


Ram81 (approved):
LGTM

PR #3536, opened 2021-07-18
Refactor static code upload challenge workers for logs, wait time and exception

savish28:
Context: Static code upload challenge workers.

Deliverables-
- [x] Add logs in code upload worker.
- [x] Change time delta -> To have monitoring setup to job if no file present.
- [x] Add time sleep in code upload challenge for messages meant for submission worker to reduce evaluation wait time.
- [x] Add exception handling in case submission input file is none in submission worker.

@KhalidRmb 
cc: @Ram81


scripts/workers/code_upload_submission_worker.py
```diff
@@ -262,7 +272,7 @@ def create_static_code_upload_submission_job_object(message):
         value=str(submission_meta["submission_time_limit"]),
     )
     SUBMISSION_TIME_DELTA_ENV = client.V1EnvVar(
-        name="SUBMISSION_TIME_DELTA", value="3600"
+        name="SUBMISSION_TIME_DELTA", value="1"
```
Ram81, line 275: This value to should be 1 hour. Let's not check in values like 1 second for testing purposes only

scripts/workers/code_upload_submission_worker.py
```diff
@@ -684,6 +694,7 @@ def main():
             ) and not message_body.get(
                 "is_static_dataset_code_upload_submission"
             ):
+                time.sleep(35)
```
Ram81, line 697: Why this big wait time?

scripts/workers/submission_worker.py
```diff
@@ -361,7 +361,15 @@ def extract_submission_data(submission_id):
...
-    submission_input_file = input_file.url
+    try:
+        submission_input_file = input_file.url
+    except Exception as e:
+        logger.exception(
+            "{} Submission input file does not exist for submission {}, error {}".format(
+                SUBMISSION_LOGS_PREFIX, submission_id, e
+            )
+        )
+        return None
```
Ram81, line 372: This case should never happen. `input_file` should never be `None`. This needs to be fixed when making a submission/updating it. We can;'t handle it in submission worker.

Ram81 (changes requested):

scripts/workers/code_upload_submission_worker.py
savish28: @Ram81 actually this is set on purpose.
Assume the following situation.

- time_delta is 1 hr
- submission worker stops after 15 mins and saving the file.
- whenever code_upload_worker would receive the message and see submission worker is completed but sidecar is running. So it would notify EvalAI that no submission file was found.
- So in order to support delete of job and updation to evalai, we need to have it 1 sec.
- Because the message is checked every 0.5 secs in code upload worker now.

scripts/workers/code_upload_submission_worker.py
savish28: @Ram81 This is to guarantee that submission worker reads this message and evaluates it. Waittime of a message is 30 secs in SQS by default. This approach removes the probablistic approach is more inclined towards deterministic and which increase speed.
Tested this approach and the whole submission lifecycle time was reduced drastically due to this.

scripts/workers/submission_worker.py
savish28: Yeah agreed, this should not happen here. 
This is a rare thing and the events which I interpreted were:
- Both the requests Submission file from the side-car container and "submission.json/csv not found" from the code_upload worker happened in parallel.
[Status update request] fetched submission object which did not have submission_input_file.
[Submisison file request] came and updated the db and created a message and returned.
[Status update request] used the variable in memory that had submission_input_file as None.

Added the exception handling here just in case. I totally agree we should take care of this.

savish28: If we have less time_delta than side-car container would exit as soon as the submission container and this situation would not occur. This is also one of the reasons of having less time_delta.

savish28:
@Ram81 Ping!

Ram81 (approved):
LGTM

RishabhJain2018 (approved):

PR #3539, opened 2021-07-19
fix auto deploy and auto push commands for monitoring instance

burnerlee:


burnerlee:
already a part of #3542

PR #3540, opened 2021-07-19
remove pushgateway from monitoring setup

burnerlee:
this PR removes pushgateway from the monitoring setup. We would use statsd-exporter instead of pushgateway to monitor the queue services


Ram81 (approved):
LGTM

PR #3542, opened 2021-07-20
migrate statsd to monitoring instance and use it for queue monitoring

burnerlee:


apps/jobs/sender.py
```diff
@@ -83,6 +84,16 @@ def publish_submission_message(message):
...
+    # increase counter for submission pushed into queue
+    submission_pk = message["submission_pk"]
+    statsd.gauge(
+        NUM_SUBMISSIONS_IN_QUEUE,
+        1,
+        tags=[
+            "queue_name:%s" % queue_name,
+            "submission_pk:%s" % submission_pk,
+        ],
+    )
```
Ram81, line 96: call for `statsd.gauge` should be in a util method. You should pass `metric_name` and `tags` to this util method

monitoring/statsd/metrics.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+statsd_host = os.environ.get("STATSD_ENDPOINT")
+statsd_port = int(os.environ.get("STATSD_PORT"))
+statsd = DogStatsd(host=statsd_host, port=statsd_port)
+
+REQUEST_LATENCY_METRIC_NAME = "django_request_latency_seconds"
+REQUEST_COUNT_METRIC_NAME = "django_request_count"
+NUM_SUBMISSIONS_IN_QUEUE = "num_submissions_in_queue"
+NUM_PROCESSED_SUBMISSIONS = "num_processed_submissions"
```
Ram81, line 13: You should put the utils method to increment a metric here

scripts/workers/submission_worker.py
```diff
@@ -750,6 +751,25 @@ def load_challenge_and_return_max_submissions(q_params):
...
+    try:
+        submission_pk = json.loads(body)["submission_pk"]
+        statsd.gauge(
+            NUM_PROCESSED_SUBMISSIONS,
+            1,
+            tags=[
+                "queue_name:%s" % queue_name,
+                "submission_pk:%s" % submission_pk,
+            ],
+        )
```
Ram81, line 764: same here

scripts/workers/submission_worker.py
```diff
@@ -750,6 +751,25 @@ def load_challenge_and_return_max_submissions(q_params):
     return maximum_concurrent_submissions, challenge
 
 
+def increment_and_push_metrics_to_statsd(body, queue_name):
+    try:
+        submission_pk = json.loads(body)["submission_pk"]
```
Ram81, line 756: message body is loaded from yaml. Refer [this](https://github.com/Cloud-CV/EvalAI/pull/3542/files#diff-ae6210f157b1ef66d873b08fe30f65f36c92c1e3d8aa61e3c8580910f5f06695L690-L691)

Ram81 (changes requested):

monitoring/statsd/metrics.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+def set_statsd_gauge(metric_name, tags, set_value):
+    statsd.gauge(metric_name, set_value, tags=tags)
```
Ram81, line 17: Why is there a gauge and increment method? We only need to increment the metric right?

monitoring/statsd/metrics.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+def increment_statsd_counter(metric_name, tags, inc_value):
+    statsd.increment(metric_name, inc_value, tags=tags)
```
Ram81, line 21: Values will be incremented by 1. Why do we have to pass increment value explicitly?

Ram81 (changes requested):
@burnerlee the changes from PR #3539 are part of this so we should close that as the changes are just two lines

monitoring/statsd/metrics.py
burnerlee: @Ram81 I didn't hardcore the value here to one as I thought it would be good to have a generic util function in which we can pass a custom increment value. Otherwise we can hardcore the increment value to one in the util function

burnerlee: Prometheus counters have an option to increment any counter by any value and not just one, moreover it's a statsd metrics here which is converted to prometheus library by datadog

Ram81 (approved):
LGTM

RishabhJain2018 (approved):

RishabhJain2018:
@burnerlee Can you please add a description for the PR.

burnerlee:
@RishabhJain2018 added the description

PR #3543, opened 2021-07-20
Increase decimal precision to 20 and display the previous precision

gautamjajoo:
Increases decimal precision to 20 and also displays the value received from backend.

@Ram81 @Kajol-Kumari


Ram81 (approved):
LGTM

RishabhJain2018 (approved):

PR #3550, opened 2021-07-25
Fix console error of change in expression value

gautamjajoo:
Fix the error that occurs in dev mode when the change detection run twice.

**Before**
image

**After**
No error in console.

@Ram81 @Kajol-Kumari

### DIFF
--- a/frontend_v2/src/app/components/publiclists/publiclists.component.ts
+++ b/frontend_v2/src/app/components/publiclists/publiclists.component.ts
@@ -1,4 +1,4 @@
-import { Component, OnInit, Inject, OnDestroy } from '@angular/core';
+import { Component, OnInit, Inject, OnDestroy, AfterViewChecked, ChangeDetectorRef } from '@angular/core';
 import { Router, ActivatedRoute } from '@angular/router';
 import { DOCUMENT } from '@angular/common';
 import { GlobalService } from '../../services/global.service';
@@ -12,7 +12,10 @@ import { AuthService } from '../../services/auth.service';
   templateUrl: './publiclists.component.html',
   styleUrls: ['./publiclists.component.scss'],
 })
-export class PubliclistsComponent {
+export class PubliclistsComponent implements AfterViewChecked {
+
+  isAuth = false;
+
   /**
    * Constructor.
    * @param document  Window document Injection.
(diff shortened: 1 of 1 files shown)


Kajol-Kumari (approved):
@gautamjajoo as you have introduced the `isAuth` variable. Let's use this variable only at [here](https://github.com/Cloud-CV/EvalAI/blob/master/frontend_v2/src/app/components/publiclists/publiclists.component.html#L1)

Rest LGTM 👍

gautamjajoo:
@Kajol-Kumari I have not replaced it because first the value is `false` that's why the page is distorted when it is loading and then the value becomes `true`.

Kajol-Kumari:
> @Kajol-Kumari I have not replaced it because first the value is `false` that's why the page is distorted when it is loading and then the value becomes `true`.

yes @gautamjajoo but now as we are using `detectChanges()` so it's totally fine to use the `isAuth` variable in the html file

gautamjajoo:
Actually, before opening the PR I tried to replace with the variable but the page is like this
Screenshot from 2021-07-26 21-07-29

Kajol-Kumari:
> Actually, before opening the PR I tried to replace with the variable but the page is like this
> Screenshot from 2021-07-26 21-07-29

That's weird as It shouldn't happen ideally. Also, I tried doing the change on my local and it seems to work fine for me. Can you please try testing it again by replacing `authService.isAuth` with `isAuth` throughout the publiclist component html file? @gautamjajoo

gautamjajoo:
@Kajol-Kumari I did change all the variables. I believe there is some issue with the changing values. Would it be better if we keep it same as `authService.isAuth` and consistent with other pages.

Kajol-Kumari:
> @Kajol-Kumari I did change all the variables. I believe there is some issue with the changing values. Would it be better if we keep it same as `authService.isAuth` and consistent with other pages?

yeah, we can keep this as this doing the same work but we need to figure out the reason behind it.

gautamjajoo:
@Kajol-Kumari Replaced with `isAuth` variable, now assigning the value upon component initialization as well.

Kajol-Kumari (approved):
@gautamjajoo please format the comment for `ngAfterViewChecked` [The one line gap should be before comment, not after]

Rest, LGTM 👍

gautamjajoo:
@Kajol-Kumari Done the changes!

Ram81 (changes requested):
@gautamjajoo can you resolve the merge conflicts

gautamjajoo:
@Ram81 Resolve the merge conflicts!

PR #3557, opened 2021-07-27
Fetch private submissions after polling when checkbox ticked

gautamjajoo:
When `Include Private Submission` is checked the private submissions are not fetched after 5seconds of polling. 
Also, when the leaderboard has all private submissions, previously no results are shown, but now the checkbox is also visible.

@Ram81 @Kajol-Kumari

### DIFF
--- a/frontend_v2/src/app/components/challenge/challengeleaderboard/challengeleaderboard.component.html
+++ b/frontend_v2/src/app/components/challenge/challengeleaderboard/challengeleaderboard.component.html
@@ -65,11 +65,14 @@
             <div *ngIf="leaderboard.length <= 0 && selectedPhaseSplit == null" class="fw-light result-wrn">
               No phase selected.
             </div>
-            <div *ngIf="leaderboard.length > 0 && selectedPhaseSplit" class="result-wrn content fw-light">
-              <mat-chip-list> <mat-chip>B</mat-chip> - Baseline submission </mat-chip-list>
-              <mat-chip-list class="leaderboard-description-span private">
-                <mat-chip>P</mat-chip> - Private submission
-              </mat-chip-list>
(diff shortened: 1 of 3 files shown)


frontend_v2/src/app/components/challenge/challengeleaderboard/challengeleaderboard.component.html
```diff
@@ -87,14 +90,15 @@ <h5 class="fw-light">Leaderboard</h5>
...
-              <mat-checkbox
-                class="pull-right complete-leaderboard"
-                *ngIf="isChallengeHost && showLeaderboardToggle"
-                (change)="toggleLeaderboard(getAllEntries)"
-                [(ngModel)]="getAllEntries"
-              >
-                <label class="fw-light complete-leaderboard-text"> {{ getAllEntriesTextOption }}</label>
-              </mat-checkbox>
+              </div>
+                <mat-checkbox
```
Kajol-Kumari, line 94: @gautamjajoo  This checkbox will appear if we have total submission `0` We need a check here that only if my total submission(both public and private) is >0, then we should show this checkbox

Kajol-Kumari (changes requested):

frontend_v2/src/app/components/challenge/challengeleaderboard/challengeleaderboard.component.html
gautamjajoo: Fixed. Now the checkbox only appears when `public + private > 0`.

Kajol-Kumari (approved):
LGTM 👍

Ram81 (approved):
LGTM

RishabhJain2018 (approved):

PR #3558, opened 2021-07-28
fix: switch to http for prometheus and grafana

burnerlee:
Currently using https for prometheus and grafana in our nginx-ingress config throws SSL Handshake error. This PR fixes this to host the services flawlessly

### DIFF
--- a/docker-compose-staging.yml
+++ b/docker-compose-staging.yml
@@ -130,7 +130,8 @@ services:
       - '--log.level=info'
       - '--web.telemetry-path=/statsd/metrics'
     ports:
-      - '9125:9125'
+      - '9125:9125/udp'
+      - '9125:9125/tcp'
       - '9102:9102'
   
   node_exporter:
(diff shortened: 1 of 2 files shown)


docker-compose-staging.yml
```diff
@@ -130,7 +130,8 @@ services:
       - '--log.level=info'
       - '--web.telemetry-path=/statsd/metrics'
     ports:
-      - '9125:9125'
+      - '9125:9125/udp'
+      - '9125:9125/tcp'
```
Ram81, line 134: @burnerlee why is this needed?

Ram81 (changes requested):

docker-compose-staging.yml
burnerlee: @Ram81 by default, docker only exposes the tcp port, while the statsd-exporter listens to both the tcp and udp port. So I have to explicity mention both of them for both of them to be exposed

burnerlee:
One file was pushed earlier by mistake. This PR removes that file as well

Ram81 (approved):
LGTM

RishabhJain2018 (approved):

PR #3559, opened 2021-07-29
[Backend] Add cron job to stop workers for unapproved challenges

Ram81:
### Description

This PR adds -

- [x] Changes to cron job to delete workers for challenges that have been unapproved for 3 days
- [x] API changes to delete workers only by EvalAI admins
- [x] Bug fix in existing logic to delete workers after challenge ends

### DIFF
--- a/scripts/monitoring/auto_start_or_stop_workers.py
+++ b/scripts/monitoring/auto_start_or_stop_workers.py
@@ -36,8 +36,20 @@ def stop_worker(challenge_id):
     return response
 
 
-def get_all_challenges():
-    all_challenge_endpoint = "{}/api/challenges/challenge/present".format(
+def delete_worker(challenge_id):
+    delete_worker_endpoint = (
+        "{}/api/challenges/{}/manage_worker/delete/".format(
+            evalai_endpoint, challenge_id
+        )
+    )
+    response = requests.put(
+        delete_worker_endpoint, headers=authorization_header
+    )
+    return response
+
+
+def get_challenges():
+    all_challenge_endpoint = "{}/api/challenges/challenge/all".format(
         evalai_endpoint
     )
     response = requests.get(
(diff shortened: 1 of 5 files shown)


scripts/monitoring/auto_start_or_stop_workers.py
```diff
@@ -46,32 +58,54 @@ def get_all_challenges():
...
+        approved_by_admin = challenge["approved_by_admin"]
         is_docker_based = challenge["is_docker_based"]
         challenge_start_date = parser.parse(challenge["start_date"])
         challenge_end_date = parser.parse(challenge["end_date"])
+        created_at = parser.parse(challenge["created_at"])
         current_date = utc.localize(datetime.now())
 
         print(
-            "Start submission worker for challenge id {}".format(challenge_id)
+            "Start/Stop submission worker for challenge id {}".format(
```
RishabhJain2018, line 82: Can we please be specific with the message? Also, why are we using `print` statement? Would this be logged on Cloudwatch?

scripts/monitoring/auto_start_or_stop_workers.py
```diff
@@ -46,32 +58,54 @@ def get_all_challenges():
...
                 response = start_worker(challenge_id)
                 if not response.ok:
                     print(
                         "ERROR: Start worker failed for challenge id {}!".format(
                             challenge_id
                         )
                     )
-            if workers is not None and challenge_end_date <= current_date:
-                response = stop_worker(challenge_id)
+            # Delete workers for challenges uploaded in last 3 days that are unapproved or inactive challenges
```
RishabhJain2018, line 99: Please make it 24 hours.

scripts/monitoring/auto_start_or_stop_workers.py
```diff
@@ -46,32 +58,54 @@ def get_all_challenges():
...
+            if (
+                workers is not None and challenge_end_date < current_date
+            ) or is_unapproved_challenge(
+                workers, approved_by_admin, created_at
+            ):
+                response = delete_worker(challenge_id)
                 if not response.ok:
                     print(
-                        "ERROR: Stop worker failed for challenge id {}!".format(
+                        "ERROR: Delete worker failed for challenge id {}!".format(
```
RishabhJain2018, line 108: We should trigger sentry. I don't think `print` will help us here as we won't check the logs manually.

RishabhJain2018 (changes requested):

scripts/monitoring/auto_start_or_stop_workers.py
Ram81: This is a independent cron that we run on prod and staging instance. Do we want the logs on cloudwatch?

Ram81:
@RishabhJain2018 this is a independent cron that I was planning to run in `crontab` on the instance. Do we want this to run inside django container?

RishabhJain2018:
Ummm..then how do we catch errors from this cron job?

Ram81:
The logs are available in crontab. If something is wrong then I just check there. As these are fixed API calls which are already validated beforehand there have been no errors so far.

RishabhJain2018:
okay.

RishabhJain2018 (approved):

PR #3560, opened 2021-07-29
[Worker]: BugFix code upload worker for limited nodes

savish28:
Context: Code upload worker.

This PR solves the following bugs

- [x] Prevent Logs in code upload worker being printed twice.
- [x] Fix read_job function in case of job not present.
- [x] Prevent failure of submission when pod is in pending state. [nodes are fully utilized and a new job is created]

Detailed Explanation for point 3:
This scenario occurs when the job is created but the number of nodes available are fully utilized, so the job is created but the corresponding pod is in a pending state in which the container statuses does not exist. 

@KhalidRmb 
cc: @Ram81

### DIFF
--- a/scripts/workers/code_upload_submission_worker.py
+++ b/scripts/workers/code_upload_submission_worker.py
@@ -36,6 +36,7 @@ handler.setFormatter(formatter)
 logger = logging.getLogger(__name__)
 logger.addHandler(handler)
 logger.setLevel(logging.INFO)
+logger.propagate = False
 
 AUTH_TOKEN = os.environ.get("AUTH_TOKEN", "auth_token")
 EVALAI_API_SERVER = os.environ.get(
@@ -487,6 +488,7 @@ def read_job(api_instance, job_name):
         api_response = api_instance.read_namespaced_job(job_name, namespace)
     except ApiException as e:
         logger.exception("Exception while reading Job with error {}".format(e))
+        return None
     return api_response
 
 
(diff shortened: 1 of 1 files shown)


scripts/workers/code_upload_submission_worker.py
```diff
@@ -550,38 +552,60 @@ def update_failed_jobs_and_send_logs(
...
-                                )
-                            )
-                            pod_log = pod_log_response.data.decode("utf-8")
-                            clean_submission = True
-                            submission_error = pod_log
-                        except client.rest.ApiException as e:
-                            logger.exception(
-                                "Exception while reading Job logs {}".format(e)
-                            )
+        if job_def:
```
Ram81, line 555: @savish28 `read_job` will return a valid kubernetes job if the job is in pending state. Have you tested this change?

Ram81 (changes requested):

scripts/workers/code_upload_submission_worker.py
savish28: @Ram81 this is the case when I manually delete the job from kubectl and in this case read_job would return None, this is a different usecase.

savish28: Yeah, I have tested this change

Ram81: Is this the reason why fargate worker was failing?

savish28: Not sure, why the fargate was restarting is still a mystery

Ram81 (approved):
LGTM

RishabhJain2018 (approved):

Issue #3485, opened 2021-06-19
2 proposed feature requests for leaderboard


johnwlambert:
Hi EvalAI team, thanks for all your hard work on an excellent platform. I have two proposed feature requests:

1. When I click on an entry in a leaderboard (say the 5th entry), the leaderboard re-centers on the object, and hides the objects above, which is a bit counter-intuitive. I would prefer behavior where that entry is actually highlighted.

2. Could we have an option to show on the leaderboard which columns have "higher is better" or "lower is better"? (via arrows)

(e.g. as shown in Table 4 of [this paper](https://arxiv.org/pdf/2006.11275.pdf))
<img width="1130" alt="Screen Shot 2021-06-19 at 1 53 25 PM" src="https://user-images.githubusercontent.com/16724970/122651271-ba906900-d105-11eb-9825-53fdd69e5af8.png">

RishabhJain2018:
Hi @johnwlambert, Thank you for the feature requests. We will try to release these ASAP.

RishabhJain2018:
Closing as this is fixed!

PR #3561, opened 2021-08-01
[Frontend] Add Leaderboard metric order metadata

Ram81:
### Description

This PR adds feature 2 proposed in #3485 - 

- [x] Show "Higher is better" or "Lower is better" for each leaderboard metric
- [x]  Show metric description in a tooltip when a user hovers on it

Required schema changes are available [here](https://github.com/Cloud-CV/EvalAI-Starters/pull/51)


Ram81:
Screenshot for reference:

Screen Shot 2021-08-02 at 1 04 41 AM

frontend/src/css/modules/leaderboard.scss
```diff
@@ -67,3 +67,24 @@ span.badge.partial-evaluation {
...
+    font-size: 1.05rem;
+    
+    /* Position the tooltip */
+    position: absolute;
+    z-index: 1;
+}
+
+.leaderboard-label:hover .description {
+    visibility: visible;
+}
```
RishabhJain2018, line 90: Please add a new line here.

RishabhJain2018 (changes requested):
@Ram81 What about the models and API changes for this feature?

Ram81:
@RishabhJain2018 we are adding metadata as part of leaderboard schema [here](https://github.com/Cloud-CV/EvalAI-Starters/pull/51) so we won't need any model and API changes. Do we want to add separate field for metadata?

RishabhJain2018:
yeah, that works but I don't see any API changes in the PR to get that value from the config.yaml and then store it?

RishabhJain2018:
Oh, nvm, got it.

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -876,6 +876,33 @@
...
+                if (metadata[metric] == undefined) {
+                    return false;
+                }
+                return metadata[metric].order_ascending;
+            }
+            return false;
+        };
+
+        vm.getLabelDescription = function(metric) {
+            console.log("");
```
RishabhJain2018, line 893: Please remove the console from here?

RishabhJain2018 (changes requested):

RishabhJain2018 (approved):

PR #3562, opened 2021-08-02
Remove submission count api call from submissions page

gautamjajoo:
Fixes the following points:

- [x] `submissionCount` in the `allSubmissions` page was giving the count of number of submissions of the participant team not all Submissions.
- [x] We don't require any API call for getting the count of submissions for a particular phase, the count is already there while we fetch the submissions.

@Ram81 @Kajol-Kumari


Kajol-Kumari (approved):
LGTM 👍

RishabhJain2018 (approved):

PR #3568, opened 2021-08-14
Move leaderboard precision value to edit leaderboard section

gautamjajoo:
We can reduce one API call if we add `leaderboard_precision_value` in phaseSplit as well.
@Ram81 @Kajol-Kumari 

Screenshot from 2021-08-14 18-03-49


Ram81 (approved):
LGTM

RishabhJain2018 (approved):

PR #3571, opened 2021-08-17
Add lazy load for images

gautamjajoo:
This PR adds lazy loading on images and the images of the challenges are now loaded when the viewport changes, hence as the user moves down the logos are displayed.
@Ram81 @Kajol-Kumari 

https://user-images.githubusercontent.com/24366008/129788273-47d702e7-1e60-4c92-a251-d894c1b958e5.mp4


gautamjajoo:
@Ram81 We can add this to all the images, but do we need to do that?

Ram81 (approved):
LGTM. Tested locally, works fine

PR #3575, opened 2021-08-20
fix: change scheme of alertmanager to https for staging env

burnerlee:
Currently, the prometheus make a POST request to the alertmanager using HTTP protocol. Using the Reverse proxy, we redirect all the requests made using using HTTP to HTTPS and return a 301 response. However, this changes the request from POST to GET and alerts are not issued. Thus, this fix allows prometheus to ping the alertmanager using the HTTPS Protocol


Ram81 (approved):
LGTM

PR #3588, opened 2021-08-31
Fix #login and #sign up at the top: Updating the font size 

GaelKBertrand:
This is to enhance visuals, expecially for people who have visual impairments difficulties in seeing small characters.


mr-vaibh (changes requested):
It would be much better to use External CSS rather than Inline CSS.
Readability would be much appreciated.

PR #3614, opened 2021-10-13
Display Challenge ID, SQS Queue, IAM Credentials

vinceli1004:


vinceli1004:
@Ram81 it says there's still some changes requested - I think I have addressed everything, just pinging you here so you can verify in case you missed it.

Ram81:
@vinceli1004 can we fix the alignment for the challenge queue name and add more padding between challenge id and queue name. We don't need to have a separate section for these details. These can go in the first section with all other details like challenge title, etc. 

For queue name we should show just first 25 chars and show `...` if queue name is longer with a copy button. You can find an example of the copy button on the profile page -> Get Auth token modal. Let me know if you have any doubts

Ram81:
@vinceli1004 there's a issue with the PR. The copy button is shown next to challenge id. We should show it next to sqs queue name and show full challenge id. We should show first 20 chars of sqs queue name. Another issue I saw is, when I click on the copy button it opens the edit terms and conditions popup. Can we fix these issues?

Screenshot from 2021-12-11 13-32-50

RishabhJain2018 (commented):
Hey @Ram81 , Can we merge this?

PR #3636, opened 2021-12-04
[Monitoring] Set up worker down metrics in alert manager

Ram81:
### Description

This PR - 

- [x] Add alert rules for workers set up on EvalAI AWS using
- [x] Add a metric tag for remote evaluation workers to filter alerts


monitoring/prometheus/rules.yml
```diff
@@ -34,3 +34,15 @@ groups:
     labels:
       severity: major
       group: 'instance'
+
+- name: Worker-Down
+  rules:
+  - alert: WorkerDown
+    expr: (count by (queue_name) (num_submissions_in_queue{is_remote=0}) - count by (queue_name) (num_processed_submissions{is_remote=0}))  >= 5
+    for: 120m
```
RishabhJain2018, line 42: Can we please make it 60m?

RishabhJain2018 (changes requested):

monitoring/prometheus/rules.yml
Ram81: some challenges have 60m evaluation time for submission so I kept this for 120m. Otherwise I'll have to set up alerts for each challenge separately and do it every time a new challenge comes in.

RishabhJain2018 (approved):

RishabhJain2018 (approved):

PR #3696, opened 2022-02-13
Fix  #3695 [FrontendV2] Disable submit button on submit page when a selected file is of incorrect type

### LINKED ISSUE #3695: [FrontendV2] Disable submit button on submit page when a selected file is of incorrect type
### Description

Currently, in our Frontend V2 submit page on the challenge page user can submit a file even if it has a extension which is not allowed by challenge hosts in (i.e. not in `allowed_submission_file_types` for a challenge phase). If a user uploads a file on the frontend which has a invalid extension we should disable the submit button and show a message that selected file type is invalid.

Screenshot for reference:

Screenshot from 2022-02-12 12-05-57

Expected behavior: Disable the submit button when file uploaded using `Upload file` option has a invalid extension.

haldendmello:
disableSubmit


haldendmello:
@Ram81 Can you check this.

PR #3708, opened 2022-03-18
[BugFix] Auth token refresh API

Ram81:
### Description

This PR fixes:

- [x] [Refresh auth token API issue](https://sentry.io/organizations/cloudcv/issues/2988106336/?project=130457&referrer=slack). To fix the issue we have to delete the existing token when a user refreshes the token. SimpleJWT API doesn't support API for refreshing tokens for which `refresh_token` has expired. [Reference](https://github.com/jazzband/djangorestframework-simplejwt/blob/master/rest_framework_simplejwt/tokens.py#L43)

### DIFF
--- a/apps/accounts/views.py
+++ b/apps/accounts/views.py
@@ -1,4 +1,3 @@
-from tokenize import Token
 from django.contrib.auth import logout
 from django.contrib.auth.models import User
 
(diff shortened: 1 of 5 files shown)


RishabhJain2018 (changes requested):
@Ram81 Can we please add on the UI, the expiry date for the token as mentioned by abhishek in one of the conversations.

Ram81:
Here's a screenshot for the UI
Screenshot from 2022-03-27 12-47-41

RishabhJain2018 (approved):
@Ram81 Please fix the travis build.

PR #3709, opened 2022-03-18
[Logging] Add error logs in challenge validation and creation API

Ram81:
### Description

- [x] Add error logs for challenge creation and update API

### DIFF
--- a/apps/challenges/views.py
+++ b/apps/challenges/views.py
@@ -1381,7 +1381,7 @@ def create_challenge_using_zip_file(request, challenge_host_team_pk):
                 challenge.save()
             else:
                 response_data = serializer.errors
-                raise RuntimeError("Challenge validation failed")
+                raise RuntimeError()
                 # transaction.set_rollback(True)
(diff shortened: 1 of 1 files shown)


apps/challenges/views.py
```diff
@@ -3261,6 +3273,9 @@ def create_or_update_github_challenge(request, challenge_host_team_pk):
                         )
                         if serializer.is_valid():
                             serializer.save()
+                        else:
+                            error_messages = serializer.errors
+                            raise RuntimeError("Leaderboard validation failed")
```
RishabhJain2018, line 3278: I am not sure why we want to raise the error here?

Shouldn't we aggregate all the error messages in one check and then display to the user? I think that is why we have `error_messages` variable, no?

RishabhJain2018 (changes requested):

RishabhJain2018 (approved):

RishabhJain2018:
Can you please open an issue for adding/updating tests for this API?

PR #3713, opened 2022-03-29
[BugFix] Auth token expiry and date conversion

Ram81:
### Description

This PR - 

- [x] Fix date conversion on profile page
- [x] Handle exception in token refresh when token is expired 
- [x] Show token expiry on refresh token modal


apps/accounts/models.py
```diff
@@ -64,6 +64,7 @@ class JwtToken(TimeStampedModel):
     user = models.OneToOneField(User, on_delete=models.CASCADE)
     access_token = models.CharField(max_length=512, blank=False, null=True)
     refresh_token = models.CharField(max_length=512, blank=False, null=True)
+    expires_at = models.PositiveIntegerField(default=86400)
```
RishabhJain2018, line 67: Shouldn't it be a date field?

RishabhJain2018: Also, why do we need this field? Can't we calculate it dynamically?

RishabhJain2018 (changes requested):
Please fix the travis build as well.

apps/accounts/models.py
Ram81: Because it is epoch time and not a date time field. Replied to second comment on channel

RishabhJain2018 (approved):

PR #3717, opened 2022-04-09
remove borders from settings tab cards

gautamjajoo:
@Ram81 

Screenshot 2022-04-09 at 6 20 24 PM


Ram81 (changes requested):
@gautamjajoo can you share screenshots of other tabs on settings page as well?

PR #3720, opened 2022-04-23
[Worker] Upgrade python version to 3.8

Ram81:
### Description

- [x] Upgrade python version of submission worker to 3.8.
- [x] Remove django container from submission worker docker dependency.


RishabhJain2018 (changes requested):
Let's create another worker version as this will create more problems for you with the existing worker containers.

RishabhJain2018 (approved):

PR #3730, opened 2022-05-03
[BugFix] Fix hosted challenge page filtering to show private challenges

Ram81:
### Description

Currently, hosted challenge page doesn't show private challenges on the tab due to a bug in the API 

This PR - 

- [x] Add support to fetch all challenges with `all, public or private` visibility and `all, approved and unapproved` challenges in `get_all_challenges` API
- [x] Fixes bug in hosted challenge page filtering to show private challenges

### DIFF
--- a/apps/challenges/urls.py
+++ b/apps/challenges/urls.py
@@ -40,7 +40,7 @@ urlpatterns = [
     ),
     # `A-Za-z` because it accepts either of `all, future, past or present` in either case
     url(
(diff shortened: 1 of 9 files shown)


RishabhJain2018 (changes requested):
Can you please fix the travis build and update the description of the PR so as to describe what is the issue?

RishabhJain2018 (approved):

PR #3751, opened 2022-05-29
[Worker] Update numpy version in submission worker

Ram81:
### Description

This PR - 

- [x] Updates numpy version in submission worker. Some new challenges need latest version of numpy and they have been facing issues with upgrading it using the `__init__.py`

### DIFF
--- a/requirements/worker.txt
+++ b/requirements/worker.txt
@@ -1,5 +1,6 @@
 matplotlib==3.3.3
 networkx==2.5
+numpy==1.21.0
 pandas==1.1.4
 pycocotools==2.0.2
 scipy==1.5.4


requirements/worker.txt
```diff
@@ -1,5 +1,6 @@
 matplotlib==3.3.3
 networkx==2.5
+numpy==1.21.0
```
RishabhJain2018, line 3: Do you know what is the error when upgrading it through `__init__.py`

RishabhJain2018 (changes requested):

requirements/worker.txt
Ram81: Due to existing numpy version the module reload is not automatically working and the force `reload` call fails with this error https://cloudcv.slack.com/archives/C2P84STC6/p1653838354324559

RishabhJain2018 (approved):

PR #3755, opened 2022-06-03
[Cron] Delete workers for all inactive challenges

Ram81:
### Description

This PR - 

- [x] Delete workers for all inactive challenges and not just unapproved challenges.

### DIFF
--- a/scripts/monitoring/auto_stop_workers.py
+++ b/scripts/monitoring/auto_stop_workers.py
@@ -49,7 +49,7 @@ def delete_worker(challenge_id):
 
 
 def get_challenges():
-    all_challenge_endpoint = "{}/api/challenges/challenge/all/unapproved/all".format(
+    all_challenge_endpoint = "{}/api/challenges/challenge/all/all/all".format(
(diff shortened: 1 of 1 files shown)


scripts/monitoring/auto_stop_workers.py
```diff
@@ -49,7 +49,7 @@ def delete_worker(challenge_id):
 
 
 def get_challenges():
-    all_challenge_endpoint = "{}/api/challenges/challenge/all/unapproved/all".format(
+    all_challenge_endpoint = "{}/api/challenges/challenge/all/all/all".format(
```
RishabhJain2018, line 52: @Ram81 Is this endpoint correct?

RishabhJain2018 (changes requested):

scripts/monitoring/auto_stop_workers.py
Ram81: Yes, it is

RishabhJain2018 (changes requested):

PR #3771, opened 2022-07-19
Add worker auto scaling script

gchhablani:
This PR attempts to add a script to auto-scale workers in AWS to prevent unnecessary usage.

### DIFF
--- a/scripts/monitoring/auto_scale_workers.py
+++ b/scripts/monitoring/auto_scale_workers.py
@@ -63,9 +63,11 @@ def get_queue_length(queue):
     # TODO: Check if we should use approximate number of messages
     # https://github.com/boto/boto3/issues/599
     # Ref: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/sqs.html#using-an-existing-queue
-    count = 0
-    for message in queue.receive_messages():
-        count += 1
+    # Ref: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sqs.html#SQS.Queue.attributes
+    return (
+        queue.attributes["ApproximateNumberOfMessages"]
+        + queue.attributes["ApproximateNumberOfMessages"]
+    )
 
 
 def get_queue_length_by_challenge(challenge):
--- a/scripts/monitoring/auto_scale_workers.py
+++ b/scripts/monitoring/auto_scale_workers.py
@@ -10,17 +10,10 @@ import boto3
 from django.conf import settings
 import botocore
 
-# from botocore.exceptions import ClientError
-# from http import HTTPStatus
+# from auto_stop_workers import start_worker, stop_worker
 
 utc = pytz.UTC
 
-# TODO: Check if need some kind of logging/logger here.
-#
-# TO CODE
-# Fetch all challenges from EvalAI
-# BOTO3 to get number of pending submissions in the queue.
-# SQS to set the number of workers to 1 or 0.
 
 # Eval AI related credentials
 evalai_endpoint = os.environ.get("API_HOST_URL")
@@ -52,24 +45,16 @@ def get_sqs_queue_by_name(queue_name):
     Returns:
         Returns the SQS Queue object
     """
-    if settings.DEBUG or settings.TEST:
-        sqs = boto3.resource(
-            "sqs",
-            endpoint_url=os.environ.get("AWS_SQS_ENDPOINT", "http://sqs:9324"),
-            region_name=os.environ.get("AWS_DEFAULT_REGION", "us-east-1"),
-            aws_secret_access_key=os.environ.get("AWS_SECRET_ACCESS_KEY"),
-            aws_access_key_id=os.environ.get("AWS_ACCESS_KEY_ID"),
-        )
-    else:
-        sqs = boto3.resource(
-            "sqs",
-            region_name=os.environ.get("AWS_DEFAULT_REGION", "us-east-1"),
-            aws_secret_access_key=os.environ.get("AWS_SECRET_ACCESS_KEY"),
-            aws_access_key_id=os.environ.get("AWS_ACCESS_KEY_ID"),
-        )
+
+    sqs = boto3.resource(
+        "sqs",
+        region_name=os.environ.get("AWS_DEFAULT_REGION", "us-east-1"),
+        aws_secret_access_key=os.environ.get("AWS_SECRET_ACCESS_KEY"),
+        aws_access_key_id=os.environ.get("AWS_ACCESS_KEY_ID"),
+    )
     if queue_name == "":
-        queue_name = "evalai_submission_queue"
-    # Check if the queue exists. If no, then create one
+        raise ValueError("Queue Name cannot be empty")
+
     try:
         queue = sqs.get_queue_by_name(QueueName=queue_name)
     except botocore.exceptions.ClientError as ex:
@@ -106,12 +91,6 @@ def get_queue_name(param, challenge_pk):
     return queue_name
 
 
-def get_queue_name_by_challenge(challenge):
-    challenge_id = challenge["id"]
-    challenge_title = challenge["title"]
-    return get_queue_name(challenge_id, challenge_title)
-
-
 def get_queue_length(queue):
     # TODO: Check if we should use approximate number of messages
     # https://github.com/boto/boto3/issues/599
@@ -122,7 +101,7 @@ def get_queue_length(queue):
 
 
 def get_queue_length_by_challenge(challenge):
-    queue_name = get_queue_name_by_challenge(challenge)
+    queue_name = challenge["queue"]
     queue = get_sqs_queue_by_name(queue_name)
     return get_queue_length(queue)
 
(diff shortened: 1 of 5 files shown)


scripts/monitoring/auto_scale_workers.py
```diff
@@ -0,0 +1,256 @@
...
+
+
+# TODO: Check if we can use challenge["queue"] directly to get queue name.
+# TODO: Understand if all the attributes are exposed in the API calls, or are they done selectively.
+def get_sqs_queue_by_name(queue_name):
+    """
+    Returns:
+        Returns the SQS Queue object
+    """
+    if settings.DEBUG or settings.TEST:
```
Ram81, line 55: we don't need to handle this case. I think these checks are from our `aws_utils.py` method. When we will run the script it will be running independent of django server. So this method can just fetch the boto3 client for sqs.

scripts/monitoring/auto_scale_workers.py
```diff
@@ -0,0 +1,256 @@
...
+        )
+    else:
+        sqs = boto3.resource(
+            "sqs",
+            region_name=os.environ.get("AWS_DEFAULT_REGION", "us-east-1"),
+            aws_secret_access_key=os.environ.get("AWS_SECRET_ACCESS_KEY"),
+            aws_access_key_id=os.environ.get("AWS_ACCESS_KEY_ID"),
+        )
+    if queue_name == "":
+        queue_name = "evalai_submission_queue"
```
Ram81, line 71: might be a good idea to throw a exception here instead of using dummy queue name

scripts/monitoring/auto_scale_workers.py
```diff
@@ -0,0 +1,256 @@
...
+    except botocore.exceptions.ClientError as ex:
+        if (
+            ex.response["Error"]["Code"]
+            != "AWS.SimpleQueueService.NonExistentQueue"
+        ):
+            raise Exception("Cannot get queue: {}".format(queue_name))
+    return queue
+
+
+def get_queue_name(param, challenge_pk):
```
Ram81, line 84: @gchhablani you don't need to rewrite this method. The SQS queue name is already stored in challenge model. If it's not exposed directly in the challenge list API you can add it in the serializer instead and fetch it from the API. Having this method here means if we update queue name generation logic at any point we have to do it in the script explicitly.

scripts/monitoring/auto_scale_workers.py
```diff
@@ -0,0 +1,256 @@
...
+    queue_name = param.replace(" ", "-").lower()[:max_challenge_title_len]
+    queue_name = re.sub(r"\W+", "-", queue_name)
+
+    queue_name = "{}-{}-{}-{}".format(
+        queue_name, challenge_pk, env, uuid.uuid4()
+    )[:max_len]
+    return queue_name
+
+
+def get_queue_name_by_challenge(challenge):
```
Ram81, line 109: This method won't be needed

scripts/monitoring/auto_scale_workers.py
```diff
@@ -0,0 +1,256 @@
...
+    challenge_title = challenge["title"]
+    return get_queue_name(challenge_id, challenge_title)
+
+
+def get_queue_length(queue):
+    # TODO: Check if we should use approximate number of messages
+    # https://github.com/boto/boto3/issues/599
+    # Ref: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/sqs.html#using-an-existing-queue
+    count = 0
+    for message in queue.receive_messages():
```
Ram81, line 120: isn't there a `count()` API in boo3?

scripts/monitoring/auto_scale_workers.py
```diff
@@ -0,0 +1,256 @@
...
+
+def get_aws_credentials_for_challenge(challenge):
+    """
+    Return the AWS credentials for a challenge using challenge pk
+    Arguments:
+        challenge {dict} -- challenge json for which credentials are to be fetched
+    Returns:
+        aws_key {dict} -- Dict containing aws keys for a challenge
+    """
+    if challenge["use_host_credentials"]:
```
Ram81, line 138: we don't need to handle this case. For this iteration you can only do auto scaling for workers that are running at our end i.e.e ` challenge["use_host_credentials"]` is either none or false

scripts/monitoring/auto_scale_workers.py
```diff
@@ -0,0 +1,256 @@
...
+            region_name=aws_keys["AWS_REGION"],
+            aws_access_key_id=aws_keys["AWS_ACCESS_KEY_ID"],
+            aws_secret_access_key=aws_keys["AWS_SECRET_ACCESS_KEY"],
+        )
+        return client
+    except Exception as e:
+        raise e
+
+
+# def update_service_by_challenge_pk(
```
Ram81, line 178: can you remove this?

scripts/monitoring/auto_scale_workers.py
```diff
@@ -0,0 +1,256 @@
...
+#     except ClientError as e:
+#         raise e
+
+
+def increase_or_decrease_workers(challenge):
+    # challenge_aws_keys = get_aws_credentials_for_challenge(challenge)
+    queue_length = get_queue_length_by_challenge(challenge)
+    if queue_length > 0:
+        # set instances to 1
+        # print statement
```
Ram81, line 227: Here you just need to make EvalAI API call

scripts/monitoring/auto_scale_workers.py
```diff
@@ -0,0 +1,256 @@
...
+        pass
+    else:
+        # set instances to 0
+        # print statement
```
Ram81, line 231: same here

Ram81 (changes requested):

scripts/monitoring/auto_scale_workers.py
gchhablani: Fixed.

scripts/monitoring/auto_scale_workers.py
gchhablani: Fixed.

scripts/monitoring/auto_scale_workers.py
gchhablani: Currently, the serializer does not contain `queue` in the fields. I will add the same and check if it is shown in the response.

Makes sense about needing to change the script if the queue name generation logic.

scripts/monitoring/auto_scale_workers.py
gchhablani: The methods listed here do not contain a `count()` method: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sqs.html
I also checked some stackoverflow answers regarding this. Will look up once again.

scripts/monitoring/auto_scale_workers.py
```diff
@@ -0,0 +1,136 @@
...
+    return response.json()
+
+
+def get_challenges():
+    all_challenge_endpoint = "{}/api/challenges/challenge/all/all/all".format(
+        evalai_endpoint  # Gets all challenges
+    )
+    response = execute_get_request(all_challenge_endpoint)
+
+    return response.json()
```
Ram81, line 33: remove `.json` from here as you are already doing it on line 24

scripts/monitoring/auto_scale_workers.py
```diff
@@ -0,0 +1,136 @@
...
+    # Ref: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/sqs.html#using-an-existing-queue
+    # Ref: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sqs.html#SQS.Queue.attributes
+    return (
+        queue.attributes["ApproximateNumberOfMessages"]
+        + queue.attributes["ApproximateNumberOfMessages"]
+    )
+
+
+def get_queue_length_by_challenge(challenge):
+    queue_name = challenge["queue"]
```
Ram81, line 74: Add `queue` in challenge serializer. The API response doesn't include queue name right now

scripts/monitoring/auto_scale_workers.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+def get_queue_length(queue):
+    # TODO: Check if we should use approximate number of messages
+    # https://github.com/boto/boto3/issues/599
+    # Ref: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/sqs.html#using-an-existing-queue
+    # Ref: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sqs.html#SQS.Queue.attributes
+    return (
+        queue.attributes["ApproximateNumberOfMessages"]
+        + queue.attributes["ApproximateNumberOfMessages"]
```
Ram81, line 69: you are adding exact same values here? This is a typo I guess. Can you please fix this?

scripts/monitoring/auto_scale_workers.py
```diff
@@ -0,0 +1,136 @@
...
+        raise e
+
+
+def increase_or_decrease_workers(challenge):
+    queue_length = get_queue_length_by_challenge(challenge)
+    if queue_length > 0:
+        # start worker
+        start_worker(challenge["id"])
+        print("Started worker for challenge: {}".format(challenge["id"]))
+        pass
```
Ram81, line 106: why are there pass commands here? please remove them

scripts/monitoring/auto_scale_workers.py
```diff
@@ -0,0 +1,136 @@
...
+    else:
+        # stop worker
+        stop_worker(challenge["id"])
+        print("Stopped worker for challenge: {}".format(challenge["id"]))
+        pass
```
Ram81, line 111: please remove this

scripts/monitoring/auto_scale_workers.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+# TODO: Factor in limits for the APIs
+def increase_or_decrease_workers_for_challenges(response):
+    for challenge in response["results"]:
+        if not challenge["use_host_credentials"]:
```
Ram81, line 117: Can you add another check here - also check if `remote_evaluation` is not set to `True` for a challenge. We only want to run auto-scaling for workers at our end.

Ram81 (changes requested):
Nicely done! Some minor comments

PR #3776, opened 2022-07-26
store email address in lowercase

gautamjajoo:
Store lowercase email ids when a user signs up on EvalAI

@Ram81


frontend_v2/src/app/components/auth/signup/signup.component.ts
```diff
@@ -62,10 +62,11 @@ export class SignupComponent implements OnInit, AfterViewInit {
   // Function to signup
   userSignUp(signupFormValid) {
     if (signupFormValid) {
+      let email = this.authService.regUser['email'];
```
savish28, line 65: NIT: why create a variable if used only once?

Ram81 (approved):

PR #3805, opened 2022-09-19
Increase CPU cores and memory for workers

gchhablani:
This PR increases default number of CPU cores and memory for the workers.


Ram81 (approved):
LGTM

PR #3811, opened 2022-09-29
Check actions for past challenges in manage worker

gchhablani:
This PR checks for the end date in manager worker API in order to allow only delete action for the past challenges.

### DIFF
--- a/apps/challenges/views.py
+++ b/apps/challenges/views.py
@@ -1,5 +1,4 @@
 import csv
-from datetime import datetime
 import json
 import logging
 import os
@@ -13,6 +12,7 @@ import yaml
 import zipfile
 
 from os.path import basename, isfile, join
+from dateutil.parser import parse
 
 from django.conf import settings
 from django.contrib.auth.hashers import make_password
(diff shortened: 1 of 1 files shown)


apps/challenges/views.py
```diff
@@ -2997,6 +2998,16 @@ def manage_worker(request, challenge_pk, action):
 
     challenge = get_challenge_model(challenge_pk)
 
+    if datetime.fromisoformat(
```
Ram81, line 3001: Can you instead use the datetime parser? See this https://dateutil.readthedocs.io/en/stable/parser.html

apps/challenges/views.py
```diff
@@ -1,4 +1,5 @@
 import csv
+from datetime import datetime
```
Ram81, line 2: Please follow the correct import order. See the codebase for reference. All imports starting with `from ` appear at the end of `import` imports

Ram81 (changes requested):
Left minor comments. Can you fix the failing build?

apps/challenges/views.py
```diff
@@ -2997,6 +3000,14 @@ def manage_worker(request, challenge_pk, action):
 
     challenge = get_challenge_model(challenge_pk)
 
+    if parse(challenge["end_date"]) < datetime.utcnow() and action in ("start", "stop", "restart"):
+        response_data = {
+            "error": "The action {} is invalid for worker as the challenge ended on {}.".format(
+                action, challenge["end_date"]
```
Ram81, line 3006: Change the error message to `Action {} worker is not supported for an inactive challenge`

Ram81 (changes requested):

Ram81 (approved):

PR #3812, opened 2022-09-29
Change URL for get all challenges endpoint

gchhablani:
This PR changes the URL for `get_all_challenges` endpoint to make it more user-friendly/understandable.


Ram81 (changes requested):
@gchhablani can you fix the failing tests?

apps/challenges/urls.py
```diff
@@ -40,7 +40,8 @@
     ),
     # `A-Za-z` because it accepts either of `all, future, past or present` in either case
     url(
-        r"^challenge/(?P<challenge_time>[A-Za-z]+)/(?P<challenge_approved>[A-Za-z]+)/(?P<challenge_published>[A-Za-z]+)$",
+        r"^challenge/time/(?P<challenge_time>[A-Za-z]+)/approved/"
+        "(?P<challenge_approved>[A-Za-z]+)/published/(?P<challenge_published>[A-Za-z]+)$",
```
Ram81, line 44: @gchhablani did you test EvalAI UI after this change? This will also need updating URLs in the frontend `challengeCtrl.js`. Changing this API must've broke the challenge dashboard page. Can you verify?

Ram81 (changes requested):

frontend/tests/controllers-test/challengeListCtrl.test.js
```diff
@@ -143,7 +143,7 @@ describe('Unit tests for challenge list controller', function () {
             expect(vm.noneUpcomingChallenge).toBeTruthy();
         });
 
-        it('check description length and calculate timezone of upcoming challenge `challenges/challenge/future`', function () {
+        it('check description length and calculate timezone of upcoming challenge `challenges/challenge/time/future`', function () {
```
gchhablani, line 146: @Ram81 I am confused about this. Should this be changed?

apps/challenges/urls.py
```diff
@@ -40,7 +40,8 @@
...
         views.get_all_challenges,
```
Ram81, line 45: @gchhablani can we change `published` in the URL to `visibility` instead? `published` is a confusing keyword in URL when a challenge is public/private

Ram81 (changes requested):
Can you also resolve the merge conflicts?

Ram81:
Any chance we can make the API call from old cli versions backward compatible with new URL we are moving to for this API ?

apps/challenges/urls.py
```diff
@@ -40,7 +40,8 @@
     ),
     # `A-Za-z` because it accepts either of `all, future, past or present` in either case
     url(
-        r"^challenge/(?P<challenge_time>[A-Za-z]+)/(?P<challenge_approved>[A-Za-z]+)/(?P<challenge_published>[A-Za-z]+)$",
+        r"^challenge/time/(?P<challenge_time>[A-Za-z]+)/approved/"
+        "(?P<challenge_approved>[A-Za-z]+)/visibility/(?P<challenge_published>[A-Za-z]+)$",
```
Ram81, line 44: Sorry for the back and forth. Looking at this API it isn't really readable. Can we instead convert it to a POST and GET API which has [this URL](https://github.com/Cloud-CV/EvalAI/commit/94f844952c4dba76c3cbe2b6929a978d7fc746d4#diff-7f035da148c99adb9058e04048f2cdb0e702f5eaa05c56f6a17adbbef702e4eeL43). 

And we will pass `is_approved` and `is_published` as request data for our cron jobs. Switching back to the old API URL will fix the existing breaking CLI and we just need to modify the cron job API calls and handle params from request data

frontend/src/js/controllers/challengeListCtrl.js
```diff
@@ -32,7 +32,7 @@
         } else {
             parameters.token = null;
         }
-        parameters.url = 'challenges/challenge/present/approved/public';
+        parameters.url = 'challenges/challenge/time/present/approved/approved/published/public';
```
Ram81, line 35: This URL will be changed to `challenges/challenge/present`

apps/challenges/urls.py
```diff
@@ -40,7 +40,8 @@
...
         views.get_all_challenges,
         name="get_all_challenges",
```
Ram81, line 46: The view for this API also needs to be changed to only return approved and present challenges by default. If the request params have other values for these flags then we will use that.

Ram81 (changes requested):
Let me know if you any doubts

PR #3820, opened 2022-10-11
Allow deletion of submission meta attributes

gchhablani:
This PR adds a fix to allow deletion of submission meta attributes and default submission meta attributes. 

Fixes two issues:
1. Default submission meta attributes were getting forcefully displayed even when some keys attributes are removed.
2. Default and other submission meta attributes were not getting updated when the key itself is removed.


Ram81 (approved):
LGTM

PR #3844, opened 2022-12-13
[Feature] Admin task to delete challenge evaluation cluster created (Part 1)

matthew-so:
This commit adds a new Django admin task to specifically delete a challenge's evaluation clusters without deleting its roles, subnets, etc...

### DIFF
--- a/apps/challenges/admin.py
+++ b/apps/challenges/admin.py
@@ -11,6 +11,7 @@ from .aws_utils import (
     scale_workers,
     start_workers,
     stop_workers,
+    delete_only_challenge_evaluation_cluster,
 )
 
 from .admin_filters import ChallengeFilter
@@ -83,6 +84,7 @@ class ChallengeAdmin(ImportExportTimeStampedAdmin):
         "scale_selected_workers",
         "restart_selected_workers",
         "delete_selected_workers",
+        "delete_only_selected_challenge_evaluation_clusters"
     ]
     action_form = UpdateNumOfWorkersForm
 
@@ -211,6 +213,30 @@ class ChallengeAdmin(ImportExportTimeStampedAdmin):
 
     delete_selected_workers.short_description = (
         "Delete all selected challenge workers."
+    )
+
+    def delete_only_selected_challenge_evaluation_clusters(self, request, queryset):
+        """Deletes the selected challenge evaluation clusters only; roles, subnets, etc... not deleted"""
+        response = delete_only_challenge_evaluation_cluster(queryset)
+        count, failures = response["count"], response["failures"]
+
+        if count == queryset.count():
+            message = "All selected challenges' evaluation clusters successfully deleted."
+            messages.success(request, message)
+        else:
+            messages.success(
+                request,
+                "{} challenges' evaluation clusters were successfully deleted.".format(count),
+            )
+            for fail in failures:
+                challenge_pk, message = fail["challenge_pk"], fail["message"]
+                display_message = "Challenge {}: {}".format(
+                    challenge_pk, message
+                )
+                messages.error(request, display_message)
+
+    delete_only_selected_challenge_evaluation_clusters.description = (
+        "Delete only all selected challenges' evaluation clusters."
     )
 
 
--- a/apps/challenges/aws_utils.py
+++ b/apps/challenges/aws_utils.py
@@ -900,6 +900,132 @@ def create_eks_nodegroup(challenge, cluster_name):
 
 
 @app.task
+def delete_only_challenge_evaluation_cluster(queryset):
+    """
+    The function called by the admin action method to delete all the challenge evaluation clusters used by the
+    selected challenges. This does not delete the roles, subnets, route tables, security groups, etc... associated with
+    these clusters.
+
+    Parameters:
+    queryset (<class 'django.db.models.query.QuerySet'>): The queryset of selected challenges in the django admin page.
+
+    Returns:
+    dict: keys-> 'count': the number of workers successfully stopped.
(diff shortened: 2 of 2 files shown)


apps/challenges/admin.py
```diff
@@ -213,6 +215,30 @@ def delete_selected_workers(self, request, queryset):
         "Delete all selected challenge workers."
     )
 
+    def delete_only_selected_challenge_evaluation_clusters(self, request, queryset):
```
Ram81, line 218: rename method to `delete_challenge_evaluation_cluster`

apps/challenges/admin.py
```diff
@@ -213,6 +215,30 @@ def delete_selected_workers(self, request, queryset):
...
+            )
+            for fail in failures:
+                challenge_pk, message = fail["challenge_pk"], fail["message"]
+                display_message = "Challenge {}: {}".format(
+                    challenge_pk, message
+                )
+                messages.error(request, display_message)
+
+    delete_only_selected_challenge_evaluation_clusters.description = (
+        "Delete only all selected challenges' evaluation clusters."
```
Ram81, line 239: Change action name to `Delete challenge evaluation cluster`

apps/challenges/aws_utils.py
```diff
@@ -899,6 +899,134 @@ def create_eks_nodegroup(challenge, cluster_name):
     create_service_by_challenge_pk(client, challenge_obj, client_token)
 
 
+@app.task
+def delete_only_challenge_evaluation_cluster(queryset):
```
Ram81, line 903: Rename to `delete_code_upload_challenge_resources`

apps/challenges/aws_utils.py
```diff
@@ -899,6 +899,134 @@ def create_eks_nodegroup(challenge, cluster_name):
...
+    for challenge in queryset:
+        try:
+            ChallengeEvaluationCluster.objects.get(challenge=challenge)
+        except ChallengeEvaluationCluster.DoesNotExist:
+            response = "Please select challenges with active evaluation clusters only."
+            failures.append(
+                {"message": response, "challenge_pk": challenge.pk}
+            )
+            continue
+        response = delete_eks_cluster(challenge=challenge)
```
Ram81, line 940: Isn't this a celery task? This should be `delete_eks_cluster.delay` right? To ensure it is called as a async task?

apps/challenges/aws_utils.py
```diff
@@ -899,6 +899,134 @@ def create_eks_nodegroup(challenge, cluster_name):
...
+        challenge_obj.title.replace(" ", "-"), environment_suffix
+    )
+    if challenge_obj.approved_by_admin and challenge_obj.is_docker_based:
+        challenge_aws_keys = get_aws_credentials_for_challenge(
+            challenge_obj.pk
+        )
+        client = get_boto3_client("eks", challenge_aws_keys)
+        try:
+            # Delete nodegroup
+            delete_eks_nodegroup(challenge, cluster_name)
```
Ram81, line 979: same here. Isn't this a celery task? Can you verify the call for celery tasks in other methods for creating the infra for code upload challenge

Ram81 (changes requested):

apps/challenges/aws_utils.py
matthew-so: Right. Fixed

apps/challenges/aws_utils.py
```diff
@@ -1004,6 +1004,133 @@ def create_eks_nodegroup(challenge, cluster_name):
     create_service_by_challenge_pk(client, challenge_obj, client_token)
 
 
+def delete_multiple_challenge_evaluation_clusters(queryset):
```
Ram81, line 1007: Can we rename this method to `delete_challenge_evaluation_clusters`

apps/challenges/admin.py
```diff
@@ -213,6 +215,30 @@ def delete_selected_workers(self, request, queryset):
         "Delete all selected challenge workers."
     )
 
+    def delete_challenge_evaluation_clusters(self, request, queryset):
```
Ram81, line 218: can we rename this method to `delete_selected_challenge_evaluation_clusters`

apps/challenges/aws_utils.py
```diff
@@ -1004,6 +1004,133 @@ def create_eks_nodegroup(challenge, cluster_name):
...
+    failures = []
+    for challenge in queryset:
+        try:
+            ChallengeEvaluationCluster.objects.get(challenge=challenge)
+        except ChallengeEvaluationCluster.DoesNotExist:
+            response = "Please select challenges with active evaluation clusters only."
+            failures.append(
+                {"message": response, "challenge_pk": challenge.pk}
+            )
+            continue
```
Ram81, line 1043: Can you instead check if `challenge.workers` count is null for this check. The `ChallengeEvaluationCluster` object will exist for each code upload challenge so this check will never fail

Ram81 (changes requested):

apps/challenges/aws_utils.py
matthew-so: Okay. Should we not delete a challenge's ChallengeEvaluationCluster when its cluster is successfully deleted in that case?

matthew-so: I noticed that we do check for a ChallengeEvaluationCluster's existence at times. https://github.com/Cloud-CV/EvalAI/blob/master/apps/challenges/views.py#L2835
Furthermore, evaluation clusters are only created for challenges that are docker based. Let's keep the response in there

matthew-so: Done

PR #3845, opened 2022-12-13
[Feature] Admin task to delete challenge evaluation cluster created (Part 2)

matthew-so:
This pull request is part 2 of the cluster deletion API being created. Part 1 can be found [here](https://github.com/Cloud-CV/EvalAI/pull/3844). This commit adds a new Django admin task to fully delete a challenge's evaluation clusters, including its roles, subnets, etc...

### DIFF
--- a/apps/challenges/admin.py
+++ b/apps/challenges/admin.py
@@ -11,6 +11,7 @@ from .aws_utils import (
     scale_workers,
     start_workers,
     stop_workers,
+    delete_only_challenge_evaluation_cluster,
 )
 
 from .admin_filters import ChallengeFilter
@@ -83,6 +84,7 @@ class ChallengeAdmin(ImportExportTimeStampedAdmin):
         "scale_selected_workers",
         "restart_selected_workers",
         "delete_selected_workers",
+        "delete_only_selected_challenge_evaluation_clusters"
     ]
     action_form = UpdateNumOfWorkersForm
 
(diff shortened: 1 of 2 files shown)


apps/challenges/admin.py
```diff
@@ -213,6 +217,54 @@ def delete_selected_workers(self, request, queryset):
...
+            messages.success(
+                request,
+                "{} challenges' evaluation clusters were successfully deleted.".format(count),
+            )
+            for fail in failures:
+                challenge_pk, message = fail["challenge_pk"], fail["message"]
+                display_message = "Challenge {}: {}".format(
+                    challenge_pk, message
+                )
+                messages.error(request, display_message)
```
Ram81, line 262: Like discussed today, let's merge these two tasks

apps/challenges/aws_utils.py
```diff
@@ -899,6 +899,339 @@ def create_eks_nodegroup(challenge, cluster_name):
...
+    challenge_aws_keys = get_aws_credentials_for_challenge(challenge_obj.pk)
+    client = get_boto3_client("iam", challenge_aws_keys)
+    environment_suffix = "{}-{}".format(challenge_obj.pk, settings.ENVIRONMENT)
+    eks_role_name = "evalai-code-upload-eks-role-{}".format(environment_suffix)
+    node_group_role_name = "evalai-code-upload-nodegroup-role-{}".format(
+        environment_suffix
+    )
+    try:
+        # Delete all EKS cluster subnets
+        delete_eks_cluster_subnets.delay(challenge)
```
Ram81, line 974: @matthew-so subnets need to be deleted at the very end. It won't work otherwise. The flow for deletion will be:

1. Delete EKS nodegroup
2. Delete EKS cluster
3. Delete EFS associated to challenge
4. Delete policies and IAM roles for the challenge
5. Delete subnets
6. Delete VPC

Can you restructure the calls based on this flow?

Ram81 (changes requested):

apps/challenges/admin.py
matthew-so: Will do!

apps/challenges/aws_utils.py
matthew-so: Will do

PR #3849, opened 2022-12-15
Fix deletion of default submission meta attributes in config

gchhablani:
This PR fixes an error in the support of deletion of default submission meta attributes through challenge configuration.


Ram81 (approved):
LGTM

PR #3851, opened 2022-12-20
[BugFix] Fix worker scaling management

matthew-so:
This pull request fixes the sole remaining issue(s) with worker scaling management through the EvalAI UI. To date, these are:

1. The UI not properly displaying the current worker configuration


apps/challenges/aws_utils.py
```diff
@@ -664,6 +664,7 @@ def scale_resources(challenge, worker_cpu_cores, worker_memory):
             challenge.task_def_arn = None
             challenge.save()
         else:
+            response["Error"] = "Scaling worker not supported for inactive worker. Full error:\n" + response["Error"]
```
Ram81, line 667: Change message to `Scaling inactive workers not supported`

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -1639,7 +1639,7 @@
             };
             parameters.callback = {
                 onSuccess: function() {
-                    $rootScope.notify("success", "Evaluation worker resources scaled successfully");
+                    $rootScope.notify("success", "Evaluation worker resources scaled successfully. Please wait until the old worker is stopped and the new worker starts before making further changes.");
```
Ram81, line 1642: Change message to `Evaluation worker resources scaled successfully.`

Ram81 (changes requested):

apps/challenges/aws_utils.py
matthew-so: Done

frontend/src/js/controllers/challengeCtrl.js
matthew-so: Done

Ram81 (approved):

PR #3853, opened 2022-12-23
[BugFix] Leaderboard Watermark

matthew-so:
The leaderboard watermark has been corrected so that it appears on leaderboards.
Screenshot 2022-12-23 at 00-33-49 EvalAI Evaluating state of the art in AI


frontend/src/views/web/challenge/leaderboard.html
```diff
@@ -76,7 +76,7 @@ <h5 class="w-300">Leaderboard</h5>
                         </div>
                     </div>
                 </div>
-                <table ng-if="challenge.isResult && challenge.leaderboard.length" ng-attr-id="leaderboard_table" ng-class="['highlight', {'watermarked': challenge.showBackground}]">
+                <table ng-if="challenge.isResult && challenge.leaderboard.length" ng-attr-id="leaderboard_table" ng-class="['highlight', 'watermarked']">
```
Ram81, line 79: @matthew-so this will enable it for both prod and staging right? We only want this change on staging for now. Can you please fix the check?

Ram81 (changes requested):

frontend/src/views/web/challenge/leaderboard.html
matthew-so: Right. I'll add showBackground back to challengeCtrl

matthew-so: Fixed

Ram81 (approved):

PR #3858, opened 2022-12-30
[BugFix] Fixed remaining worker scaling bugs

matthew-so:
The following improvements were made:

1. Summaries of error/success messages (no changes made to worker configuration, workers inactive, scaling successful, etc...) now appear in the UI when scaling is attempted.
2. The UI now correctly defaults to showing the worker's existing configuration.
3. Error/success message phrasing has been improved.


apps/challenges/views.py
```diff
@@ -3016,6 +3016,16 @@ def scale_resources_by_challenge_pk(request, challenge_pk):
         return Response(response_data, status=status.HTTP_400_BAD_REQUEST)
 
     challenge = get_challenge_model(challenge_pk)
+    response_data = {
+        "error": "Issue with ECS."
+    }
+    return Response(response_data, status=status.HTTP_503_SERVICE_UNAVAILABLE)
```
Ram81, line 3022: You are calling return here so the code after this will never get executed. Is this a indentation bug?

apps/challenges/aws_utils.py
```diff
@@ -638,9 +638,8 @@ def scale_resources(challenge, worker_cpu_cores, worker_memory):
     client = get_boto3_client("ecs", aws_keys)
 
     if challenge.worker_cpu_cores == worker_cpu_cores and challenge.worker_memory == worker_memory:
-        message = "Worker cores and memory were not modified."
         return {
-            "Error": message,
+            "IsWorkerConfigModified": True,
```
Ram81, line 642: Can you instead send `"Success": True, "message" : "Worker scaled successfully!"`

apps/challenges/aws_utils.py
```diff
@@ -663,9 +662,6 @@ def scale_resources(challenge, worker_cpu_cores, worker_memory):
         ):
             challenge.task_def_arn = None
             challenge.save()
-        else:
-            response["Error"] = "Scaling inactive workers not supported"
-            return response
     except ClientError as e:
         logger.exception(e)
         return e.response
```
Ram81, line 667: Sam here send a dict with `"error": True, "message": "Scaling inactive workers not supported"`

apps/challenges/views.py
```diff
@@ -3026,16 +3036,28 @@ def scale_resources_by_challenge_pk(request, challenge_pk):
...
         response = scale_resources(challenge, worker_cpu_cores, worker_memory)
         if response["ResponseMetadata"]["HTTPStatusCode"] != 200:
-            response_data = {
-                "error": "Issue with ECS."
-            }
+            if response.get('Error', {'Message': 'No error', 'Code': 'No error'}).get('Code', 'No error code') == \
+                    'ClientException':
+                response_data = {
+                    "error": "Challenge workers are inactive or do not exist."
+                }
```
Ram81, line 3043: This case won't occur because you are checking `if worker is None` on L3023. We can remove this error response and just send a error message saying `"error": "Scaling worker resources failed!"`

apps/challenges/views.py
```diff
@@ -3026,16 +3036,28 @@ def scale_resources_by_challenge_pk(request, challenge_pk):
...
+                response_data = {
+                    "Success": "The challenge \"{}\"'s worker cores and memory were not "
+                               "modified.".format(challenge.title)
+                }
+            else:
+                response_data = {
+                    "Success": "The challenge \"{}\" has been scaled successfully!".format(
+                        challenge.title
+                    )
+                }
```
Ram81, line 3060: The success message should be `Scaled worker resources successfully!` we don't need to show title of the challenge.

Ram81 (changes requested):

apps/challenges/views.py
matthew-so: I added this case because while testing on staging, the recorded and actual worker counts often became out of sync, leading to the response not matching the actual situation. Doing it like this should help debug these desyncs and fix the cause of these issues.

apps/challenges/aws_utils.py
matthew-so: Done (I'm assuming that this is for the logger)

apps/challenges/aws_utils.py
matthew-so: Error messages here aren't propagated to the user; I decided to change the message in views.py corresponding to this case instead.

apps/challenges/views.py
matthew-so: I used that to test Django. Removed

apps/challenges/aws_utils.py
Ram81: @matthew-so I am saying even if they are not used we should use still follow same schema for each response from this method. We shouldn't introduce new changes to the response message schema. So, basically if we follow the convention of having either a `success` or a `error` key in the response we should stick to that and not add new keys like `IsWorkerConfigModified` Let me know if you have any questions.

matthew-so: Got it, thanks!

matthew-so: Fixed

Ram81 (approved):

PR #3864, opened 2023-01-15
Add auto scaling script for EKS nodegroups

gchhablani:
This PR adds auto-scaling script for EKS nodegroups.

I have exposed the size attributes for the challenges in the serializer in order to use them for scaling the nodegroups up.

Instead of `num_workers` from the challenge, I use `minSize` from `scalingConfig` from the nodegroups using the EKS boto3 client. I manually update it to 0 when there are no messages and it already isn't 0, otherwise I use the `minSize`/`maxSize`/`desiredSize` from the challenge which are exposed in the serializer.

The rest is similar to the worker autoscaling script.


apps/challenges/serializers.py
```diff
@@ -67,7 +67,10 @@ class Meta:
             "created_at",
             "queue",
             "worker_cpu_cores",
-            "worker_memory"
+            "worker_memory",
+            "min_worker_instance",
+            "max_worker_instance",
+            "desired_worker_instance"
```
Ram81, line 73: You didn't have to modify the relevant test for this to pass CI?

scripts/monitoring/auto_scale_eks_nodes.py
```diff
@@ -0,0 +1,267 @@
...
+utc = pytz.UTC
+NUM_PROCESSED_SUBMISSIONS = "num_processed_submissions"
+NUM_SUBMISSIONS_IN_QUEUE = "num_submissions_in_queue"
+PROMETHEUS_URL = os.environ.get(
+    "MONITORING_API_URL", "https://monitoring-staging.eval.ai/prometheus/"
+)
+
+PROD_INCLUDED_CHALLENGE_QUEUES = [
+    "habitat-rearrangement-challenge-2022-1820-production-e900ef02-eeb1-439a-b4a4-833",
+    "habitat-challenge-2022-1615-production-be08fce5-72a7-40bc-aa2e-4df7b3380a8a",
```
Ram81, line 25: Can we also add habitat challenge 2021 to this list?

scripts/monitoring/auto_scale_eks_nodes.py
```diff
@@ -0,0 +1,267 @@
...
+    environment_suffix = "{}-{}".format(challenge["id"], ENV)
+    nodegroup_name = "{}-{}-nodegroup".format(
+        challenge["title"].replace(" ", "-")[:20], environment_suffix
+    )
+    return nodegroup_name
+
+
+def get_eks_meta(challenge):
+    # TODO: Check if eks_client should be a global thing. Clients must have an expiry/timeout.
+    # Maybe it is better to re-create for every challenge. Not sure about this.
```
Ram81, line 89: Can we resolve this comment?

Ram81 (changes requested):
How was this PR tested?

scripts/monitoring/auto_scale_eks_nodes.py
gchhablani: @Ram81
I am adding separate lists for challenges on both prod and staging to handle the `ENV=dev` case.

apps/challenges/serializers.py
gchhablani: I will modify the test.

scripts/workers/code_upload_submission_worker.py
```diff
@@ -3,16 +3,15 @@
...
 
-
-from worker_utils import EvalAI_Interface
-
+import yaml
 from kubernetes import client
-
 # TODO: Add exception in all the commands
 from kubernetes.client.rest import ApiException
+from worker_utils import EvalAI_Interface
```
Ram81, line 12: Why do you need the EvalAI_Interface here?

Ram81 (changes requested):

scripts/workers/code_upload_submission_worker.py
gchhablani: I do not, I might have added it by mistake. Also will fix the formatting changes due to my version of black.

Ram81 (approved):
LGTM

PR #3871, opened 2023-02-12
Fix statsd increment in code-upload challenges

gchhablani:
This PR fixes the issue of not incrementing of `num_processed_submissions` when the submission is `finished`, `failed` or `cancelled` in case of code-upload challenge workers.

**Note**: I have changed the import order (fixed by isort).


Ram81 (approved):
LGTM

PR #3880, opened 2023-03-10
Fix #2142 Added a Highlighted border

### LINKED ISSUE #2142: Highlight entry in My Submissions Tab
## Current scenario

If a participant had made multiple entries public then the participant doesn't know which entry is visible on the leaderboard under `My submissions Tab` (Until they manually check it for the accuracies)

## Expected scenario

The entry which appears on the leaderboard should be highlighted in My submissions Tab.

Suryansh5545:
Fix #2142 Added a Highlighted border to show if the submission is public or private.

Peek 2023-03-10 13-15

### DIFF
--- a/frontend/src/css/modules/submission.scss
+++ b/frontend/src/css/modules/submission.scss
@@ -117,3 +117,7 @@ md-radio-button .md-on {
 	flex-direction: row;
 	align-items: center;
 }
+
+.is-public {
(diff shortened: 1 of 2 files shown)


gchhablani:
Closing this PR as it is not exactly what the expected feature is. @Suryansh5545 Let's try to implement what we discussed at some point.

PR #3881, opened 2023-03-10
Fix #2618 Added a Cookie Consent to Frontend_v1

### LINKED ISSUE #2618: Add a Cookie Consent
Currently we don't show the user the cookies consent which might make us non compiling to the international regularities . We should have a cookies consent display below for new users

Suryansh5545:
Fix #2618 Added a Cookie Consent to the Frontend_v1. As i am unsure if frontend_v2 is still in roadmap or not i have not implemented the consent in it. The message is kept simple and short. 
Peek 2023-03-10 14-47


Suryansh5545:
Not a Priority right now

PR #3911, opened 2023-03-28
Fix #3908 : Render phases data from API Response

### LINKED ISSUE #3908: Frontend: Show new value on validation success only
**Observed Behaviour**
- A host can update submission values in the "phases section" to any number, though it has a validation check which is failing but still host can see these new values at the client till the page is reloaded.

client_validation

**Expected Behaviour**
- New values should only be reflected at the client when the validation succeeds.

**Steps To Reproduce**
- Go to all challenges
- Select Any 1 challenge
- Go to "Phases" Section
- Edit any phase
- Type a very long number

TriveniBhat:
Issues - 
1. After editing, phase values were not updated with the API response.
2. Encountered Lint Issues while performing build action.

Solution - 
1. Corrected the described behavior by considering API response on success.
    In case of failures, the original data is not updated.
2.  Corrected the above behavior by adding a semi-colon and formatting.

### DIFF
--- a/frontend/src/js/controllers/challengeCtrl.js
+++ b/frontend/src/js/controllers/challengeCtrl.js
@@ -2562,7 +2562,7 @@
         };
 
         vm.challengePhaseDialog = function(ev, phase) {
-            vm.page.challenge_phase = phase;
+            vm.page.challenge_phase = Object.assign({},phase);
             vm.page.max_submissions_per_day = phase.max_submissions_per_day;
             vm.page.max_submissions_per_month = phase.max_submissions_per_month;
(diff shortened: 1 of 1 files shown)


gchhablani:
@TriveniBhat Can you please add a recording showing the frontend?

TriveniBhat:
@gchhablani I have attached the screencast for the UI behavior.
[Screencast from 04-25-2023 05:55:20 PM.webm](https://user-images.githubusercontent.com/49990128/234439113-d21a8215-e482-47c8-9d9a-b0d561ae5546.webm)

PR #3925, opened 2023-04-06
Fix representation in submission serializer

gchhablani:
This PR fixes the issue with submission serializer which is currently failing as `context` is not defined.


Ram81 (approved):

Ram81:
@gchhablani can you fix the failing build?

PR #3930, opened 2023-04-09
Fix #3923 : Show user timezone on submission worker logs

### LINKED ISSUE #3923: Show user timezone on submission worker logs
Currently, the timezone used on submission worker logs is UTC.

image

We would like to show the user's timezone on frontend. This will require processing the logs to convert the time from UTC and taking the user timezone. Might also require changes on the backend.

AyushR1:
This modifies the logs from the submission worker to show time according to the user's timezone on the front end. 

fix #3923 

Original
image

Modified (This displays according to client timezone (In this screenshot it is +5:30) )
image


gchhablani:
@AyushR1 This PR needs to be fixed.
The data format after adding the change is with `/`. I believe the format should be the same.

Issue #3926, opened 2023-04-07
Negative Execution Time on Django Admin during submission run


gchhablani:
The current Django Admin shows negative run time during the submission run. This is because we do not have a final time while the code is running but it should either be `None` (easy) or it should show elapsed time (harder).

![D2151881-4DFF-40C0-B01F-8F24D8A2EA4F_4_5005_c](https://user-images.githubusercontent.com/29076344/230675385-17b9ce36-24b6-4af4-adb0-ce08597af8e6.jpeg)

TriveniBhat:
@gchhablani I am interested in working on this. Could you assign this to me?

gchhablani:
@TriveniBhat Please go ahead. I will assign it to you.

priyapahwa:
Hey @TriveniBhat 
Are you still working on this issue?

TriveniBhat:
Hi @priyapahwa, I'm still working on it

PR #3943, opened 2023-04-19
Fixes negative execution time

TriveniBhat:
Case 1: when submission status is finished and completed_at and started_at values exist, execution_time is from the values 
Case 2: When the submission is running and only the started_at value exists, then execution_time is elapsed_time 
Case 3: Submission status is Submitted and Neither values are initialized then execution_time is None

Tested the above cases in admin site

### DIFF
--- a/apps/jobs/models.py
+++ b/apps/jobs/models.py
@@ -147,12 +147,13 @@ class Submission(TimeStampedModel):
     def execution_time(self):
         """Returns the execution time of a submission"""
         # if self.self.completed_at and self.started_at:
-        try:
+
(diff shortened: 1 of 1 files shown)


TriveniBhat:
fix #3926 
Test in Django admin
[Screencast from 04-18-2023 08:08:40 PM.webm](https://user-images.githubusercontent.com/49990128/232957488-130a1600-9db3-49ed-bd0a-8d790b8b1d9d.webm)

RishabhJain2018 (dismissed):
LGTM :) Thanks for the contribution.

PR #3945, opened 2023-04-19
[Deployment] Remove frontend v2 from deployment

gchhablani:
This PR removed frontend v2 from deployment on staging and production.


Ram81 (changes requested):
@gchhablani you also need to remove nodejs_v2 from here https://github.com/Cloud-CV/EvalAI/blob/master/scripts/deployment/deploy.sh#L53

Ram81 (approved):

PR #3967, opened 2023-05-25
Added function to disable user

Suryansh5545:
Using already described method disable_user. Disables user account and sends user to home page

Peek 2023-05-25 19-44


frontend/src/js/controllers/profileCtrl.js
```diff
@@ -123,6 +123,30 @@
...
+        // Disable User Account
+        vm.disableUserAccount = function() {
+            parameters.token = userKey;
+            parameters.method = 'POST';
+            parameters.url = 'accounts/user/disable';
+            parameters.callback = {
+                onSuccess: function(response) {
+                    var status = response.status;
+                    if (status == 200) {
+                        utilities.resetStorage();
```
gchhablani, line 135: What does this exactly do here?

gchhablani:
This PR looks good to me. The only thing I am not sure about is the placement of the button and whether it should be a button. For example `Change password` is not a button.

@rahul2240 @gautamjajoo Can you guys also review this PR?

frontend/src/js/controllers/profileCtrl.js
Suryansh5545: it cleans the local storage/cookies to log out the disable user. without it my browser was stuck on unexceptional error trying to access the disable account.

rahul2240: are we storing anything else in local storage in other places in the code?

rahul2240:
@gchhablani what about moving this to the update profile section and making it look like text instead of button? I think giving this button directly on the main profile is little straight forward :P and there should be a confirmation popup or modal too.

gchhablani:
@rahul2240 I am not sure whether this should be a part of update profile. Update profile is just for changing information.
I am still not sure where exactly on this page it should be, though. Maybe we can check with @RishabhJain2018 .

Also, 100% agree that there should be a confirmation popup.

@Suryansh5545 Can you please add that?

gautamjajoo (commented):
Deactivate button seems fine on the profile page, maybe it can be below the update profile button, with something in red.

and yes agreed with @rahul2240, we should add a confirmation modal. 
Code wise looks good

frontend/src/js/controllers/profileCtrl.js
Suryansh5545: we are storing the userkey, jwt token and profile image, other than this mostly errors using utilities.storeData

Suryansh5545:
Added The confirm dialog
Peek 2023-05-27 21-24

frontend/src/js/controllers/profileCtrl.js
```diff
@@ -123,6 +123,43 @@
             });
         };
 
+        // Disable User Account
+        vm.confirmdisableUserAccount = function(ev) {
+            ev.stopPropagation();
+            // Confirms that the user wants to deactivate the account
+            var confirm = $mdDialog.confirm()
+                .title('Would you like to deactivate your account?')
+                .textContent('Note: This action will deactivate your account and you will not be able to login again.')
```
gchhablani, line 132: @Suryansh5545 Can you check if you can add a page like this instead: https://github.com/Cloud-CV/EvalAI/blob/master/frontend/src/views/web/challenge/cancel-submission.html?

Check the code in challenge controller here: https://github.com/Cloud-CV/EvalAI/blob/3af33a2aee2f7a6dab90ed574dd3e2178d0901e8/frontend/src/js/controllers/challengeCtrl.js#L2079-L2114

gchhablani (commented):
Still not sure about the placement of the deactivate button. Pinging on the Slack channel, let's discuss there.

gautamjajoo (commented):
Here are a couple of things, @gchhablani I am also not sure about these points, you can also share your thoughts.
1. When a user deactivates an account, should we allow users to sign up again with the same username and email address? 
2. When a user has a submission in one of the challenges, and then deactivates his account, what should be the ideal response? should we highlight it or the host should know about it?
3. @Suryansh5545 I was thinking if we can have the modal in the same manner as we do when a host wants to delete a challenge. We can add an input to take the username. Attached the screenshot as well for reference.
Screenshot 2023-05-30 at 11 28 30 AM

PR #3968, opened 2023-05-26
Username can be changed from profile

Suryansh5545:
removed username field from read_only making it changeable from the update profile page.
Peek 2023-05-26 14-41


gautamjajoo (changes requested):
just check for edge cases as well, like whether a user with the updated username exists or not. Also, could you add test cases as well?

PR #3970, opened 2023-05-27
[Frontend] Add domains and tags to challenges

Suryansh5545:
Added Tags to backend, frontend, updated test case for the same.

Peek 2023-05-27 18-11


apps/challenges/challenge_config_utils.py
```diff
@@ -506,6 +506,26 @@ def validate_challenge_config_util(
             )
             error_messages.append(message)
 
+        # Verify Tags are limited to 4
+        if "tags" in yaml_file_data:
+            tags_data = yaml_file_data["tags"]
+            if len(tags_data) > 4:
+                message = (
```
gchhablani, line 513: Nice!

apps/challenges/challenge_config_utils.py
```diff
@@ -506,6 +506,26 @@ def validate_challenge_config_util(
...
+                    "ERROR: Tags are limited to 4. Please remove extra tags"
+                    " then try again!"
+                )
+                error_messages.append(message)
+
+        # Verify Domain name is correct
+        if "domain" in yaml_file_data:
+            domain_value = yaml_file_data["domain"]
+            if domain_value not in ["CV", "NLP", "RL"]:
```
gchhablani, line 522: I am not sure if this should be hardcoded.
Is there no way to take this list of domains from somewhere else?
Maybe import from models or somewhere?

apps/challenges/serializers.py
```diff
@@ -15,12 +15,40 @@
...
+
+    class Meta:
+        model = ChallengeTags
+        fields = "__all__"
+
+
 class ChallengeSerializer(serializers.ModelSerializer):
 
     is_active = serializers.ReadOnlyField()
+    list_tags = ChallengeTagsSerializer(many=True)
```
gchhablani, line 46: Why are we doing this?

docker/dev/django/container-start.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/sh
+python manage.py makemigrations && \
```
gchhablani, line 2: What is this change for?

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -1076,6 +1075,11 @@
 
             utilities.sendRequest(parameters);
         };
+
+        if (vm.phaseSplitId) {
```
gchhablani, line 1079: What is this change for?

apps/challenges/challenge_config_utils.py
Suryansh5545: It is definitely possible,  I will change this part

apps/challenges/admin.py
```diff
@@ -431,3 +432,10 @@ def get_challenge_name_and_id(self, obj):
 
     get_challenge_name_and_id.short_description = "Challenge Name - ID"
     get_challenge_name_and_id.admin_order_field = "challenge_phase__challenge"
+
+
+@admin.register(ChallengeTags)
+class TagsAdmin(ImportExportTimeStampedAdmin):
```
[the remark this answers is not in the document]
Suryansh5545: Exposes the ChallengeTags model to diango admin ui

apps/challenges/models.py
```diff
@@ -27,6 +27,22 @@ def get_default_eval_metric():
     return ["Accuracy"]
 
 
+class ChallengeTags(models.Model):
```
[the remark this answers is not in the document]
Suryansh5545: Like a ArrayField? We can do that but that won't be good looking on the admin side.

apps/challenges/serializers.py
Suryansh5545: We get the Id or index of list_tags as it is a manytomany in challenge. So we serialize it to get the value or name of tag

apps/challenges/views.py
```diff
@@ -1403,6 +1404,37 @@ def create_challenge_using_zip_file(request, challenge_host_team_pk):
...
+                # Remove tags not present in the YAML file
+                challenge.list_tags.set(ChallengeTags.objects.filter(tag_name__in=new_tags))
+
+                # Add new tags to the challenge
+                for tag_name in new_tags:
+                    tag_obj, _ = ChallengeTags.objects.get_or_create(tag_name=tag_name, challenge=challenge)
+                    challenge.list_tags.add(tag_obj)
+            else:
+                # Remove all existing tags if no tags are defined in the YAML file
+                challenge.list_tags.clear()
```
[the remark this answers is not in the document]
Suryansh5545: Yes it assumes that the tags that aren't listed in the updated config send are suppose to be removed

docker/dev/django/container-start.sh
Suryansh5545: Whenever a change is made to django make we have to run the command makemigrations separate every time. Shouldn't we put it in the start so all migrations are made and then migrated to db?

frontend/src/css/modules/challenge.scss
```diff
@@ -60,7 +60,7 @@ a.active-challenge {
 }
 
 .ev-challenge-card {
-    height: 425px;
+    min-height: 460px;
```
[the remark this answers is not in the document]
Suryansh5545: Adding the tags and domain made the card data get a  bit cut due to small size. It tries to compensate for it. But we still have that effect

apps/challenges/views.py
```diff
@@ -3392,6 +3469,37 @@ def create_or_update_github_challenge(request, challenge_host_team_pk):
...
+                        # Remove tags not present in the YAML file
+                        challenge.list_tags.set(ChallengeTags.objects.filter(tag_name__in=new_tags))
+
+                        # Add new tags to the challenge
+                        for tag_name in new_tags:
+                            tag_obj, _ = ChallengeTags.objects.get_or_create(tag_name=tag_name, challenge=challenge)
+                            challenge.list_tags.add(tag_obj)
+                    else:
+                        # Remove all existing tags if no tags are defined in the YAML file
+                        challenge.list_tags.clear()
```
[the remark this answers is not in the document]
Suryansh5545: yes it removed the old tag, in new approach too

PR #3989, opened 2023-06-19
Backend: Add approval requested parameter to avoid multple approval re…

AyushR1:
This would avoid the case when multiple approvals request are made


gchhablani:
I will let this PR be for now. I am closing it as it is not needed. We haven't seen too many multiple requests in a day. If we miss the message, they should be able to resend.

PR #4000, opened 2023-06-21
[Frontend] Make profile page tabular

Suryansh5545:
Showing Auth Token, Password , Profile in tabular form.

https://github.com/Cloud-CV/EvalAI/assets/34577232/d8b88f3d-6bcc-4b0f-867b-36e9802347a5


Suryansh5545:
The New UI look with fixed css

Selection_015

frontend/src/js/controllers/profileCtrl.js
```diff
@@ -169,6 +175,200 @@
...
+            else if (ev.currentTarget.id === "first_name") {
+                vm.titleInput = "First Name";
+                vm.editid = "first_name";
+            }
+            else if (ev.currentTarget.id === "last_name") {
+                vm.titleInput = "Last Name";
+                vm.editid = "last_name";
+            }
+            else if (ev.currentTarget.id === "affiliation") {
+                vm.titleInput = "Affiliation";
```
rahul2240, line 204: use switch case

frontend/src/js/controllers/profileCtrl.js
```diff
@@ -169,6 +175,200 @@
...
+                targetEvent: ev,
+                templateUrl: 'dist/views/web/profile/edit-profile/edit-profile.html',
+                escapeToClose: false
+            });
+        };
+
+        // function to update Profile
+        vm.updateProfile = function(resetconfirmFormValid, editid) {
+            if (resetconfirmFormValid) {
+                vm.user.github_url = vm.user.github_url === null ? "" : vm.user.github_url;
```
rahul2240, line 232: try "nullish coalescing operator" here

frontend/src/js/controllers/profileCtrl.js
Suryansh5545: ok

apps/accounts/serializers.py
```diff
@@ -20,15 +20,16 @@ class Meta:
             "last_name",
             "password",
         )
-        read_only_fields = ("email", "username")
```
gchhablani, line 23: Don't change this line. We will do username change later.

apps/accounts/serializers.py
```diff
@@ -20,15 +20,16 @@ class Meta:
...
+        read_only_fields = ("email",)
 
 
 class ProfileSerializer(UserDetailsSerializer):
     """
     Serializer to update the user profile.
     """
 
-    affiliation = serializers.CharField(source="profile.affiliation")
+    affiliation = serializers.CharField(source="profile.affiliation", allow_blank=True)
```
gchhablani, line 32: Why do we allow blank affiliation?

apps/accounts/serializers.py
```diff
@@ -53,6 +54,7 @@ class Meta(UserDetailsSerializer.Meta):
         )
 
     def update(self, instance, validated_data):
+        username = validated_data.get("username")
```
gchhablani, line 57: Remove this.

apps/accounts/serializers.py
```diff
@@ -64,7 +66,8 @@ def update(self, instance, validated_data):
         )
 
         profile = instance.profile
-        if profile_data and affiliation:
+        if profile_data:
+            profile.username = username
```
gchhablani, line 70: Remove this change.

frontend/src/js/controllers/profileCtrl.js
```diff
@@ -53,7 +58,7 @@
                             if (i === "linkedin_url" || i === "github_url" || i === "google_scholar_url") {
                                 result[i] = "";
                             } else {
-                                result[i] = "-";
+                                result[i] = "";
```
gchhablani, line 61: Is this change there to show nothing instead of `-`?
Why are we doing this?

frontend/src/js/controllers/profileCtrl.js
```diff
@@ -169,6 +175,193 @@
             };
             utilities.sendRequest(parameters);
         };
+
+        vm.isURLValid = function(url) {
+            if (url === undefined || url === null) {
+                return true;
+            }
+            return (url.length <= 200);
```
gchhablani, line 183: Why is this needed? Was is there earlier?

frontend/src/js/controllers/profileCtrl.js
```diff
@@ -169,6 +175,193 @@
...
+                            password1_valid = typeof(response.data.new_password1) !== 'undefined' ? true : false;
+                            password2_valid = typeof(response.data.new_password2) !== 'undefined' ? true : false;
+                            if (oldpassword_valid) {
+                                vm.FormError = Object.values(response.data.old_password).join(" ");
+                            }else if (password1_valid) {
+                                vm.FormError = Object.values(response.data.new_password1).join(" ");
+                            } else if (password2_valid) {
+                                vm.FormError = Object.values(response.data.new_password2).join(" ");
+                            }
+                        } catch (error) {
```
gchhablani, line 352: Why was this method added here?
How was it handled previously?

frontend/src/js/route-config/route-config.js
```diff
@@ -362,10 +362,47 @@
...
+            title: 'Editprofile',
+            authenticate: true
+        };
+
+        var ChangePassword = {
+            name: "web.profile.ChangePassword",
+            parent: "web.profile",
+            url: "/change-password",
+            templateUrl: baseUrl + "/web/change-password.html",
+            title: 'ChangePassword',
```
gchhablani, line 405: Keep the variable names like `update_profile` as they were earlier. The title should be `Update Profile`. Don't introduce new formats unless needed.

frontend/tests/controllers-test/profileCtrl.test.js
```diff
@@ -70,7 +71,7 @@ describe('Unit tests for profile controller', function () {
             for (var i = 0; i < successResponse.length; i++) {
                 if (successResponse[i] === "" || successResponse[i] === undefined || successResponse[i] === null) {
                     successResponse[i] = "-";
-                    expect(vm.countLeft).toEqual(vm.countLeft + 1);
```
gchhablani, line 73: Why was this supposed to be true?

frontend/tests/controllers-test/profileCtrl.test.js
```diff
@@ -70,7 +71,7 @@ describe('Unit tests for profile controller', function () {
...
+                    expect(vm.countLeft).toEqual(vm.countLeft);
```
gchhablani, line 74: What does this test anyway?

tests/unit/accounts/test_views.py
```diff
@@ -54,8 +54,8 @@ def test_cannot_update_username(self):
         response = self.client.put(
             os.path.join("api", "auth", str(self.url)), self.data
         )
-        self.assertNotContains(response, "anotheruser")
-        self.assertContains(response, "someuser")
+        self.assertNotContains(response, "someuser")
+        self.assertContains(response, "anotheruser")
```
gchhablani, line 58: Remove updating usernames!

gchhablani (commented):
Left some comments.

apps/accounts/serializers.py
Suryansh5545: This function is called for each instance of dialog we open , so we must allow it or we will have to make a completely separate function for this. in the earlier version we were editing everything in one place so we could make it a must but in this it isn't possible without a view function

frontend/src/js/controllers/profileCtrl.js
Suryansh5545: These all new functions were previously in a seprate updateprofile and changepassword files, since we are now working on a single page with most of the stuff. i added them here.

frontend/tests/controllers-test/profileCtrl.test.js
Suryansh5545: This and other instance of countLeft is basically to check if the value we get for any field except the url fields is null or empty if it is then we add a "-" to it for display purposes, i replaced that with emtpy as i have put other way of display this with doesn't affect our dialog box values

tests/unit/accounts/test_views.py
Suryansh5545: Done

PR #4002, opened 2023-06-26
[Backend] Add tests for code-upload worker

AyushR1:


scripts/workers/code_upload_submission_worker.py
```diff
@@ -10,8 +10,8 @@
 
 # TODO: Add exception in all the commands
 from kubernetes.client.rest import ApiException
-from statsd_utils import increment_and_push_metrics_to_statsd
-from worker_utils import EvalAI_Interface
+from .statsd_utils import increment_and_push_metrics_to_statsd
+from .worker_utils import EvalAI_Interface
```
gchhablani, line 14: Why was this change needed?

gchhablani (commented):
This PR looks good to me. Just that I'm not sure why the code upload submission worker was modified.

@gautamjajoo @rahul2240 Please provide another review.

scripts/workers/code_upload_submission_worker.py
gchhablani: @AyushR1 Are you sure this doesn't break any of the existing code? For example, if I create a code-upload challenge, the worker will work fine?

PR #4020, opened 2023-07-14
Fix issue with disable logs in submission worker

gchhablani:
This PR fixes the issue with disable logs in the submission worker.


Ram81 (changes requested):
Wouldn't it be better to just revert to original code and reset the `submission.stderr` and `submission.stdout` to None values?

Ram81 (approved):

PR #4021, opened 2023-07-15
[BugFix] Fix validate challenge for missing/extra YAML files

AyushR1:


gchhablani:
@AyushR1 I went over this PR.
Maybe there is some misunderstanding in the expectation.
Here is what is needs to be checked - a challenge config can have one or several YAML files. It doesn't matter as long as there is exactly one file named `challenge_config.yaml`. If there is no `challenge_config.yaml`, the validation should fail.

PR #4025, opened 2023-07-18
[Frontend] Added Ability to hide leaderboard from participants

Suryansh5545:
adds a variable to challenge model for leader board  visibility, and function to change the state

https://github.com/Cloud-CV/EvalAI/assets/34577232/969f6b39-fbc5-46f0-b7d6-9e1241390260


gchhablani:
When the task was there, it meant even if the leaderboard is private, it is visible to the challenge hosts. The ability to hide the leaderboard should come from challenge configuration. Basically, in the host's dropdown, we should not see particular leaderboard if they are marked hidden.

We can hide leaderboard from participants already.

Suryansh5545:
@gchhablani  i will make another PR with that changes and will delete this one

PR #4034, opened 2023-07-20
[Monitoring] Add auto cancel script

AyushR1:


scripts/monitoring/auto_cancel_submissions.py
```diff
@@ -0,0 +1,82 @@
...
+        submissions = evalai.get_submissions_for_challenge(
+            challenge_pk, "finished"
+        )
+        submissions += evalai.get_submissions_for_challenge(
+            challenge_pk, "running"
+        )
+        submissions += evalai.get_submissions_for_challenge(
+            challenge_pk, "resuming"
+        )
+        print(submissions)
```
gchhablani, line 38: Remove this print statement.

scripts/monitoring/auto_cancel_submissions.py
```diff
@@ -0,0 +1,82 @@
...
+        current_time = datetime.now(pytz.utc)
+        for submission in submissions:
+            status = submission["status"]
+            submission_time = get_submission_time(submission)
+            submission_time = pytz.utc.localize(submission_time)
+
+            time_difference = current_time - submission_time
+            if time_difference > timedelta(days=days_threshold):
+                data = {"status": "cancelled"}
+                evalai.update_submission_status(data, submission["challenge"])
```
gchhablani, line 49: What does `submission["challenge"]` return?

gchhablani: Are you sure this is correct?

scripts/monitoring/evalai_interface.py
```diff
@@ -119,8 +132,10 @@ def get_challenge_by_pk(self, challenge_pk):
         response = self.make_request(url, "GET")
         return response
 
-    def get_challenges(self):
+    def get_challenges(self, next_page_url=None):
         url = URLS.get("get_challenges")
         url = self.return_url_per_environment(url)
+        if next_page_url:
+            url = next_page_url
```
gchhablani, line 139: This is not needed. Add a while loop in the auto-cancel script just like other scripts.

tests/unit/challenges/test_views.py
```diff
@@ -4096,6 +4096,7 @@ def test_get_all_submissions_when_user_is_host_of_challenge(self):
                     "submitted_at": "{0}{1}".format(
                         submission.submitted_at.isoformat(), "Z"
                     ).replace("+00:00", ""),
+                    "rerun_resumed_at": None,
```
gchhablani, line 4099: These are all wrong.
Use `self.submission<x>.rerun_resumed_at` following others.

gchhablani (commented):
Added some comments.

tests/unit/challenges/test_views.py
gchhablani: Same as `started_at` etc.

scripts/monitoring/evalai_interface.py
ayushr1-zk: I will need to write the request function then here only as the url is passed with a param for the next page. The other places have there own request function written and doesn't really use the eval interface

gchhablani:
@AyushR1 Separate the addition of the new attribute and the script in two PRs. Keep this one only for auto-cancel.

scripts/monitoring/evalai_interface.py
gchhablani: @AyushR1 No. Have you checked EvalAI Interface? There is a `make_request` function that you can use.

gchhablani:
@AyushR1 There are other APIs to update submissions.

PR #4039, opened 2023-07-21
[Frontend] Add search and filter for challenges

Suryansh5545:
Uses angularjs filter directive with get_domain_choice function to pull the domain options dynamically from the backend.
image
image


Suryansh5545:
@gautamjajoo @gchhablani   Does it look good i am not sure about the placement of filter . also if we should have a reset button for the filter or just get user refresh the page

PR #4045, opened 2023-07-28
[Backend] Add method to check if a user is staff

Suryansh5545:
Adeed function to check if a user is staff


tests/unit/base/test_utils.py
```diff
@@ -134,3 +134,26 @@ def test_if_slack_notification_works(self):
         response = send_slack_notification(message=message)
         self.assertEqual(type(response), requests.models.Response)
         self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+
+class TestUserisStaff(BaseAPITestClass):
```
gchhablani, line 139: TestUserIsStaff

PR #4059, opened 2023-07-31
[Backend] Add duplicate challenge phase split validation

### LINKED ISSUE #4046: Duplicate Challenge Phase split
When creating a challenge phase split, we are able to create challenge phase split with same dataset split id,  challenge phase id and leaderboard id, this shouldn't be the case as it creates a duplicate entry for no use.

To maintain a well-organized and clear competition environment, each challenge phase split should have a unique combination of identifiers:

1. Dataset split ID: This identifier designates a specific split of the dataset, such as "train," "validation," or "test."
2. Challenge phase ID: This identifier represents the particular challenge phase, such as "Phase 1" or "Phase 2."
3. Leaderboard ID: This identifier is used to differentiate between multiple leaderboards, for instance, "Leaderboard 1" or "Leaderboard 2."

Suryansh5545:
Fixes #4046 
**Current Scenario**
The host can add a duplicate phase which same leaderboard, challenge phase and dataset id, this can create a duplicate entry which is useless.

**New Scenerio**
Host can no longer create a duplicate and is shown a error

image


gchhablani:
Show all duplicates in the error message.

PR #4064, opened 2023-07-31
[Feature] Add automatic EC2 creation for challenges

AyushR1:


gchhablani:
@AyushR1 Can you not create separate bash script that can be copied? It should not be a string.

ayushr1-zk:
> @AyushR1 Can you not create separate bash script that can be copied? It should not be a string.

Sure I will try it out and will read the file input on user data

PR #4098, opened 2023-08-10
[Monitoring] Add EC2 auto scaling script

AyushR1:


apps/challenges/serializers.py
```diff
@@ -83,6 +83,7 @@ class Meta:
             "cpu_only_jobs",
             "job_cpu_cores",
             "job_memory",
+            "uses_ec2_worker"
```
gchhablani, line 86: Add `ec2_instance_id` as well.

scripts/monitoring/auto_manage_ec2_workers.py
```diff
@@ -0,0 +1,172 @@
...
+    )
+
+    instances = [
+        instance
+        for reservation in response["Reservations"]
+        for instance in reservation["Instances"]
+    ]
+
+    if instances:
+        for instance in instances:
```
gchhablani, line 69: I don't think you should hard code the instance name. Instead, you can fetch the instance using `ec2_instance_id` directly and use that. Print the name of the instance you are scaling up or down.

Just check if the instance is in `Running` state or not. If it is `Running`, then stop.

scripts/monitoring/auto_manage_ec2_workers.py
```diff
@@ -0,0 +1,172 @@
...
+            response = ec2.stop_instances(InstanceIds=[instance_id])
+            print("AWS API Response: {}".format(response))
+            print(
+                "Stopped worker for Challenge ID: {}, Title: {}".format(
+                    challenge["id"], challenge["title"]
+                )
+            )
+    else:
+        print(
+            "No running instances with the specified name found for Challenge ID: {}, Title: {}. Skipping.".format(
```
gchhablani, line 80: `specified id`.

scripts/monitoring/auto_manage_ec2_workers.py
```diff
@@ -0,0 +1,172 @@
...
+    instances = [
+        instance
+        for reservation in response["Reservations"]
+        for instance in reservation["Instances"]
+    ]
+
+    if instances:
+        for instance in instances:
+            instance_id = instance["InstanceId"]
+            response = ec2.start_instances(InstanceIds=[instance_id])
```
gchhablani, line 108: Same thing here. Also, if multiple instances are created with same name due to approving and unapproving several times, this will start all the instances which is wrong.

gchhablani (commented):
Added some comments.

PR #4111, opened 2023-08-11
[Backend] Add new domain options

Suryansh5545:
Add option Multimodal and Audio to domain options


gchhablani:
@Suryansh5545 Add tabular as well.

PR #4117, opened 2023-08-11
[BugFix] Fix EC2 auto scaling when initializing instance

### LINKED ISSUE #4116: No auto scale on intializing/stopping: Auto scaling should not work when instance is initializing or stopping.

AyushR1:
fix #4116


apps/challenges/aws_utils.py
```diff
@@ -471,54 +471,51 @@ def service_manager(
...
-            except ClientError as e:
-                logger.exception(e)
+    response = ec2.describe_instance_status(InstanceIds=[target_instance_id])
+
+    instance_statuses = response["InstanceStatuses"]
+
+    if instance_statuses:
+        instance_status = instance_statuses[0]
+        if instance_status["SystemStatus"]["Status"] == "ok" and instance_status["InstanceStatus"]["Status"] == "ok":
+            instance_state = ec2.describe_instances(InstanceIds=[target_instance_id])["Reservations"][0]["Instances"][0]["State"]["Name"]
```
gchhablani, line 492: We don't need to call this API again. We already have the `response`. Use it.

PR #4135, opened 2023-08-16
Add custom scaling for challenge 356 and 830

Suryansh5545:
The current issue of not being able to edit the challenge 356 and 830 has made it difficult to use the regular auto scaling on these challenge so the custom function fixes this by assuming a default name format to get the instance and then start or stop it


Suryansh5545:
@AyushR1 Can you please review this PR

PR #4139, opened 2023-08-17
[Backend] Custom SQS retention time

Suryansh5545:
Currently we have a fixed sqs retention time for all challenge which we get from settings file. this PR adds a sqs time element to the model


gchhablani:
Will review this PR after GSoC period. This is not a part of GSoC.

PR #4156, opened 2023-08-31
[Frontend] Add search and filter for challenges

Suryansh5545:
Uses angularjs filter directive with get_domain_choice function to pull the domain options dynamically from the backend.


Suryansh5545:
adds a "all" and "none" options
https://github.com/Cloud-CV/EvalAI/assets/34577232/af82f096-e072-43c1-be89-6801936b3911

PR #4180, opened 2023-09-26
[Backend] Allow disabling leaderboard data

Suryansh5545:
To fix the duplicate leaderboarddata problem, instead of deleting them we are disabling them so they don't show up on leaderboard or are used

https://github.com/Cloud-CV/EvalAI/assets/34577232/e988a2fc-3f3a-4a01-8f6a-7ea2877817e0


apps/challenges/views.py
```diff
@@ -2,6 +2,7 @@
 import json
 import logging
 import os
+from django.shortcuts import get_object_or_404
```
gchhablani, line 5: I don't think we use this anywhere except home page. Can you please fix this? 
Use a regular error instead of this. Things should be consistent.

apps/challenges/views.py
```diff
@@ -160,6 +161,10 @@
     send_emails,
 )
 
+from jobs.utils import (
+    get_submission_model,
+)
+
```
gchhablani, line 167: Can be done in a single line.

gchhablani (commented):
Left some comments.

PR #4188, opened 2023-10-05
[Backend] Add is_disabled to Challenge Phase and Challenge Phase Splits and allow new phase and split creation

Suryansh5545:
Instead of deleting the challenge phases and splits we are allowing to disable them so they won't be pulled or used if disabled. This PR also adds the feature to add new phases


apps/challenges/migrations/0106_challenge_phase_is_disabled.py
```diff
@@ -0,0 +1,28 @@
...
+            model_name='challengephasesplit',
+            name='is_disabled',
+            field=models.BooleanField(default=False),
+        ),
+        migrations.AlterField(
+            model_name='challenge',
+            name='worker_image_url',
+            field=models.CharField(blank=True, default='', max_length=200, null=True),
+        ),
+    ]
```
gchhablani, line 28: What is this addition? Why are you adding `worker_image_url` here?

gchhablani (commented):
Are you sure you have covered all cases where we get Challenge Phases? Why would we want non disabled challenge phases in `challenge_phase_list` and `challenge_phases`? I think we would want all (even disabled ones).
Please go over each again and make sure that only when we are using these for some other purpose (getting submissions, leaderboard data), only then we check for is disabled.

PR #4225, opened 2023-11-26
[Backend] Add automatic Forum Creation

Suryansh5545:
Auto Creates Forum page if enable_forum is true


RishabhJain2018:
Closing this since we don't have forum anymore.

Issue #4068, opened 2023-07-31
Reduce size of deregister button


gchhablani:
The size of deregister button is too much compared to surrounding text.

gchhablani:
Also
- Team name is not shown properly.
<img width="518" alt="Screenshot 2023-07-31 at 2 50 20 PM" src="https://github.com/Cloud-CV/EvalAI/assets/29076344/5f7afc63-bb82-4288-b2f2-2cd2d21c1968">
- Use lowercase `t` in team.
- After deregister, I see this:
<img width="642" alt="Screenshot 2023-07-31 at 2 51 04 PM" src="https://github.com/Cloud-CV/EvalAI/assets/29076344/0d35cf3b-1e8a-43fd-8b7e-2f95178521c1">

gchhablani:
<img width="1171" alt="Screenshot 2023-07-31 at 2 52 28 PM" src="https://github.com/Cloud-CV/EvalAI/assets/29076344/500bfa1e-72d0-4522-a5ee-61d30c2d44cf">

PR #4267, opened 2024-02-10
#4068 shift deregister button right

harryytsao:


gchhablani:
Can you please explain why we are trying to do this?

PR #4268, opened 2024-02-18
"Fix #4259:" Handles the AWS credentials error .

### LINKED ISSUE #4259: Add better error handling for getting AWS credentials for a challenge
- In case the API doesn't work.
- In case it works but there is an error in getting the credentials.

Vinnu124:
This PR handles the aws credentials as mentioned in the issue #4259 by implementing a try except block in the get_aws_credentials_for_participant_team and get_aws_credentials_for_submission functions.

### DIFF
--- a/apps/challenges/utils.py
+++ b/apps/challenges/utils.py
@@ -373,24 +373,27 @@ def get_aws_credentials_for_submission(challenge, participant_team):
             "docker_repository_uri"
         }
     """
-    aws_keys = get_aws_credentials_for_challenge(challenge.pk)
-    ecr_repository_name = "{}-participant-team-{}".format(
-        challenge.slug, participant_team.pk
-    )
-    ecr_repository_name = convert_to_aws_ecr_compatible_format(
-        ecr_repository_name
-    )
-    repository, created = get_or_create_ecr_repository(
-        ecr_repository_name, aws_keys
-    )
-    name = str(uuid.uuid4())[:32]
-    docker_repository_uri = repository["repositoryUri"]
-    federated_user = create_federated_user(name, ecr_repository_name, aws_keys)
-    return {
-        "federated_user": federated_user,
-        "docker_repository_uri": docker_repository_uri,
-    }
+    try:
+        aws_keys = get_aws_credentials_for_challenge(challenge.pk)
+        ecr_repository_name = "{}-participant-team-{}".format(
+            challenge.slug, participant_team.pk
+        )
+        ecr_repository_name = convert_to_aws_ecr_compatible_format(
+            ecr_repository_name
+        )
+        repository, created = get_or_create_ecr_repository(
+            ecr_repository_name, aws_keys
+        )
+        name = str(uuid.uuid4())[:32]
+        docker_repository_uri = repository["repositoryUri"]
+        federated_user = create_federated_user(name, ecr_repository_name, aws_keys)
+        return {
+            "federated_user": federated_user,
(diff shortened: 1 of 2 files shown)


apps/challenges/utils.py
```diff
@@ -493,7 +498,8 @@ def add_tags_to_challenge(yaml_file_data, challenge):
         tags_data = yaml_file_data["tags"]
         new_tags = set(tags_data)
         # Remove tags not present in the YAML file
-        challenge.list_tags = [tag for tag in challenge.list_tags if tag in new_tags]
```
Suryansh5545, line 496: Why this change? seems unrelated to the issue

apps/challenges/utils.py
```diff
@@ -542,7 +548,8 @@ def add_prizes_to_challenge(yaml_file_data, challenge):
             amount = prize["amount"]
             description = prize["description"]
 
-            prize_obj = ChallengePrize.objects.filter(rank=rank, challenge=challenge).first()
```
Suryansh5545, line 545: Same here?

apps/challenges/utils.py
```diff
@@ -589,7 +596,8 @@ def add_sponsors_to_challenge(yaml_file_data, challenge):
         for sponsor in sponsors_data:
             # Checking if the sponsors already exists in the database.
             sponsor_name_set.add(sponsor['name'])
-            check_sponsor_status = ChallengeSponsor.objects.filter(name=sponsor['name'], challenge=challenge).exists()
```
Suryansh5545, line 592: Same

Suryansh5545 (changes requested):
Fix the mentioned changes

apps/challenges/utils.py
Vinnu124: It was an oversight at my end.Rectified the changes.

apps/challenges/views.py
```diff
@@ -2830,25 +2830,32 @@ def get_aws_credentials_for_participant_team(request, phase_pk):
...
+            return Response(response_data, status=status.HTTP_400_BAD_REQUEST)
 
-    if participant_team is None:
+        data = get_aws_credentials_for_submission(challenge, participant_team)
+        response_data = {"success": data}
+        return Response(response_data, status=status.HTTP_200_OK)
+    except Exception as e:
         response_data = {
-            "error": "You have not participated in this challenge."
+            "error": f"An error occurred in fetching the credentials: {str(e)}"
```
gchhablani, line 2856: Not sure why this was changed?

apps/challenges/utils.py
```diff
@@ -373,23 +373,28 @@ def get_aws_credentials_for_submission(challenge, participant_team):
...
+        docker_repository_uri = repository["repositoryUri"]
+        federated_user = create_federated_user(
+            name, ecr_repository_name, aws_keys)
+        return {
+            "federated_user": federated_user,
+            "docker_repository_uri": docker_repository_uri,
+        }
+
+    except Exception as e:
+        raise e
```
gchhablani, line 397: The entire point of handling exceptions is to not raise errors, but return a response so user knows where things are breaking.
This is not what is expected.

gchhablani: Check how we return responses elsewhere.

gchhablani (commented):
Added some comments.
Just wanted to know your thoughts on this: is a single try-except enough?

apps/challenges/utils.py
Vinnu124: Will look into this.

PR #4270, opened 2024-02-22
Fix #4089: Add warning for submission regarding deregistering

### LINKED ISSUE #4089: Add warning for submission regarding deregistering
Whenever a submission is made for the first time by a participant team, they should get a warning that they will not able to deregister after that.

Harshit28j:
This PR adds a warning message for submissions, reminding users about the implications of deregistering after making their first submission. 
**Issue #4089 is now addressed.**

### DIFF
--- a/frontend/src/js/controllers/challengeCtrl.js
+++ b/frontend/src/js/controllers/challengeCtrl.js
@@ -82,7 +82,7 @@
         vm.currentPhaseLeaderboardPublic = false;
         vm.eligible_to_submit = false;
         vm.evaluation_module_error = null;
-
+        vm.warningFlag = false;
         vm.filter_all_submission_by_team_name = '';
         vm.filter_my_submission_by_team_name = '';
         // show loader
@@ -294,10 +294,14 @@
             parameters.data={};
             parameters.callback = {
                 onSuccess: function(response) {
-                     var details = response.data;
+                    var details = response.data;
                     vm.participated_team_name = details["team_name"];
                     vm.eligible_to_submit = details["approved"];
-                },
+                    vm.hasSubmission=(details["hasSubmissions"]);
+                    if(!vm.hasSubmission){      //if participant has made no submission before warn them for 1st submission
+                        vm.warningFlag = true;
+                    } 
+                },  
             };
             utilities.sendRequest(parameters);
         };
@@ -3046,7 +3050,6 @@
                 if (vm.termsAndConditions) {
                     vm.selectExistTeam();
                     $mdDialog.hide();
-                    localStorage.setItem('warningFlag', 'true');
                 }
             } else {
                 $mdDialog.hide();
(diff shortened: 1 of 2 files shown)


frontend/src/views/web/challenge/submission.html
```diff
@@ -378,8 +378,11 @@ <h5 class="w-300">Make Submission</h5>
                                     <div class="row rm-gut">
                                         <div class="col s12 m12">
                                             <div class="align-left">
-                                                <button class="waves-effect waves-dark btn ev-btn-dark w-300 fs-16"
-                                                    ng-disabled="challenge.disableSubmit || (!challenge.input_file && !challenge.fileUrl)"
```
gchhablani, line 382: Why are we removing these lines?

frontend/src/views/web/challenge/submission.html
```diff
@@ -378,8 +378,11 @@ <h5 class="w-300">Make Submission</h5>
...
+                                                <div class="alert alert-warning" role="alert" ng-show="challenge.warningFlag">
+                                                    <strong>Warning:</strong> You won't be able to deregister after making your first submission.
+                                                </div>
+                                                <br>
```
gchhablani, line 384: I will let @Suryansh5545 review this.

gchhablani:
Can you please also add a GIF or a video showing the changes?

frontend/src/js/controllers/challengeCtrl.js
```diff
@@ -3044,6 +3045,7 @@
         vm.acceptTermsAndConditions = function (acceptTermsAndConditionsForm) {
             if (acceptTermsAndConditionsForm) {
                 if (vm.termsAndConditions) {
+                    vm.warningFlag=true;
```
Suryansh5545, line 3048: @Harshit28j This approach assumes that the participant will do submission after accepting the terms. But if the page is refreshed after that then this doesn't work. Try to come up with some way to identify if he has made the submission through the API calls.

Suryansh5545 (changes requested):
Added a comment address that. Also i believe it would be much better to show warning to user through a dialog box like we do in deregister button.

Suryansh5545:
> Here's another approach, updated with **local storage (cookie),** Now its working even after refresh @Suryansh5545
> InShot_20240307_023447630.mp4

Can you try to do it using the data that backend sends on submit tab instead if possible. i think that would be a better approach instead of localStorage.

Suryansh5545:
> > > Here's another approach, updated with **local storage (cookie),** Now its working even after refresh @Suryansh5545
> > > InShot_20240307_023447630.mp4
> > 
> > 
> > @Harshit28j Can you try to do it using the data that backend sends on submit tab instead if possible. i think that would be a better approach instead of localStorage.
> 
> @Suryansh5545 I'll give it a try, but could you please provide more details about the specific data you're referring to on the submit tab? Also, if I make changes based on .isParticipated, we'd end up with the same issue as before, where the warning isn't displayed upon refresh.

Hey @Harshit28j, I was thinking that we could enhance the `participant_team/team_detail` API call to include a check for existing submissions. This way, we can provide a warning message if the team doesn't have a submission yet. What do you think?

PR #4276, opened 2024-03-07
Fix #4242: Updated Deploy and Pull scripts for TravisCI to report it as failed on issue

### LINKED ISSUE #4242: Travis Build Status to Reflect Deployment Script Failures
We've observed that in certain scenarios, such as the recent [example](https://github.com/Cloud-CV/EvalAI/pull/4239), when our deployment or push script fails, TravisCI reports the build as successful. This can be misleading and may lead to confusion.

We need the TravisCI to report it as failed if the scripts have failed to push or deploy

abhi-bhatra:
## Expected Behaviour
Deploy and Pull Scripts should send exit code 1 when failure occurs

## Actual Behaviour
From the [example](https://github.com/Cloud-CV/EvalAI/pull/4239), deploy.sh script is sending exit code 0, which means success and Travis CI, in return make build successful.
image

## Fix: 
1. Deploy.sh: Create a function to check for *AWS_ACCOUNT_ID*, and this function will only be called for the cases which uses this variable, for others cases, it will not check for this value.

2. Pull.sh: if env is production/staging, and error occurs in *build_and_push* function, it exits with error code 1.

### DIFF
--- a/scripts/deployment/deploy.sh
+++ b/scripts/deployment/deploy.sh
@@ -8,10 +8,12 @@ aws_login() {
     eval $(aws ecr get-login --no-include-email)
 }
 
-if [ -z ${AWS_ACCOUNT_ID} ]; then
-    echo "AWS_ACCOUNT_ID not set."
-    exit 0
-fi
+aws_account_id_check() {
+    if [ -z ${AWS_ACCOUNT_ID} ]; then
+        echo "AWS_ACCOUNT_ID not set."
+        exit 1
+    fi
+}
 
 if [ -z ${COMMIT_ID} ]; then
     export COMMIT_ID="latest"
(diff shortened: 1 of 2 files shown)


abhi-bhatra:
Hi @Suryansh5545 you are referring to this build failure, where django installation failed, right ?

image

Suryansh5545:
> Hi @Suryansh5545 you are referring to this build failure, where django installation failed, right ?
> 
> image

@abhi-bhatra Yes, we want to make sure that if the scripts exits due to any kind of error then the exit is set to 1 instead of 0. To indicate failure of deployment

abhi-bhatra:
Hi @Suryansh5545 as you said that the exit 0 was just fine in deploy and push scripts and working as expected. Then, I believe there is an issue with Travis yml manifest.
There is a flag `set -e` is set in both the scripts, which send a non-zero exit code immediately, the script fails on any step. I have set a **travis_terminate 1;** which exits the pipeline if any of the script is failed.

PR #4281, opened 2024-03-09
Fix #4280: Improve challenge title validation

### LINKED ISSUE #4280: Improve challenge validation to check for alphanumeric characters

ay-bh:
Allow only alphanumerics and symbols (space, hyphen, underscore, period, comma, apostrophe, exclamation point, ampersand) in challenge titles. It resolves issue #4280.

### DIFF
--- a/apps/challenges/challenge_config_utils.py
+++ b/apps/challenges/challenge_config_utils.py
@@ -227,6 +227,7 @@ error_message_dict = {
     "multiple_yaml_files": "There are {} challenge config YAML files instead of 1 in the zip file!",
     "yaml_file_read_error": "\n{} in line {}, column {}\n",
     "missing_challenge_title": "Please add the challenge title",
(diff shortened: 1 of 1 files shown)


apps/challenges/challenge_config_utils.py
```diff
@@ -227,6 +227,7 @@ def get_value_from_field(data, base_location, field_name):
     "multiple_yaml_files": "There are {} challenge config YAML files instead of 1 in the zip file!",
     "yaml_file_read_error": "\n{} in line {}, column {}\n",
     "missing_challenge_title": "Please add the challenge title",
+    "invalid_challenge_title": "Challenge title contains invalid characters.",
```
Suryansh5545, line 230: Error message is a bit vague, it should include the allowed characters range.

ay-bh: @Suryansh5545 Please check now.

PR #4285, opened 2024-03-09
Fix #4280: Improve challenge validation to check for alphanumeric characters

### LINKED ISSUE #4280: Improve challenge validation to check for alphanumeric characters

AishwaryJain07:
This Fix solves the problem of names that aren't recognised by the database.

### DIFF
--- a/apps/challenges/challenge_config_utils.py
+++ b/apps/challenges/challenge_config_utils.py
@@ -378,6 +378,13 @@ class ValidateChallengeConfigUtil:
         if not challenge_title or len(challenge_title) == 0:
             message = self.error_messages_dict.get("missing_challenge_title")
             self.error_messages.append(message)
+        else:
+            # THese are allowed characters
+            allowed_pattern = r"^[a-zA-Z0-9\s\-_\.\,\'\!\&]+$"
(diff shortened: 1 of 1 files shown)


apps/challenges/challenge_config_utils.py
```diff
@@ -378,6 +378,13 @@ def validate_challenge_title(self):
         if not challenge_title or len(challenge_title) == 0:
             message = self.error_messages_dict.get("missing_challenge_title")
             self.error_messages.append(message)
+        else:
+            # THese are allowed characters
+            allowed_pattern = r"^[a-zA-Z0-9\s\-_\.\,\'\!\&]+$"
+            match = re.match(allowed_pattern, challenge_title)
+            if not match:
+                message = "Challenge title can only contain alphanumeric characters, spaces, hyphen, underscore, period, comma, apostrophe, exclamation point, and ampersand."
```
Suryansh5545, line 386: @AishwaryJain07 Instead of putting the message here, i think we should keep it with all the other error messages in `error_message_dict`

AishwaryJain07: @Suryansh5545 done

Suryansh5545 (approved):
LGTM🎉

PR #4286, opened 2024-03-10
Fix #4280 Improve challenge title validation

### LINKED ISSUE #4280: Improve challenge validation to check for alphanumeric characters

poju3185:
Fix #4280 

Use regex to match alphanumeric characters and the common punctuation symbols, specificly `.,!?()-[]`


Suryansh5545 (approved):
LGTM

PR #4297, opened 2024-03-13
Fix #3935 Warn users that their tokens have expired.

### LINKED ISSUE #3935: Warn users that their tokens have expired
Warn users that their tokens have expired.
Find a way to suggest to the users that their tokens might have expired when they try to fetch the tokens.
Suggest is in a error message or remind them to refresh tokens. This might also need backend changes to check if token is expired and probably writing a frontend function to call that endpoint.

Harshit28j:
This PR resolves #3935

image


Suryansh5545 (commented):
This looks good. But I was thinking maybe we should prompt the user with a dialog instead and the dialog should have the button to dismiss and refresh the token. What do you think @gchhablani ?

RishabhJain2018:
Since the comments are not addressed, I am closing this PR.

PR #4298, opened 2024-03-14
FIX#3935: Warn users that their tokens have expired

### LINKED ISSUE #3935: Warn users that their tokens have expired

AishwaryJain07:
Added Feature to Warn the users If their Tokens are expired and Give them a Warning that suggests to refresh the Token and try again.
closes #3935


AishwaryJain07:
@Suryansh5545 can you please review this !!

Suryansh5545 (commented):
LGTM, but same as the other PR i think we should go with a dialog instead.

RishabhJain2018:
Since the comments are not addressed, I am closing this PR.

PR #4301, opened 2024-03-15
FIX#4229: Check challenge phase is corresponding to challenge in update submission API

### LINKED ISSUE #4229: Check challenge phase is corresponding to challenge in update submission API
Here, we do check for whether the host is accessing the challenge: https://github.com/Cloud-CV/EvalAI/blob/master/apps/jobs/views.py#L1117-L1121
```
    if not is_user_a_staff(request.user) and not is_user_a_host_of_challenge(request.user, challenge_pk):
        response_data = {
            "error": "Sorry, you are not authorized to make this request!"
        }

```

But, we don't verify whether the other details (challenge phase) are corresponding to that challenge PK:
https://github.com/Cloud-CV/EvalAI/blob/8bf1c3dca1f753bcb59dbe12b1f340383755eb60/apps/jobs/views.py#L1124
```
        challenge_phase_pk = request.data.get("challenge_phase")
```

We need to fix this.

AishwaryJain07:
added try-except block to verify the other details while accessing the challenge
closes #4229

### DIFF
--- a/apps/jobs/views.py
+++ b/apps/jobs/views.py
@@ -1120,8 +1120,23 @@ def update_submission(request, challenge_pk):
         }
         return Response(response_data, status=status.HTTP_400_BAD_REQUEST)
 
+    challenge_phase_pk = request.data.get("challenge_phase")
+
+    try:
+        challenge_phase = ChallengePhase.objects.get(pk=challenge_phase_pk)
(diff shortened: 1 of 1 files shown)


AishwaryJain07:
@RishabhJain2018  please review this one,  aded a try except block

PR #4310, opened 2024-03-21
FIX#4259: Add better error handling for getting AWS credentials for a challenge 

AishwaryJain07:
Here, I added Try-catch block to look for the errors, and also added the looging message to indicate the retrieval of challenge_phase id and passes a m,message after it retrieves of challenge_phase id.

### DIFF
--- a/apps/challenges/views.py
+++ b/apps/challenges/views.py
@@ -2830,24 +2830,33 @@ def get_aws_credentials_for_participant_team(request, phase_pk):
             - When participant_team has not participanted in challenge
             - When Challenge is not Docker based
     """
-    challenge_phase = get_challenge_phase_model(phase_pk)
-    challenge = challenge_phase.challenge
-    participant_team = get_participant_team_of_user_for_a_challenge(
-        request.user, challenge.pk
-    )
-    if not challenge.is_docker_based:
-        response_data = {
-            "error": "Sorry, this is not a docker based challenge."
-        }
-        return Response(response_data, status=status.HTTP_400_BAD_REQUEST)
+    try:
+        logging.debug(f"Retrieving challenge phase for ID: {phase_pk}")
+        challenge_phase = get_challenge_phase_model(phase_pk)
+        logging.debug(f"Challenge phase retrieved: {challenge_phase}")
 
-    if participant_team is None:
-        response_data = {
-            "error": "You have not participated in this challenge."
-        }
+        challenge = challenge_phase.challenge
+
+        participant_team = get_participant_team_of_user_for_a_challenge(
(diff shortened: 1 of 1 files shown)


AishwaryJain07:
closes #4259

AishwaryJain07:
@Suryansh5545 @gchhablani @RishabhJain2018 can you please review

apps/challenges/views.py
```diff
@@ -2830,24 +2830,33 @@ def get_aws_credentials_for_participant_team(request, phase_pk):
...
-    participant_team = get_participant_team_of_user_for_a_challenge(
-        request.user, challenge.pk
-    )
-    if not challenge.is_docker_based:
-        response_data = {
-            "error": "Sorry, this is not a docker based challenge."
-        }
-        return Response(response_data, status=status.HTTP_400_BAD_REQUEST)
+    try:
+        logging.debug(f"Retrieving challenge phase for ID: {phase_pk}")
```
Suryansh5545, line 2834: Could you please provide an explanation for why this change was added?

apps/challenges/views.py
```diff
@@ -2830,24 +2830,33 @@ def get_aws_credentials_for_participant_team(request, phase_pk):
...
+        challenge_phase = get_challenge_phase_model(phase_pk)
+        logging.debug(f"Challenge phase retrieved: {challenge_phase}")
```
Suryansh5545, line 2836: Same here @AishwaryJain07

Suryansh5545 (commented):
Left some comments to be addressed

apps/challenges/views.py
AishwaryJain07: @Suryansh5545 it creates a log entry, developers can access for troubleshooting and understanding the code's behavior.

the idea of logging was mentioned by @gchhablani in his pr, "https://github.com/Cloud-CV/EvalAI/pull/4261".

this was all the behaviour i tried to add in the code, kindly let me know further any.
plus this also adds the try-catch block which already @gchhablani did.

AishwaryJain07:
> Left some comments to be addressed

@Suryansh5545 the log only creates an entry for developers to understand.
like a debugging message.

apps/challenges/views.py
Suryansh5545: @AishwaryJain07 i think you misunderstood it. we want to give a error message to the user as return if something goes wrong instead of a generic message. you are going in the right direction by capturing the error and sending it back as a response. but the log thing isn't required for now.

gchhablani: @AishwaryJain07 We do not want logging.

PR #4333, opened 2024-04-07
Fixes #4330: Auth Token Not updated in UI

### LINKED ISSUE #4330: Frontend: Auth Token doesn't get updated on UI if the user logs on after a long time.
**Issue:**

If a user logs in after a very long time, the UI still shows the old auth token. See the screenshot -
Screenshot 2024-04-06 at 12 25 28 PM

**Expected Behaviour:**

The updated token should be shown to the user on the UI.

cc: @gchhablani @gautamjajoo

Yashovardhan08:
Bug #4330 fix : Removing user key from local storage on every login to clear old token data, so that the UI is updated with the proper API Token

### DIFF
--- a/frontend/src/js/controllers/authCtrl.js
+++ b/frontend/src/js/controllers/authCtrl.js
@@ -232,6 +232,7 @@
                 };
                 parameters.callback = {
                     onSuccess: function(response) {
+                        utilities.deleteData("userKey");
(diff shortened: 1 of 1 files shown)


RishabhJain2018:
Closing this since it is an old PR and might not work with the current codebase.

PR #4345, opened 2024-04-19
[BugFix] Add success condition for result file update in worker

gchhablani:
This PR adds success condition for the result file to be updated in the submission worker. This prevents getting results for the failed submissions.


RishabhJain2018 (commented):
LGTM!

PR #4356, opened 2024-04-26
[Deployment] Use FluentBit insights instead of FluentD

gchhablani:


gchhablani:
Need to switch to older version because of similar issues:
https://github.com/fluent/fluent-bit/issues/2927

PR #4372, opened 2024-06-07
Delete infrastructure for code upload challenge

MinhThieu145:
This pull request updates the challenge/aws_utils.py file to delete the infrastructure for the code upload challenge. 

Resources got delete:
- IAM roles
- EFS 
- EKS cluster
- VPC

### DIFF
--- a/apps/challenges/aws_utils.py
+++ b/apps/challenges/aws_utils.py
@@ -1060,6 +1060,268 @@ def scale_resources(challenge, worker_cpu_cores, worker_memory):
         logger.exception(e)
         return e.response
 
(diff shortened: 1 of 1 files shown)


MinhThieu145:
Currently having errors with passing the Travis Test Cases. The error return was 
```
  File "/code/apps/challenges/aws_utils.py", line 15, in <module>

    from .models import ChallengeEvaluationCluster

ImportError: cannot import name 'ChallengeEvaluationCluster' from 'challenges.models' (/code/apps/challenges/models.py)
```

The import is accurate, VS can detect the import, and there are similar import already in the code (but the import was inside a function, so I add 1 at the top)

PR #4376, opened 2024-06-08
[Backend] Update EKS cluster version to 1.29

MinhThieu145:
Modify the challenges/aws_utils.py to change the new EKS cluster created with Kubernete version 1.29, upgrade from the previous 1.23


gchhablani:
LGTM.
Let's test it out on staging.

PR #4392, opened 2024-07-03
Fix #4274 Add or delete challenge phase/phase splits

### LINKED ISSUE #4274: Add ability to add or delete challenge phase and splits
Currently its not possible to add or delete challenge phases and splits. The host should be able to delete phases and splits by removing them from the yaml file. This was previously implemented in #4157. But it was revert due to a bug that the phases were deleted and recreated everytime the yaml file is used to update the challlenge.

Harshit28j:
This PR resolves #4274

I debug each and every functionality and statements to find out the root cause of this problem and I figured out that issue is from the challenge_phase_split_uuids (which actually holds data value from yaml)

I removed the loop that tell us about the combination in which we have error and added a condition to append the combination if it does not exist in current_phase_split_id (which is from host)

This is how the workflow looks like for validate_challenge_phase_splits

image

Its somewhat similarly for validate_challenge_phase

### DIFF
--- a/apps/challenges/challenge_config_utils.py
+++ b/apps/challenges/challenge_config_utils.py
@@ -824,7 +824,6 @@ class ValidateChallengeConfigUtil:
                 ].format(combination[0], combination[1], combination[2])
                 self.error_messages.append(message)
 
-        challenge_phase_split_uuids = []
         if challenge_phase_splits:
             phase_split = 1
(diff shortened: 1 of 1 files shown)


gautamjajoo:
@Harshit28j We should remove (1,2,2) instead of appending. Thoughts?

PR #4399, opened 2024-07-24
Fix Swagger API

Harshit28j:
This PR fixes the Swagger API

I have verified and tested the changes by rebuilding the Docker images. The API is now working as expected.

image

### DIFF
--- a/docker/dev/django/Dockerfile
+++ b/docker/dev/django/Dockerfile
@@ -8,8 +8,4 @@ WORKDIR /code
 ADD requirements/* /code/
 RUN pip install -r dev.txt
 
-RUN pip install --upgrade pip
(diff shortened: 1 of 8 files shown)


docker/dev/django/Dockerfile
```diff
@@ -8,4 +8,8 @@ WORKDIR /code
 ADD requirements/* /code/
 RUN pip install -r dev.txt
 
+RUN pip install --upgrade pip
```
gautamjajoo, line 11: Do we need this?

gautamjajoo: We can upgrade the pip version so that we don't need to run this command everytime

RishabhJain2018:
Closing this for now!

PR #4419, opened 2024-08-31
Add feature to disable private submission

Harshit28j:
image

### DIFF
--- a/apps/challenges/aws_utils.py
+++ b/apps/challenges/aws_utils.py
@@ -1665,7 +1665,7 @@ def create_eks_cluster(challenge):
         try:
             response = client.create_cluster(
                 name=cluster_name,
-                version="1.23",
+                version="1.29",
                 roleArn=cluster_meta["EKS_CLUSTER_ROLE_ARN"],
                 resourcesVpcConfig={
                     "subnetIds": [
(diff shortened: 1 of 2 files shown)


apps/challenges/views.py
```diff
@@ -4186,6 +4186,7 @@ def create_or_update_github_challenge(request, challenge_host_team_pk):
                         "evaluation_script": files[
                             "challenge_evaluation_script_file"
                         ],
+                        "disable_private_submission": request.data.get('disable_private_submission')
```
gchhablani, line 4189: Probably don't need this.

gchhablani (commented):
Great PR!

In submission.html, make sure that this thing is disabled when the private submissions are disabled. i.e. don't give this option during submission.

image

In my-submission page, there is a bunch of logic here which shows a column to make an old/existing submission public or private. That column should only be visible when `disable_private_submission` is False:

image

Also, make sure to understand the logic here before making changes, will help with a sanity check.

gchhablani:
@Harshit28j Can you share screenshot of the my submissions page too?

SSHurtado:
Hello all,

We’re hosting a challenge at the end of this month, and this feature would make a big difference. In past editions, everyone kept their submissions private, which left the leaderboard _outdated_ and _discouraged competition_.

Disabling private submissions will keep the leaderboard active and make the event more engaging. **Could you help speed up the review of this pull request?** It’s exactly what we need for this challenge!

Thanks a lot!

PR #4498, opened 2025-03-19
Fix #4497: Make worker and statsd optional in development setup

ilyasd3:
## Summary
- Updated `docker-compose.yml` to use profiles for `worker` and `statsd-exporter`
- Updated `README.md` to document how to start optional containers using profiles
- Ensures a lightweight development setup with only necessary containers running

## Changes Introduced
- The `worker` and `statsd-exporter` containers are now optional in development.
- The **README.md** has been updated to explain this.

## Testing
- Verified that `docker-compose up` only starts required services.
- Verified that `worker` starts correctly with `--profile worker`.
- Verified that `statsd-exporter` starts correctly with `--profile statsd`.
- Checked that both services work when running together.

## Bug Number
Fixes #4497

PR #4499, opened 2025-03-20
Updated Dockerfiles to support ARM64 architecture 

Zahed-Riyaz:
-**AMD64-Specific Stage (amd64)**
Installs Google Chrome and libxss1 (as it’s available on amd64).
This stage is intended for environments requiring Chrome (e.g., end-to-end testing).

**ARM64-Specific Stage (arm64)**
Skips Chrome installation because Google does not provide ARM64 builds for Chrome in this setup.
Creates a placeholder /usr/bin/google-chrome that logs a warning and exits gracefully on ARM64 systems.
This ensures that builds don’t fail on ARM64 and still have a consistent environment for non-Chrome-dependent tasks.
Final Stage

Uses ${TARGETARCH} to automatically select the appropriate architecture stage. Defaults to amd64 if TARGETARCH is not set.


RishabhJain2018:
Hey @Zahed-Riyaz , Why did you close this PR?

Zahed-Riyaz:
Hi @RishabhJain2018  so I accidentally included the resolution for issue #4498  into it which I wasn't assigned, so I'm going to modify the PR to change only the dockerfile for nodejs

RishabhJain2018:
Also, please don't change the production/staging config for now. Just the development config.

Zahed-Riyaz:
Got it, thanks for the clarification

PR #4500, opened 2025-03-21
Update nodejs dockerfile

ilyasd3:
## What does this PR do?

- Refactored `docker/dev/nodejs/Dockerfile` to support both ARM64 and AMD64 platforms.

## Why is this needed?

- On M1/M2/M3 MacBooks (ARM64), the Docker build fails because `google-chrome-stable` cannot be installed using `apt`.
- This fix ensures the build doesn't fail on ARM while still enabling Chrome functionality on AMD64 machines.

## Related to

#4498

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,25 @@ The steps are:
     docker-compose up --build
     ```
 
+    By default, this starts only the required services (`db`, `sqs`, and `django`).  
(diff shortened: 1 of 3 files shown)


RishabhJain2018:
Hey @ilyasd3 , This is similar to your previous PR. Closing it.

PR #4515, opened 2025-03-26
Added Google auth and Github Auth  UI  including Tab section  of all challenges 

devjayswal:
@RishabhJain2018  Hello sir  this is the PR raised for the issue google auth and github auth 
since our web is associated with github we can just use google and Github auth further more 

since  all the challenge in a page seems cluter and user needs to scroll much for reaching the bottom of page to  find past or ongoing challenges 

implementing Google Auth And Github Auth for authentication 

image

Image


RishabhJain2018:
Hi @devjayswal , Thanks for the PR. We don't need the google Auth / Github Auth for now. Also, for the other feature someone is already assigned for it. Hence, closing this PR.

PR #4533, opened 2025-03-30
[DOCS]: update remote challenge evaluation template link

Samyak-Jayaram:
The evaluation scripts md had an outdated link which resulted in a 404 not found. It is now updated with proper link.

The original link lead to a single file which no longer exists as it is now a couple of scripts in the same folder.


RishabhJain2018:
Closing since it is similar to #4536

PR #4536, opened 2025-03-30
Docs: Update link of remote challenge evaluation in Readme

### LINKED ISSUE #3358: [Backend] Returns with code 500 upon submission of wrong URL
Backend returns with code **500 (Internal server error** when a file with wrong extension is submitted for a challenge. It should return with **400 (Bad request)**

Example: Submitting a `.py` file URL to a challenge that doesn't accept `.py` files

If you'd like to report a bug, please include the following information:
- Expected behaviour
It should return with a `400 Bad Request`

- Observed behaviour
Server gives a `500 Internal Server Error`

- Steps to reproduce the problem
  1. Choose any challenge
  2. Submit a file URL with `.py` (or `.cpp`, etc) extension
  3. Observe the error

Upon submitting a file URL with `.py` extension
image

Samyak-Jayaram:
Returning appropriate http response codes from backend according to user actions.


RishabhJain2018 (approved):
LGTM :) Thank you for opening the PR.

PR #4582, opened 2025-04-11
fix: drf-yasg dump function

TheArchitect19:
The issue was caused by drf-yasg using the deprecated ruamel.yaml.dump() method, which was removed in ruamel.yaml>=0.18, leading to a 500 error when rendering the API docs.

[Screencast from 04-13-2025 04:05:03 AM.webm](https://github.com/user-attachments/assets/c4c63743-0ce6-41c3-9a3e-6da350f4cc8e)


requirements/common.txt
```diff
@@ -31,5 +31,6 @@ PyJWT==2.1.0
 PyYaml==5.1
 proc==1.0
 rstr==2.2.6
+ruamel.yaml<0.18
```
RishabhJain2018, line 34: What if we upgrade the `drf-yasg` version?

RishabhJain2018 (changes requested):

requirements/common.txt
TheArchitect19: yes we can do it,
but this package is not actively maintained, and also there is no official release that is solving this problem

ref: https://sourceforge.net/p/ruamel-yaml/code/ci/default/tree/CHANGES#l456

so the issue may persist if we upgrade the package

RishabhJain2018: Oh I see..Can you please check what are the other solutions to host the API docs?

TheArchitect19: okay, let me check

evalai/urls.py
```diff
@@ -22,24 +22,31 @@
...
 handler500 = "web.views.internal_server_error"
 
-swagger_api_info = openapi.Info(
-    title="EvalAI API",
-    default_version="v1",
-    description="EvalAI Documentation",
-    contact=openapi.Contact(email="team@cloudcv.org"),
-    license=openapi.License(name="BSD License"),
-)
+# swagger_api_info = OpenApiInfo(
```
RishabhJain2018, line 39: Please don't comment it, let's remove it since you are adding a different package.

evalai/urls.py
```diff
@@ -80,16 +87,29 @@
...
-        r"^api/docs/docs(?P<format>\.json|\.yaml)$",
-        schema_view.without_ui(cache_timeout=0),
-        name="schema-yaml",
-    ),
-    url(
-        r"^api/docs/$",
-        schema_view.with_ui("redoc", cache_timeout=0),
-        name="schema-redoc",
-    ),
+    # url(
```
RishabhJain2018, line 90: Please remove all these comments.

RishabhJain2018 (changes requested):

requirements/common.txt
```diff
@@ -31,5 +32,6 @@ PyJWT==2.1.0
...
```
RishabhJain2018, line 35: Should we remove this now?

settings/common.py
```diff
@@ -322,6 +324,14 @@
     },
 }
 
+SPECTACULAR_SETTINGS = {
```
RishabhJain2018, line 327: `SPECTACULAR_SETTINGS` -> `DRF_SPECTACULAR_SETTINGS`

evalai/urls.py
```diff
@@ -22,24 +22,24 @@
 from django.views.generic.base import TemplateView
 from drf_yasg import openapi
```
RishabhJain2018, line 23: Should we remove this and all its related components since we are adding a new package?

evalai/urls.py
```diff
@@ -22,24 +22,24 @@
...
 from drf_yasg.views import get_schema_view
+from drf_spectacular.views import (
+    SpectacularAPIView,
+    SpectacularRedocView,
+    SpectacularSwaggerView,
+)
+# from drf_spectacular.openapi import OpenApiInfo
```
RishabhJain2018, line 30: Can we remove this commented code?

evalai/urls.py
```diff
@@ -22,24 +22,24 @@
...
 from rest_framework import permissions
 from rest_framework_expiring_authtoken.views import obtain_expiring_auth_token
 from web import views
 
+
```
RishabhJain2018, line 35: Please remove this line.

RishabhJain2018 (changes requested):

requirements/common.txt
```diff
@@ -18,6 +18,7 @@ django-rest-auth[with_social]==0.9.5
 django-ses==1.0.3
 docker-compose==1.27.4
 drfdocs==0.0.11
+drf-spectacular>=0.23.1,<1.0
 drf-yasg==1.17.0
```
RishabhJain2018, line 22: Let's remove this from requirements?

scripts/workers/submission_worker.py
```diff
@@ -30,7 +30,7 @@
 django.setup()  # isort:skip noqa:E402
 # fmt: on
 
-from challenges.models import ChallengePhase  # noqa:E402
+from challenges.models import ChallengePhase  # noqa:E402; noqa:E402
```
RishabhJain2018, line 33: We already have #E402.

settings/common.py
```diff
@@ -75,6 +75,7 @@
     "rest_framework_expiring_authtoken",
     "drf_yasg",
```
RishabhJain2018, line 76: Let's remove this?

RishabhJain2018 (changes requested):

scripts/workers/submission_worker.py
RishabhJain2018: @TheArchitect19 Please remove this duplocate value.

RishabhJain2018 (changes requested):

RishabhJain2018 (approved):

PR #4607, opened 2025-04-20
fixed countdowntimer function which was showing negative values

### LINKED ISSUE #4594: BUG: Negative Time Displayed in Submission Countdown Timer
**Description:**

We encountered a UI bug where the countdown timer for the next set of submissions displays a negative value for seconds. This happened right after hitting the submission limit for the day.

**Steps to Reproduce:**
1. Submit until the daily limit is reached.
2. Observe the countdown timer for when submissions will be available again.

**Expected Behavior:**
The countdown timer should not display negative values. It should reset cleanly to `0 seconds` or roll over appropriately.

**Actual Behavior:**
The seconds field briefly or persistently displays a negative number, e.g., `-1 seconds`.

**Screenshot:**
Image

Sekhar-Kumar-Dash:
Fix #4594 

This PR solves the issue of negative values in the countdowntimer

the changed code now checks for
1-checks if remainingTime is already at or below zero at the beginning of the function
2-properly handles the case when time runs out by setting all display values to 00
3-avoids the situation where remainingTime could go negative

### DIFF
--- a/frontend_v2/src/app/components/challenge/challengesubmit/challengesubmit.component.ts
+++ b/frontend_v2/src/app/components/challenge/challengesubmit/challengesubmit.component.ts
@@ -298,45 +298,60 @@ export class ChallengesubmitComponent implements OnInit {
    * @param SELF current context
    * @param eachPhase particular phase of a challenge
    */
-  countDownTimer(SELF, eachPhase) {
-    if (!SELF.isClockStarted) {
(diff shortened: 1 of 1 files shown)


RishabhJain2018:
Hey @Sekhar-Kumar-Dash , We are not using `frontend_v2`. Please make the changes in `frontend` folder.

Issue #4357, opened 2024-05-02
BugFix: Show all information on challenge card


gautamjajoo:
### Expected behaviour
End date should also be visible. All the information should be visible on the card.

### Observed behaviour
Information is hidden due to increase in the number of tags.

<img width="299" alt="Screenshot 2024-05-02 at 12 26 09 PM" src="https://github.com/Cloud-CV/EvalAI/assets/24366008/70f67845-d97a-4b28-acd2-99b8f16b329e">

### Deliverable:

- [ ] Truncate Tags with 'Show More': Display a limited number of tags and provide a "Show More" button to reveal additional tags.

Plomo-02:
hello, Is it possible being assigned to this issue?

Aditi571:
@gautamjajoo Can you assign this issue to me?

YdvAkash:
Reduce Tag Size or Font Size:
Decrease the size of the tags or the font size to make sure all information fits within the card.

Multi-line Tags:
Allow the tags to wrap to multiple lines instead of a single line to create more space.

Scroll View for Tags:
Implement a scrollable view for the tags so that users can scroll through them without taking up too much space.

Expand Card Height:
Increase the height of the card to accommodate all the information and tags without hiding any content.

Collapsible Tags:
Make the tags collapsible, with an option to expand and view all tags if necessary.

huntermarchi:
@gautamjajoo please assign the issue to me i can fix it.

omsaha01:
Can I get this issue assigned @gautamjajoo ?

chandanSahoo-cs:
Can I be assigned this issue

harikrishna-au:
Hi @gautamjajoo,

 I wanted to follow up regarding the issue I raised earlier. Could you please confirm if it has been resolved?

Additionally, I was able to successfully run the repository locally, but I am facing an issue where I am unable to log in using different accounts. Could you please assist me with this?

Lastly, I would greatly appreciate it if you could assign me some "Good First Issues" so I can contribute more effectively to the project.

DipayanDasgupta:
Has any work been done on this issue? Is it still open? @gautamjajoo ?

DipayanDasgupta:
can I be assigned this issue @gautamjajoo

Divya-Raj7488:
[...]

PR #4623, opened 2025-04-29
Modified the challenge card 

akanshajain231999:
Modified the challenge card  by setting 2 tags only and removing GMT timezone from the card.
FrontendUI


akanshajain231999:
Fix #4357

RishabhJain2018 (approved):

PR #4625, opened 2025-04-29
Added github icon on hosted challenge page

akanshajain231999:
#FIX 4591
This PR is about adding github icon on hosted challenge page which will link to challenge repository.
Github Icon

Changes Implemented:
Introduced a GitHub icon on each challenge page that links to the challenge's GitHub repository.
The icon is only visible to the challenge host and host team members.
The icon is conditionally rendered only if there is a valid GitHub repository for the challenge


RishabhJain2018 (changes requested):
Hey @akanshajain231999 , Please fix the test cases.

RishabhJain2018 (approved):

PR #4745, opened 2025-07-16
Add github_branch to models

Zahed-Riyaz:


apps/challenges/migrations/0113_add_github_branch_field_and_unique_constraint.py
```diff
@@ -28,9 +54,14 @@ class Migration(migrations.Migration):
...
             field=models.CharField(
-                blank=True, default="", max_length=200, null=True
+                blank=True, default="challenge", max_length=200, null=True
             ),
         ),
+        # Data migration to populate existing records
+        migrations.RunPython(
+            populate_github_branch_default,
+            reverse_populate_github_branch_default,
+        ),
```
gchhablani, line 64: Why are we running populate github branch default and then reversing it??

gchhablani, line 64: Is this the best way to handle this scenario?

scripts/migration/populate_github_branch.py
```diff
@@ -0,0 +1,79 @@
...
+    Populate existing challenges with empty github_branch fields to use "challenge" as default.
+    """
+    print("Starting github_branch field population...")
+
+    challenges_to_update = (
+        Challenge.objects.filter(github_repository__isnull=False)
+        .exclude(github_repository="")
+        .filter(
+            models.Q(github_branch__isnull=True) | models.Q(github_branch="")
+        )
```
gchhablani, line 27: If we already have this script to populate and back-fill, why do we need to the populate/reverse populate methods at all?

apps/challenges/challenge_config_utils.py
```diff
@@ -364,6 +365,23 @@ def __init__(
...
+    def validate_github_branch_format(self):
+        """
+        Ensure the github branch name matches challenge-<year>-<version>
+        """
+        branch = self.request.data.get(
+            "GITHUB_BRANCH_NAME"
+        ) or self.request.data.get("BRANCH_NAME")
+        if not branch:
+            branch = "challenge"
+        pattern = r"^challenge-\d{4}-[a-zA-Z0-9]+$"
```
gchhablani, line 377: Will this also be okay with just "challenge" (default) if passed?

tests/unit/challenges/test_views.py
```diff
@@ -6020,14 +6020,13 @@ def test_create_challenge_using_github_without_branch_name(self):
             self.assertEqual(response.status_code, 201)
             self.assertEqual(response.json(), expected)
 
-        # Verify github_branch defaults to empty string when not provided
+        # Verify github_branch defaults to "challenge" when not provided
```
gchhablani, line 6023: Please also add when the branch name is "challenge", when branch name is "xyzabc", and when branch name is "challenge-2025-v2" (whatever format you have chosen).

gchhablani (commented):
Left some initial comments.

gchhablani:
Please also add back the github success test later on which we will be removing at master.

apps/challenges/migrations/0113_add_github_branch_field_and_unique_constraint.py
Zahed-Riyaz: The migration itself was written to handle the populating at first, but now that a script is explicitly written for it, it might be best to remove it from the migration entirely.

PR #4852, opened 2025-12-28
Fix #4848:  Enforce LF line endings for shell scripts to prevent Docker startup issues on Windows

### LINKED ISSUE #4848: Docker startup fails due to interactive input in seed command
### Problem
When running EvalAI using Docker, the Django container fails to start due to an `EOFError` raised from an interactive `input()` call in the seed command.

This happens because `python manage.py seed` is executed automatically during container startup in a non-interactive environment.

### Steps to Reproduce
1. Clone the EvalAI repository
2. Run `docker compose up --build`
3. Observe the Django container exit with:

### Root Cause
The seed script prompts for user input when existing database data is detected. Docker containers are non-interactive, so `input()` raises `EOFError`.

### Expected Behavior
The seed command should not require interactive input in Docker or CI environments.

### Suggested Fix
Skip database reseeding safely in non-interactive environments instead of prompting for input or crashing.

sayan24kole:
This PR enforces LF line endings for shell scripts in the repository via .gitattributes to prevent Docker startup failures on Windows.

The original issue caused containers to exit with code 127:

    /usr/bin/env: ‘bash\r’: No such file or directory

By adding .gitattributes and normalizing shell scripts, this PR improves cross-platform Docker compatibility and avoids the CRLF-related errors during container startup.

Fixes #4848

### DIFF
--- a/evalai/.gitattributes
+++ b/evalai/.gitattributes
@@ -1,0 +1,1 @@
+*.sh text eol=lf


sayan24kole:
Hi maintainers, 

This PR only adds a `.gitattributes` file to enforce LF line endings for shell scripts. It does not modify any script logic or Python code. 

The Travis CI build failure seems unrelated to the changes in this PR. Please let me know if any additional updates are required on my side. 

Thanks!

PR #4855, opened 2026-01-04
Fix #4330: Auth token not updating in UI after re-login

### LINKED ISSUE #4330: Frontend: Auth Token doesn't get updated on UI if the user logs on after a long time.

Thebinary110:
### Summary
Fix #4330 by resolving a frontend state synchronization issue where the auth token
was not updated in the UI after a user re-logs in, especially after long inactivity.

### Problem
When a user logs in again after a long time, the backend returns a new auth token,
but the UI continues to display and use the old token due to stale in-memory state.

### Fix
- Sync the new auth token into in-memory state on login
- Prefer in-memory token when fetching auth token to avoid stale values
- Clear in-memory token on logout to prevent reuse
- Ensures Profile page and CLI token display show the updated token immediately

### Scope
- Frontend-only change
- No backend or API changes
- Minimal and safe update

### DIFF
--- a/frontend_v2/src/app/components/auth/login/login.component.ts
+++ b/frontend_v2/src/app/components/auth/login/login.component.ts
@@ -98,11 +98,36 @@ export class LoginComponent implements OnInit {
 
     this.apiService.postUrl(this.endpointsService.loginURL(), LOGIN_BODY).subscribe(
       (data) => {
+        /**
+         * FIX:
+         * Explicitly clear old token and propagate new token
+         * to avoid stale auth state after long inactivity.
+         */
+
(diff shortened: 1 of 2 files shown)


Thebinary110:
cc: @gautamjajoo @RishabhJain2018 please review and confirm is it working now

PR #4864, opened 2026-01-06
Use permission-based authorization for admin email endpoint

### LINKED ISSUE #4858: Security: Replace inline superuser check with permission-based access control in bulk email endpoint
The **notify_users_about_challenge** endpoint currently relies on inline **request.user.is_superuser** checks for authorization.

This approach is fragile and risks accidental misuse if the view is reused or middleware behavior changes.

Proposed solution:

Use DRF permission classes (e.g. IsAdminUser or a custom permission)

Add endpoint-specific throttling

Add audit logging for bulk email triggers

If this approach is acceptable, I’d be happy to open a PR.

Riddhima8:
Fix #4858 : Use permission-based authorization for admin email endpoint
### Summary
This PR replaces inline `request.user.is_superuser` authorization logic in
`notify_users_about_challenge` with framework-level permission-based access
control using Django REST Framework permissions.

### Motivation
Inline authorization checks are fragile and can be accidentally bypassed if the
view is reused or modified. Permission decorators ensure authorization is
enforced declaratively before view execution.

### Scope of change
- No functional behavior change for authorized users
- Only authorization logic updated
- No refactors or unrelated changes

### Related Issue
Fixes #4858

### DIFF
--- a/apps/web/views.py
+++ b/apps/web/views.py
@@ -7,16 +7,13 @@ from django.conf import settings
 from django.contrib.auth.models import User
 from django.core.mail import EmailMessage
 from django.shortcuts import render
+from django.contrib.auth.decorators import login_required
+from django.contrib.admin.views.decorators import staff_member_required
+
 from rest_framework import permissions, status
-from rest_framework.decorators import (
-    api_view,
-    permission_classes,
-    throttle_classes,
-)
+from rest_framework.decorators import api_view, throttle_classes, permission_classes
 from rest_framework.response import Response
 from rest_framework.throttling import AnonRateThrottle
-from rest_framework.permissions import IsAdminUser
-from rest_framework.decorators import api_view, permission_classes
 
 from .models import Subscribers, Team
 from .serializers import ContactSerializer, SubscribeSerializer, TeamSerializer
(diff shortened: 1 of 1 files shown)


Riddhima8:
Hi maintainers,

I wanted to add some context and also check if there’s anything I should adjust on this PR.

This endpoint renders HTML templates and the existing tests expect HTML responses with a 200 status code. Using DRF permission decorators caused the request to be intercepted earlier, resulting in 401/403 responses and breaking those expectations.

To preserve the existing behavior while keeping admin-only access, I switched this view to Django-native authentication decorators (login_required and staff_member_required) instead of DRF permissions.

Please let me know if there’s any concern with this approach or if you’d prefer a different direction (e.g. test changes or an alternative auth pattern). Happy to update accordingly.

Riddhima8:
After investigating the CI failures and test expectations, I realized this endpoint intentionally uses explicit checks to preserve HTML response behavior (no redirects) for unauthenticated and non-superuser access.

Since the original implementation already matches this contract, I’m closing the PR to avoid introducing a no-op change.

Thanks for the maintainers’ time — this helped clarify the intended behavior of this view.

PR #4866, opened 2026-01-07
Fix unique subscriber email

Riddhima8:
Fix #4865: Enforce unique constraint on subscriber email

### Summary
Adds a database-level uniqueness constraint to the `Subscribers.email` field.

### Problem
Although the subscription API checks for existing emails at the application layer, the absence of a database-level unique constraint can still result in duplicate subscriber records due to race conditions, concurrent requests, admin panel usage, or direct database inserts.

### Solution
- Marked the `email` field as `unique=True` in the `Subscribers` model to enforce uniqueness at the database level.

### Notes
Migration not included since Django is not set up locally. Happy to add it if preferred.

Fixes #4865

Issue #4865, opened 2026-01-06
Bug: Subscribers email field does not enforce uniqueness at database level


Riddhima8:
### Description
The `Subscribers` model does not enforce a database-level uniqueness constraint on the `email` field.

Although the subscription API checks for existing emails at the application layer, the absence of a uniqueness constraint at the database level can still result in duplicate subscriber records.

### Why this is an issue
Application-level checks alone are not sufficient to guarantee data integrity, especially in cases of:
- Concurrent requests
- Race conditions
- Direct database inserts
- Admin panel usage

This can lead to duplicate entries for the same email address.

### Affected code
```python
class Subscribers(TimeStampedModel):
    email = models.EmailField(max_length=70)

aviralgarg05:
I would like to solve the issue. Please assign me the issue

neo0007777:
Hey i want to work on this issue

PR #4869, opened 2026-01-12
resolved subscribers email uniqueness issue

yashdodwani:
Fix #4866 (https://github.com/Cloud-CV/EvalAI/issues/4865) Enforce unique constraint on subscriber email

## Summary

This PR fixes the issue where the `Subscribers` model lacked a database-level uniqueness constraint on the `email` field, which could lead to duplicate subscriber records.

## Problem

The `Subscribers` model only had application-level validation for duplicate emails in the subscription API endpoint (`apps/web/views.py`). This approach is vulnerable to:
- **Race conditions** when multiple requests arrive simultaneously
- **Concurrent database inserts** that bypass application logic
- **Direct database operations** (SQL, admin panel, data migrations)
- **Bulk imports** that don't go through the API

### 1. Model Change
Added `unique=True` constraint to the `email` field in the `Subscribers` model:

**File:** `apps/web/models.py`
```python
email = models.EmailField(max_length=70, unique=True)
```

### 2. Database Migration
Created migration `0011_add_unique_constraint_to_subscribers_email.py` to apply the constraint at the database level.

**File:** `apps/web/migrations/0011_add_unique_constraint_to_subscribers_email.py`

### 3. Test Coverage
Added a test to verify the unique constraint enforcement:

**File:** `tests/unit/web/test_models.py`
```python
def test_email_unique_constraint(self):
    """Test that duplicate email addresses raise IntegrityError"""
    with self.assertRaises(IntegrityError):
        Subscribers.objects.create(email="subscriber@domain.com")
```

### 4. Duplicate Check Script
Created a utility script to check for existing duplicates before migration:

**File:** `scripts/check_subscriber_duplicates.py`

This script helps identify any existing duplicate records that would prevent the migration from running successfully.

## Files Changed

- `apps/web/models.py` - Added `unique=True` to email field
- `apps/web/migrations/0011_add_unique_constraint_to_subscribers_email.py` - New migration
- `tests/unit/web/test_models.py` - Added test for uniqueness constraint
- `scripts/check_subscriber_duplicates.py` - New utility script

### Run the test
```bash
pytest tests/unit/web/test_models.py::SubscribersTestCase::test_email_unique_constraint -v
```

### Check for duplicates before migration
```bash
python scripts/check_subscriber_duplicates.py
```

### Apply the migration
```bash
python manage.py migrate web
```

Or with Docker:
```bash
docker-compose run --rm django python manage.py migrate web
```

## Notes

- The existing application-level check in the subscription API remains valuable for providing user-friendly error messages
- The database constraint serves as a final safety net to guarantee data integrity
- If existing duplicates are found, they must be cleaned up before the migration can be applied

## Backward Compatibility

This change enforces stricter data integrity but should not break existing functionality:
- ✅ API behavior remains the same (application-level check still returns the same error)
- ✅ No breaking changes to the API interface
- ⚠️ Migration requires no existing duplicates in the database

## Resolves

Fixes the issue: "Subscribers model does not enforce a database-level uniqueness constraint on the email field"


yashdodwani:
@vik05h i have run those commands but still getting the same error

OS: linux

PR #4918, opened 2026-01-20
[Frontend] Disable challenge submissions for free tier challenges

gchhablani:
This PR restricts submission capabilities to users on a paid tier. Users on the free payment tier will no longer see submission forms or instructions and will instead be prompted to upgrade.

**Technical Changes**
**Access Control**: Updated `ng-if` logic across all submission-related sections to exclude users where `challenge.payment_tier` is 'free'.

**UI Restrictions**: Applied these conditions to:

- Docker-based submission instructions and command snippets.

- The Docker submission limits/remaining counts table.

- The standard file upload and CLI submission forms.

**User Feedback**: Added a conditional notification section that displays a "Submission Unavailable" message specifically for free-tier users to explain the missing functionality.

Screenshot example for free tier challenge:
Screenshot 2026-01-19 at 5 30 56 PM

Screenshot example for paid tier challenge:
Screenshot 2026-01-19 at 5 32 14 PM

### DIFF
--- a/apps/challenges/models.py
+++ b/apps/challenges/models.py
@@ -273,6 +273,26 @@ class Challenge(TimeStampedModel):
             return True
         return False
 
+    @property
+    def payment_tier(self):
+        """Returns the payment tier name for this challenge"""
+        try:
+            from payments.models import ChallengePaymentTier
+            payment_tier_mapping = ChallengePaymentTier.objects.select_related('payment_tier').get(challenge=self)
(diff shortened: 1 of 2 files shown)


RishabhJain2018 (changes requested):
Can we please create new app to manage the Payments and separate models for payments? We can join it on per challenge model. This model is not scalable in future.

PR #4920, opened 2026-01-20
Disable resume/restart for free-tier challenges

gchhablani:
This PR disables the resume/restart operations for free-tier challenges by both participants and hosts on the front-end.

It extends payment tier restrictions to the "All Submissions" view.
- Hid the Rerun submission column and action button for free tier users.
- Hid the Resume submission column and action button for free tier users (even if they are challenge hosts).

Verified manually by changing a challenge from free-tier to core-tier:

Free:
Screenshot 2026-01-19 at 11 55 38 PM
Core:
Screenshot 2026-01-19 at 11 56 02 PM

### DIFF
--- a/apps/challenges/serializers.py
+++ b/apps/challenges/serializers.py
@@ -96,6 +96,7 @@ class ChallengeSerializer(serializers.ModelSerializer):
             "sqs_retention_period",
             "github_repository",
             "github_branch",
+            "payment_tier",
         )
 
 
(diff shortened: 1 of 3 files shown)


RishabhJain2018 (changes requested):
Hey @gchhablani , Can we please create new app to manage the Payments and separate models for payments? We can join it on per challenge model. This model is not scalable in future.

PR #4966, opened 2026-01-29
Fix: harden submission analytics endpoints against QuerySet edge cases

Youssef-SH:
This PR hardens submission analytics endpoints by removing unsafe QuerySet usage patterns while preserving existing response behavior.

Key changes:

- Replaces QuerySet truthiness checks and direct indexing with explicit .exists() and .first() calls to avoid potential TypeError and IndexError paths.
- Removes broad exception handling that was masking genuine server-side errors and returning misleading 400 responses.
- Keeps responses unchanged for valid requests while making failure modes explicit and debuggable.

No functional or API-level changes are introduced.

### DIFF
--- a/apps/analytics/views.py
+++ b/apps/analytics/views.py
@@ -164,15 +164,11 @@ def get_challenge_phase_submission_count_by_team(
     challenge_phase_submission_count = ChallengePhaseSubmissionCount(
         participant_team_submissions, challenge_phase.pk
     )
-    try:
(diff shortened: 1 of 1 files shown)


Youssef-SH:
CI note:
The failure occurs during Docker image build (apt dependency resolution) and is unrelated to this change.
This PR only touches application-level Python code and does not modify CI or build configuration.

PR #5069, opened 2026-04-17
Backend: Fix slow DB query in remaining_submissions endpoint

akanshajain231999:
## Problem

Sentry flagged \`/api/jobs/{challenge_pk}/remaining_submissions/\` as a slow endpoint — the annotated queryset was taking **1.21 s out of 1.32 s** total request time (92% of the request).

The root cause was a redundant self-join on \`challenge_phase\` generated by Django ORM.

### The bad SQL (simplified)

\`\`\`sql
FROM challenge_phase
LEFT OUTER JOIN submission ON (challenge_phase.id = submission.challenge_phase_id)
LEFT OUTER JOIN challenge_phase T4 ON (submission.challenge_phase_id = T4.id)  -- redundant!
WHERE challenge_phase.challenge_id = %s
\`\`\`

\`T4\` is an alias for the same \`challenge_phase\` table. It was introduced because \`submissions__challenge_phase__challenge=challenge_pk\` was included inside the \`COUNT(...) FILTER\` annotation. Django ORM traversed \`submission → challenge_phase → challenge\` and emitted an extra join — even though \`challenge_phases\` was already filtered to that challenge in the outer \`WHERE\` clause, making \`T4.challenge_id = challenge_pk\` always true.

### 1. Remove the redundant condition from \`submission_filter\` (\`apps/jobs/views.py\`)

The \`challenge_phases\` queryset is already scoped with \`.filter(challenge=challenge)\`, so the LEFT JOIN on \`submission.challenge_phase_id\` naturally limits submissions to the correct challenge. The removed condition was provably redundant.

### 2. Add composite indexes on \`submission\` (\`apps/jobs/models.py\` + migration \`0027\`)

Covering indexes on \`(challenge_phase, participant_team, status)\` and \`(challenge_phase, participant_team, submitted_at)\` to speed up the three \`COUNT(...) FILTER\` aggregations.

## Test

Added \`test_submissions_from_other_challenge_not_counted\` to \`GetRemainingSubmissionTest\` — creates a second challenge, submits to it from the same participant team, and asserts the original challenge's counts are unaffected. This guards the correctness invariant that JOIN-based scoping is sufficient.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


RishabhJain2018 (approved):
LGTM!

Issue #5225, opened 2026-09-14
Submission with `is_public=True` returns HTTP 500 instead of being accepted - apps/jobs/views.py:365


Shivansh1205:
Submitting with `is_public=True` — Python's own spelling of the boolean — returns HTTP 500. Only the lowercase JSON spellings `true`/`false` are accepted.

## Offending code

`challenge_submission()` parses the form-encoded `is_public` flag with `json.loads()`:

https://github.com/Cloud-CV/EvalAI/blob/f8aff86a7e613c61fae3e4b0e5c0382fe3b52a25/apps/jobs/views.py#L360-L365

```python
        if request.data.get("is_public") is None:
            request.data["is_public"] = (
                True if challenge_phase.is_submission_public else False
            )
        else:
            request.data["is_public"] = json.loads(request.data["is_public"])
```

The same line appears in `get_submission_file_presigned_url()`:

https://github.com/Cloud-CV/EvalAI/blob/f8aff86a7e613c61fae3e4b0e5c0382fe3b52a25/apps/jobs/views.py#L3050

## Why it is wrong

`is_public` is a `BooleanField` on the model, and these are form-encoded (`multipart`) requests, so the value arrives as a **string**. `json.loads()` only accepts the JSON spellings, which is a much narrower set than what clients actually send:

| value sent      | `json.loads()` result          |
|-----------------|--------------------------------|
| `"true"`        | `True`                         |
| `"false"`       | `False`                        |
| `"True"`        | **JSONDecodeError → HTTP 500** |
| `"False"`       | **JSONDecodeError → HTTP 500** |
| `"on"` (HTML checkbox) | **JSONDecodeError → HTTP 500** |
| `"yes"`         | **JSONDecodeError → HTTP 500** |
| `""`            | **JSONDecodeError → HTTP 500** |
| `True` (JSON body, real bool) | **TypeError → HTTP 500** |

`json.JSONDecodeError` subclasses `ValueError`, not `APIException`, and the call sits outside every `try` block in both views. DRF's `exception_handler` returns `None` for it, so it re-raises into Django and the request is answered with HTTP 500:

```
>>> from rest_framework.views import exception_handler
>>> import json
>>> try: json.loads("True")
... except Exception as e: exc = e
>>> exception_handler(exc, {})
None      # -> re-raised -> HTTP 500
```

[...]

Issue #4830, opened 2025-12-07
Bug: Race Condition in Public Submission Toggling Allows Multiple Public Submissions


THE-Amrit-mahto-05:
## Description 
When is_restricted_to_select_one_submission is enabled for a challenge phase, multiple submissions can become public simultaneously due to a race condition. This violates the business rule that only one submission should be public at a time.

## Location 
**File:** apps/jobs/views.py 
**Lines:**
 - 342-349 (in challenge_submission function) 
 - 475-485 (in change_submission_data_and_visibility function)
    

## Reproduction 
1. Save the attached test script as test_bug_race_condition.py 
2. Run: python3 test_bug_race_condition.py 
3. Observe the output showing 2 public submissions instead of 1

**Test Output:** 
Initial: 1 public submission(s) Final: 2 public submission(s) - IDs: [2, 3]

## Expected Behavior 
When a challenge phase has is_restricted_to_select_one_submission=True: 
- Only 1 submission should be public at any given time
- When making a new submission public, the existing public submission should be made private first 
- This should work correctly even with concurrent requests
              

## Actual Behavior 
- Multiple submissions can become public simultaneously 
- The race condition occurs because there is no database-level locking 
- Both concurrent requests pass the count() == 1 check at the same time 
- Both proceed to make their submissions public

![Image](https://github.com/user-attachments/assets/48c0b968-7bfd-43c3-acad-6c723494d722)

PR #5228, opened 2026-09-14
Fix #5227: Coerce `is_public` before testing it in submission visibility updates

### LINKED ISSUE #5227: `is_public is True` never matches form-encoded requests, skipping when_made_public and the one-public-submission rule - apps/jobs/views.py:509
`change_submission_data_and_visibility()` compares a request value with `is True`. Form-encoded clients send `"true"` as a **string**, so the check never matches: the submission still becomes public, but `when_made_public` is never stamped and the "only one public submission" rule is silently skipped.

## Offending code

https://github.com/Cloud-CV/EvalAI/blob/f8aff86a7e613c61fae3e4b0e5c0382fe3b52a25/apps/jobs/views.py#L507-L523

```python
    try:
        is_public = request.data["is_public"]
        if is_public is True:
            when_made_public = datetime.datetime.now()
            request.data["when_made_public"] = when_made_public

            submissions_already_public = Submission.objects.filter(...)
            # Make the existing public submission private before making the new
            # submission public
            if (
                challenge_phase.is_restricted_to_select_one_submission
                and is_public
                and submissions_already_public.count() == 1
            ):
```

`is_public` is taken straight from `request.data` with no coercion, then compared by **identity** against `True`.

## Why it is wrong

This is a `PATCH` endpoint with DRF's default parsers, which include `FormParser` and `MultiPartParser`:

```
>>> from rest_framework.settings import api_settings
>>> [p.__name__ for p in api_settings.DEFAULT_PARSER_CLASSES]
['JSONParser', 'FormParser', 'MultiPartParser']
```

Form-encoded and multipart bodies carry no type information, so every value arrives as a **string**. `"true" is True` is `False`, and so is `"True" is True` and `"1" is True`. Only a JSON body carrying a real boolean satisfies the check.

The damaging part is that the request does not fail. The serializer further down applies DRF's `BooleanField` to `is_public`, which happily coerces `"true"` to `True` and saves it — so the submission **does** become public, while the entire block guarded by `is True` was skipped:

| client | `is_public` | becomes public | `when_made_public` stamped | old public demoted |
|---|---|---|---|---|
| JSON body (web UI) | `True` (bool) | ✅ | ✅ | ✅ |
| form-encoded | `"true"` | ✅ | ❌ | ❌ |
| form-encoded | `"True"` | ✅ | ❌ | ❌ |
| form-encoded | `"1"` | ✅ | ❌ | ❌ |

The web UI is unaffected — `challengeCtrl.js` sends `parameters.data = {"is_public": submissionVisibility}` as JSON, a real boolean. This is reached by API and CLI clients.

## Relationship to #4830

#4830 reports that multiple submissions can be public at once in a phase with `is_restricted_to_select_one_submission=True`, and attributes it to a race condition; PR #4831 proposes `select_for_update()` locking at these same lines.

This is a **second, independent cause of the same broken state**, and it needs no concurrency at all — a single form-encoded PATCH reaches it deterministically. `is_public is True` does not appear anywhere in PR #4831's diff, so that locking fix leaves this path live. I mention it because a reviewer looking at #4830 may otherwise assume the locking change closes it entirely.

## Steps to reproduce

1. Create a challenge phase with `is_restricted_to_select_one_submission=True` and one public submission.
2. As a participant, `PATCH` a second submission at
   `/api/jobs/challenge/<challenge_pk>/challenge_phase/<challenge_phase_pk>/submission/<submission_pk>`
   sending **form-encoded** (`multipart/form-data` or `application/x-www-form-urlencoded`) data:
   ```
   is_public=true
   ```
3. The response is HTTP 200 and the second submission is now public.
4. Query the phase: **two** submissions have `is_public=True`, and the new one's `when_made_public` is `NULL`.

Repeating step 2 with a JSON body (`{"is_public": true}`) demotes the first submission correctly, which shows the divergence.

## Expected behaviour

- `is_public` is coerced before being tested, so form-encoded and JSON clients behave identically.
- `when_made_public` is stamped whenever a submission is made public.
- The restrict-one-public branch runs for form-encoded clients too.
- A value that is not a recognised boolean returns HTTP 400 naming `is_public`.

---

I have a branch that coerces the value through DRF's `BooleanField` before the check and adds regression tests; I'll open a PR referencing this issue.

Shivansh1205:
Fixes #5227

## Problem

`change_submission_data_and_visibility()` reads `is_public` straight from `request.data` and compares it with `is True`:

```python
is_public = request.data["is_public"]
if is_public is True:
    when_made_public = datetime.datetime.now()
    ...
```

This is a `PATCH` endpoint with DRF's default parsers (`JSONParser`, `FormParser`, `MultiPartParser`). Form-encoded bodies carry no type information, so the value arrives as the **string** `"true"` — and `"true" is True` is `False`.

What makes this quiet rather than loud is that the request does not fail. The serializer downstream applies DRF's `BooleanField`, which coerces `"true"` to `True` and saves it. So the submission **becomes public** while the entire block guarded by `is True` is skipped:

| client | `is_public` | becomes public | `when_made_public` | old public demoted |
|---|---|---|---|---|
| JSON body (web UI) | `True` (bool) | ✅ | ✅ | ✅ |
| form-encoded | `"true"` / `"True"` / `"1"` | ✅ | ❌ **NULL** | ❌ **skipped** |

In a phase with `is_restricted_to_select_one_submission=True`, that leaves **two public submissions from a single request**, with no concurrency involved.

The web UI is unaffected — `challengeCtrl.js` sends `{"is_public": submissionVisibility}` as JSON, a real boolean. This is reached by API and CLI clients.

## Relationship to #4830 / #4831

#4830 reports exactly this broken state (multiple public submissions in a restricted phase) and attributes it to a race condition; PR #4831 adds `select_for_update()` locking at these same lines.

This PR fixes a **second, independent cause** of the same state — one that needs no concurrency at all. `is_public is True` does not appear anywhere in #4831's diff, so that locking change leaves this path live. The two fixes are complementary and do not conflict; I'd suggest #4830 stays open until both land.

## What changed

```python
is_public = serializers.BooleanField(required=False).to_internal_value(
    request.data["is_public"]
)
if is_public:
```

`BooleanField` is what the `Submission` serializer already applies to this same value downstream, so the view now tests exactly what gets stored. Invalid input raises `ValidationError`, which is caught and returned as HTTP 400 naming `is_public` rather than crashing.

The `and is_public` term in the inner condition became redundant (it is already inside `if is_public:`) and was removed. The existing `except KeyError: pass` is preserved, so a request that omits `is_public` entirely is still a no-op.

Verified the coerced branch behaves correctly for every input, including the absent-key no-op:

```
value           | when_made_public | demoted | 400
True/'true'/'True'/'1'/'yes'  | True  | True  | False
False/'false'/'0'             | False | False | False
'banana' / ''                 | False | False | True
<absent>                      | False | False | False
```

## Tests

Added three tests to `ChangeSubmissionDataAndVisibilityTest`:

- `test_change_visibility_stamps_when_made_public_for_form_request` — a form-encoded `is_public=true` results in a public submission **with** `when_made_public` set. Fails before this change: `when_made_public` stays `None`.
- `test_change_visibility_demotes_existing_public_for_form_request` — in the restricted phase, the previously public submission is demoted. Fails before this change: both submissions remain public, which is the #4830 symptom.
- `test_change_visibility_rejects_non_boolean_is_public` — `is_public=banana` returns 400 naming the field.

**Please note:** I could not execute these view-level tests locally — the app registry does not load on a current Python/Django (the pinned `django-rest-auth` is abandoned and will not install) — so they are written against this class's existing fixtures but have not been run green by me. I'd appreciate CI confirming them. The defect itself is verified independently of the tests, by simulating the exact branch and by confirming DRF's parser and `BooleanField` behaviour directly.

## Pre-existing failures

None introduced. I could not run the full suite for the environment reason above, so I am not claiming a clean run — only that `ruff check` and `ruff format --check` pass on both touched files.

## Scope

One defect, one call site. Related to my #5225 / #5226 (same field, different endpoint and different mechanism — that one is a 500, this one is silent data corruption), but deliberately kept as a separate branch so each can merge independently.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed form submissions so valid string values for the public visibility setting are correctly interpreted.
  * Invalid visibility values now return a clear validation error.
  * Preserved the rule that only one submission can be public in restricted phases, including when visibility is changed through form data.


Shivansh1205:
Verified, and the finding is correct — but it is **pre-existing**, not introduced here, so I have filed it separately as #5231 rather than widening this PR.

**It is not caused by this change.** The demote-before-validate ordering is identical on `master`. This PR's entire diff to that function is the boolean comparison:

```diff
-        is_public = request.data["is_public"]
-        if is_public is True:
+        is_public = serializers.BooleanField(required=False).to_internal_value(
+            request.data["is_public"]
+        )
+        if is_public:
```

plus removing the now-redundant `and is_public` and catching `ValidationError`. No write ordering changed.

**I confirmed the defect is real.** A PATCH carrying `is_public` plus one invalid field (`project_url` is `max_length=1000`, so 1001 characters is enough) commits the demotion at line 538, fails `is_valid()` at 553, and returns 400 at 558 — leaving the phase with zero public submissions.

**Why separately rather than here:**

1. The concurrency half of the suggested fix — `transaction.atomic()` plus row locking at these exact lines — is what open PR #4831 is already implementing for #4830. Doing it here would conflict directly with another contributor's open work.
2. This PR is a targeted fix for a type-coercion bug. Turning it into a transactional refactor of the same block would make a small, reviewable diff into a heavy one and couple two unrelated defects.

**One thing worth flagging honestly:** fixing #5227 makes this path *easier to hit*. Before this change the demotion never fired for form-encoded clients at all, so the ordering flaw was unreachable for them. It is still pre-existing rather than new, but the exposure does increase once this lands — I have said so explicitly in #5231 so it is not lost.

To summarise the three distinct defects now tracked in this block, since they are easy to conflate:

| | Mechanism | Result | Status |
|---|---|---|---|
| #4830 | concurrency | **two** public submissions | open, PR #4831 |
| #5227 | type coercion | demotion **skipped** | this PR |
| #5231 | ordering/atomicity | **zero** public submissions | filed, unassigned |

